PR #10, opened 2021-11-16
Gravatar - Lowercase emails before MD5 hash

JosephKav:
`Force all characters to lower-case`
https://en.gravatar.com/site/implement/hash/


kolaente:
Thanks!

PR #13, opened 2021-12-17
Add details of using NGINX Proxy Manager to the Reverse Proxy docs

TheBig-O:
Maybe not the best solution, but one that works. Please adjust the wording as you like. The key bit of information relates to directly editing the `.conf` file within NPMs setup.

### DIFF
--- a/docs/content/doc/setup/reverse-proxies.md
+++ b/docs/content/doc/setup/reverse-proxies.md
@@ -85,8 +85,8 @@ server {
 1. Create a Proxy Host within NPM and point it to whatever URL you plan to use. 
 2. Verify that the page will pull up in your browser. (Do not bother trying to log in. It won't work. Trust me.)
 3. Identify the container name for your NPM installation. e.g. `NGINX-PM`
-4. From the command line, enter sudo docker exec -it NGINX-PM /bin/bash and navigate to the proxy hosts folder where the `.conf` files are stashed. Probably `/data/nginx/proxy_host`. 
-5. Locate the `.conf` file where the server_name matches your Vikunja Proxy Host. Once found, add the following just above the existing location block in that file.
+4. From the command line, enter sudo docker exec -it NGINX-PM /bin/bash and navigate to the proxy hosts folder where the `.conf` files are stashed. Probably `/data/nginx/proxy_host`.
+5. Locate the `.conf` file where the server_name matches your Vikunja Proxy Host. Once found, add the following code, unchanged, just above the existing location block in that file. (They are listed by number, not name.)  
 ```
 location ~* ^/(api|dav|\.well-known)/ {
         proxy_pass http://api:3456;
--- a/docs/content/doc/setup/reverse-proxies.md
+++ b/docs/content/doc/setup/reverse-proxies.md
@@ -80,6 +80,22 @@ server {
 <b>NOTE:</b> If you change the max upload size in Vikunja's settings, you'll need to also change the <code>client_max_body_size</code> in the nginx proxy config.
 </div>
 
+## NGINX Proxy Manager (NPM)
+
+1. Create a Proxy Host within NPM and point it to whatever URL you plan to use. 
+2. Verify that the page will pull up in your browser. (Do not bother trying to log in. It won't work. Trust me.)
+3. Identify the container name for your NPM installation. e.g. `NGINX-PM`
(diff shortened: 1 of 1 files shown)


docs/content/doc/setup/reverse-proxies.md
```diff
@@ -80,6 +80,22 @@ server {
 <b>NOTE:</b> If you change the max upload size in Vikunja's settings, you'll need to also change the <code>client_max_body_size</code> in the nginx proxy config.
 </div>
 
+## NGINX Proxy Manager (NPM)
+
+1. Create a Proxy Host within NPM and point it to whatever URL you plan to use. 
+2. Verify that the page will pull up in your browser. (Do not bother trying to log in. It won't work. Trust me.)
+3. Identify the container name for your NPM installation. e.g. `NGINX-PM`
+4. From the command line, enter sudo docker exec -it NGINX-PM /bin/bash and navigate to the proxy hosts folder where the `.conf` files are stashed. Probably `/data/nginx/proxy_host`.
```
kolaente: Are these mounted on the host? Otherwise this won't last a restart of npm.

docs/content/doc/setup/reverse-proxies.md
```diff
@@ -80,6 +80,22 @@ server {
...
+5. Locate the `.conf` file where the server_name matches your Vikunja Proxy Host. Once found, add the following code, unchanged, just above the existing location block in that file. (They are listed by number, not name.)  
+```
+location ~* ^/(api|dav|\.well-known)/ {
+        proxy_pass http://api:3456;
+        client_max_body_size 20M;
+    }
+```
+6. After saving the edited file, return to NPM's browser window and refresh the page to verify your Proxy Host for Vikunja is still online. 
+7. The last step is to adjust the address in the top of the login subscreen to match your proxy address.
```
kolaente: Actually, this should not be required if the proxy is set up correctly.

kolaente (commented):
Awesome, thanks for the pr!

I'm not quite sure, but isn't this missing the container that should receive the traffic from vikunja? (In the first step)

docs/content/doc/setup/reverse-proxies.md
TheBig-O: Yes, these are mounted on the host as part of the NPM container. (If you don't point NPM to a persistent folder, all of your proxy hosts will disappear with each restart.
If you Docker exec -it [NPM container name] /bin/bash/, you'll cd into the `/data/nginx/proxy_host` folder and be able to edit the `.conf` files.

docs/content/doc/setup/reverse-proxies.md
TheBig-O: I would agree, but propose that it is the `app.ini` file or original Vikunja container that needs to be set up with the correct TLD. It's looking for the proxied host address and I hadn't set that up initially. (I've changed my Docker Stack so that it now has the correct address.)

docs/content/doc/setup/reverse-proxies.md
TheBig-O: For your first note about which container, you're right. I missed that, but am not sure how you want me to handle the rewrite for that. I would change it to the following:
1. Create a Proxy Host for the Vikunja Frontend within NPM and point it to whatever URL you plan to use. The steps that follow will enable the Proxy Host to successfully navigate to the API (on port 3456).
2. Verify that the page will pull up in your browser. (Do not bother trying to log in. It won't work. Trust me.)
3. The next steps apply to the NPM container so you need to Identify the container name for your NPM installation. e.g. `NGINX-PM`

TheBig-O:
I updated my branch and it should reflect all the changes I mention above. Let me know if I missed something.

docs/content/doc/setup/reverse-proxies.md
kolaente: >Yes, these are mounted on the host as part of the NPM container. (If you don't point NPM to a persistent folder, all of your proxy hosts will disappear with each restart.
> If you Docker exec -it [NPM container name] /bin/bash/, you'll cd into the /data/nginx/proxy_host folder and be able to edit the .conf files.

Ah okay, makes sense. I'd have edited them on the host but I guess that's a matter of preference. Should be fine the way you put it.

> For your first note about which container, you're right. I missed that, but am not sure how you want me to handle the rewrite for that. I would change it to the following:

I think that would work.

kolaente (approved):
Thanks again!

PR #50, opened 2022-09-27
fix(docs): Fix redirect_url example

fleaz:
Hey,

we just setup a Vikunja connected to Authelia and I was very pleased to see that there is an example for exact this usecase but we discovered a small mistake in the example. According to [this line](https://github.com/go-vikunja/api/blob/54b7f7127cb5aa5aba6a97798685efb59ac8ca83/pkg/modules/auth/openid/providers.go#L145) (and to our experience :D) the name of the provider from the config gets appended to the redirect_url therefore the example won't work like it was in the docs.

Best,
Felix


kolaente (approved):
Thanks!

PR #60, opened 2023-01-08
fix(mailer): forcessl config

Rein-R3:
1. add ssl setting into mail client.

I have tested the following mail service providers, all can be used.(port 465)

- gmail
- alibaba exchange mail
- tencent exchange mail

In gmail should  use [App Password](https://support.google.com/accounts/answer/185833?hl=en) to setup the vikunja config.

### DIFF
--- a/pkg/mail/mail.go
+++ b/pkg/mail/mail.go
@@ -58,6 +58,10 @@ func getClient() (*mail.Client, error) {
 		mail.WithTimeout((config.MailerQueueTimeout.GetDuration() + 3) * time.Second), // 3s more for us to close before mail server timeout
 	}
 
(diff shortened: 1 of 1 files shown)


kolaente:
How is this different to the currently set tls mode in https://github.com/go-vikunja/api/blob/main/pkg/mail/mail.go#L46-L48

Rein-R3:
> How is this different to the currently set tls mode here: https://github.com/go-vikunja/api/blob/main/pkg/mail/mail.go#L46-L48

image

In the 465 port connection must set SSL mode. tls mode is the options for the port.

kolaente:
okay, makes sense

PR #65, opened 2023-01-30
fix(docs): fix traefik v2 example

rriski:
The example from the docs lead to the following error in Traefik logs:

	ERR error="unexpected number of parameters; got 3, expected one of [1]"

Related discussion: https://community.vikunja.io/t/error-405-when-creating-a-user/928


kolaente (approved):
Thanks!

PR #87, opened 2023-10-09
When modifying/deleting users, if ParseInt fails for the user argument,

KaibutsuX:
assume argument is username instead of an id

### DIFF
--- a/pkg/cmd/user.go
+++ b/pkg/cmd/user.go
@@ -100,12 +100,16 @@ func getPasswordFromFlagOrInput() (pw string) {
 }
 
 func getUserFromArg(s *xorm.Session, arg string) *user.User {
+	filter := user.User{}
 	id, err := strconv.ParseInt(arg, 10, 64)
 	if err != nil {
-		log.Fatalf("Invalid user id: %s", err)
+		log.Info(fmt.Sprintf("Invalid userId [ %s ], assuming userName", arg))
+		filter.Username = arg
+	} else {
+		filter.ID = id
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/user.go
```diff
@@ -100,12 +100,16 @@ func getPasswordFromFlagOrInput() (pw string) {
 }
 
 func getUserFromArg(s *xorm.Session, arg string) *user.User {
+	filter := user.User{}
 	id, err := strconv.ParseInt(arg, 10, 64)
 	if err != nil {
-		log.Fatalf("Invalid user id: %s", err)
+		log.Info(fmt.Sprintf("Invalid userId [ %s ], assuming userName", arg))
```
kolaente:
```suggestion
		log.Infof("Invalid user ID [%s], assuming username instead", arg)
```

pkg/cmd/user.go
```diff
@@ -100,12 +100,16 @@ func getPasswordFromFlagOrInput() (pw string) {
 }
 
 func getUserFromArg(s *xorm.Session, arg string) *user.User {
+	filter := user.User{}
```
kolaente, line 103: Please rename this to `u` instead

kolaente (changes requested):

pkg/cmd/user.go
KaibutsuX: rename filter to u? `u` is also the return value and response from GetUserWithEmail.

kolaente: Didn't pay attention to that, makes sense. Please keep it.

kolaente (approved):

PR #241, opened 2024-03-20
Update zh-CN.json

ntlv1:
Corrected a translation mistake about "doneBucket".


kolaente:
Translation changes should happen via Crowdin: https://vikunja.io/docs/translations/

PR #608, opened 2025-04-02
feat: align caching and node version

dpschen:
- read out nvmrc for node version


kolaente (approved):
We should keep this up to date with the version that's used in docker image builds, but I assume renovate will take care of that.

dpschen:
> We should keep this up to date with the version that's used in docker image builds, but I assume renovate will take care of that.

That was also my assumption here.

PR #613, opened 2025-04-02
fix: align go version

dpschen:


dpschen:
@kolaente In order to test this, this needs to re-run, since the caches weren't full in the first run.

Currently I'm not able to trigger that.

kolaente:
I've triggered it now.

dpschen:
Since there is still [a cache hit for a newer go version](https://github.com/go-vikunja/vikunja/actions/runs/14217874844/job/39843363125?pr=613#step:5:24) I'm still not sure if this change is working.

@kolaente Could you clear the cache?

kolaente:
@dpschen Done and re-triggered the workflow, we'll see

dpschen:
> @dpschen Done and re-triggered the workflow, we'll see

Well unsure now. Seems like it's still hitting the cache:
https://github.com/go-vikunja/vikunja/actions/runs/14217874844/job/39858342713#step:5:24

But it's not matching. Any ideas? Else I guess we'll just dismiss this.

kolaente:
Maybe because the action does not use the version read from go.mod to construct the cache key? That would be a bug in the action.

In any case, I think this is fine. The step runs in 13s, which is pretty fast already.

dpschen:
Do you mean with this change or as it was before. Because how I understand it: the cache is not used, but at the same time the correct go version is used _with_ this PR.

kolaente:
With this change, it takes [13s](https://github.com/go-vikunja/vikunja/actions/runs/14217874844/job/39858342713?pr=613). On current main, it takes [8s](https://github.com/go-vikunja/vikunja/actions/runs/14351287793/job/40230543822). I'm unsure if this is worth putting much effort into, if it is solely about caching.

However, I think we should use the go version that's specified in the `go.mod` file.

Maybe this is worth filing an issue with the action repo?

.github/workflows/release.yml
```diff
@@ -76,9 +76,15 @@ jobs:
       - name: Git describe
         id: ghd
         uses: proudust/gh-describe@v2
+      # Workaround for `Cannot open: File exists` go caching issues
```
dpschen, line 76: @kolaente I found the issue. Not sure if that helps though. Also quite verbose

kolaente: Feels like a workaround, but fine by me if it works

kolaente:
It seems like this is no longer an issue: https://github.com/go-vikunja/vikunja/actions/runs/20751231423/job/59580930536

I'll close this, please ping if you feel there are still things to do.

PR #641, opened 2025-04-07
fix(auth): jwt token decoding

### LINKED ISSUE #388: Login Failure Due to Incorrect Base64 Decoding of Japanese Characters in JWT
### Description

The current implementation of base64 decoding in the frontend application leads to incorrect handling of non-ASCII characters, specifically Japanese characters. This leads to a **complete login failure** when processing user information extracted from JWT tokens received during the OpenID Connect login flow.

## Current behavior

The code responsible for decoding the base64-encoded payload of the JWT token (obtained via OpenID Connect) incorrectly replaces only the first instance of - with + and _ with /. Additionally, it fails to properly decode non-ASCII characters.

## Steps to reproduce

1. Attempt to log in to Vikunja using OpenID Connect.
1. Use an account where the returned JWT token contains Japanese characters in the user information.
1. Observe that the login process fails.

## Additional context

The current code for base64 decoding in `frontend/src/stores/auth.ts` is:

``` 
const base64 = jwt.split('.')[1].replace('-', '+').replace('_', '/')
const info = new UserModel(JSON.parse(atob(base64)))
```

This code only replaces the first instance of - and _. A correct implementation should use regular expressions with the g (global) flag to replace all occurrences.

```
// Replace all occurrences of - and _
const base64 = jwt.split('.')[1].replace(/-/g, '+').replace(/_/g, '/')
// Decode correctly, handling non-ASCII characters
const info = new UserModel(JSON.parse(decodeURIComponent(window.atob(base64).split('').map(function(c) {
        return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
    }).join(''))))
```

## Reference

This issue is similar to a problem reported in the microsoft-authentication-library-for-js repository: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/985

## Note
I am including the code directly in this report because my Gitea account is not yet approved. I apologize for any inconvenience this may cause.

### Vikunja Version

v0.24.6

### Browser and version

Chrome/131.0.6778.206

No

### Screenshots

_No response_

miikun77:
Globally replace '-' and '_' in JWT Base64 decoding
Fixes #388


kolaente (approved):
Thanks!

PR #655, opened 2025-04-10
fix: update repo and issue urls

dpschen:


kolaente:
I've added redirects for the frontend and desktop to `code.vikunja.io`. We should use these urls everywhere, to make sure they can be changed easily if needed.

desktop/package.json
```diff
@@ -3,7 +3,10 @@
   "version": "v0.1.0",
   "description": "Vikunja's frontend as a standalone desktop application.",
   "main": "main.js",
-  "repository": "https://code.vikunja.io/desktop",
+  "repository": "https://code.vikunja.io/vikunja/tree/main/desktop",
```
kolaente, line 6: This could stay at code.vikunja.io/desktop

frontend/package.json
```diff
@@ -6,10 +6,10 @@
   "license": "AGPL-3.0-or-later",
   "repository": {
     "type": "git",
-    "url": "https://kolaente.dev/vikunja/vikunja"
+    "url": "https://code.vikunja.io/vikunja/tree/main/frontend"
```
kolaente, line 9: This could be code.vikunja.io/frontend

desktop/package.json
dpschen: I see. Before I update this: I think I was confused by the fact that https://code.vikunja.io/frontend/tree/main/src/i18n/lang redirects to the old frontend repo. Is that intended?

dpschen: To be more precise:

Since

```
https://code.vikunja.io/frontend
```

redirects to

```
https://github.com/go-vikunja/vikunja/tree/main/frontend
```

I would expect

```
https://code.vikunja.io/frontend/tree/main/src/i18n/lang
```
to redirect to

```
https://github.com/go-vikunja/vikunja/tree/main/frontend/src/i18n/lang
```

----

Instead we end up at

```
https://github.com/go-vikunja/frontend/tree/main/src/i18n/lang
```

kolaente: Ah well I guess that's a bug in the redirection

https://github.com/go-vikunja/go-package-redirect/blob/ae38a9d60e73f18abc837af6edb19c232c422098/src/index.ts#L52

Yeah seems like it only matches the exact suffix and not the thing after it.

dpschen: I created a pr for the redirects https://github.com/go-vikunja/go-package-redirect/pull/1. I'll adjust this pr, when the redirect is fixed.

dpschen:
@kolaente also found https://hub.docker.com/r/vikunja/golang-build (the golang-build or is broken)

PR #662, opened 2025-04-11
wip: fix build status badges

dpschen:


README.md
```diff
@@ -1,6 +1,6 @@
 <img src="https://vikunja.io/images/vikunja-logo.svg" alt="" style="display: block;width: 50%;margin: 0 auto;" width="50%"/>
 
-[![Build Status](https://drone.kolaente.de/api/badges/vikunja/vikunjaa/status.svg)](https://drone.kolaente.de/vikunja/vikunja)
+[![Build Status](https://github.com/go-vikunja/vikunja/actions/workflows/ci.yml/badge.svg)](https://github.com/go-vikunja/vikunja/actions/workflows/ci.yml)
```
dpschen, line 3: If we build frontend and desktop in dedicated workflows the badge could better represent what was actually built.

kolaente:
Isn't the Frontend missing?

dpschen:
@kolaente There wasn't a badge for frontend so far. And as said I'm unsure I should even add one for desktop, since the badge mirrors the overall ci state and not the one from individual apps, since for that to work with the standard github badges these builds would need dedicated worklows.

README.md
kolaente: I mean we could, but then we couldn't share the built frontend files between steps

PR #663, opened 2025-04-11
Matrix for arches

### LINKED ISSUE #528: .deb package available only for amd64
_Original issue by BCMM on 2022-12-02T10:29:52.000Z_

I would like to install the backend on a Debian armhf server. Is it intentional that the Debian package is only available for amd64?

[Original issue on Gitea](https://kolaente.dev/vikunja/vikunja/issues/1331)

---

_**@kolaente** commented on 2022-12-02T15:24:49.000Z_:

It's not intentional, it just hasn't been implemented yet. Looks like nfpm does support different architectures but not different binaries for these architectures. If I understand it correctly that means we'd need to template the nfpm config file and run it once for every architecture, adding the previously built binaries for each arch.

Do you want to send a PR?

---

_**BCMM** commented on 2023-01-25T18:00:58.000Z_:

> Do you want to send a PR?

I'm afraid the whole Go ecosystem, including its build tools, is rather unfamiliar to me. I'm still not quite clear which components are involved in building Vikunja and how they interact.

(Sorry for the closing/reopening. I pressed the wrong button.)

---

_**@kolaente** commented on 2023-01-25T21:18:46.000Z_:

Sure, that's fine!

The nfpm config docs show they expand environment variables, maybe that could be used to release packages from different binaries (talking about the ci)

https://nfpm.goreleaser.com/configuration/

andrewwippler:
Resolves #528 

I am hoping this adds multiarch support, but there is a gotcha listed here: https://nfpm.goreleaser.com/goarch-to-pkg/ ... not every distro has the same architecture extension.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -140,6 +140,12 @@ jobs:
           - deb
           - apk
           - archlinux
+        arch:
+          - amd64
+          - arm64
+          - "386"
+          - armv6
+          - armv7
     
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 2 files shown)


kolaente:
> I am hoping this adds multiarch support, but there is a gotcha listed here:

From the docs I'd assume nfpm does the translation?

kolaente (commented):
You'll need to get the correct binary from earlier release artifacts as well (in the prepare step and when downloading the binary from artifacts)

andrewwippler:
I am currently out of bandwidth at the moment. I am going to close this pull request so someone else can add this feature if they need it.

PR #701, opened 2025-04-24
fix(db): refactor filtering with subqueries

### LINKED ISSUE #285: Adding 'labels != <anything>' filter filters out all issues without a label set
### Description

If you try to add any filter to NOT show issues with a particular label it also filters out all issues that have no label set. Expected behavior would be to show also issues without any label set and only filter out issues with specified label in the filter.

How to reproduce on the demo site:
- go to Inbox project
- edit List view and add '&& labels != Personal' to the filter
- all issues without label set will disappear

### Vikunja Version

0.24

### Browser and version

_No response_

Yes

### Screenshots

_No response_

vovochka404:
This refactoring allows correctly apply positive and negative filtering to all sub-queries.

Fixes #285


pkg/models/task_search.go
```diff
@@ -34,6 +34,61 @@ import (
...
+	},
+	"reminders": {
+		Table:           "task_reminders",
+		BaseFilter:      "tasks.id = task_id",
+		FilterableField: "reminder",
+		AllowNullCheck:  true,
+	},
+	"assignees": {
+		Table:           "task_assignees",
+		BaseFilter:      "tasks.id = task_id",
```
kolaente, line 67: If this is always the same, would it make sense to add this as fallback?

On the other hand, it might be good to have it explicit.

kolaente (commented):
Looks quite good already, will do an in-depth review later.

Can you add a few tests for the new cases?

vovochka404:
> Looks quite good already, will do an in-depth review later.
> 
> Can you add a few tests for the new cases?

Fixed test and added one more for filtering with `IncludeNulls`. Andd yep, there was a bug in my logic. I thought that builder object is immutable. Thus, had to redesign work with base filter.

pkg/models/task_collection_test.go
```diff
@@ -1272,8 +1314,38 @@ func TestTaskCollection_ReadAll(t *testing.T) {
 			want: []*Task{
 				task1,
 				task2,
+				task3,
```
kolaente, line 1317: Shouldn't this also depend on `include nulls`?

pkg/models/task_search.go
```diff
@@ -44,6 +99,20 @@ type dbTaskSearcher struct {
 	hasFavoritesProject bool
 }
 
+func (sf *SubTableFilter) ToBaseSubQuery() *builder.Builder {
+	var cond = builder.
+		Select("1").
+		From(sf.Table).
+		Where(builder.And(builder.Expr(sf.BaseFilter)))
```
kolaente: No need for `builder.And` if there's always only one condition.

pkg/models/task_search.go
```diff
@@ -102,103 +170,46 @@ func convertFiltersToDBFilterCond(rawFilters []*taskFilter, includeNulls bool) (
...
-				field:      "reminder",
-				value:      f.value,
-				comparator: f.comparator,
-				isNumeric:  f.isNumeric,
-			}, includeNulls)
-			if err != nil {
-				return nil, err
+		subTableFilterParams, ok := subTableFilters[f.field]
+		if ok {
+			// in fact, it's better to skip all unstrict filters when we're filtering with subquery.
```
kolaente: Is this a todo? Because the comment seems to be different from the code around it.

pkg/models/task_search.go
```diff
@@ -102,103 +170,46 @@ func convertFiltersToDBFilterCond(rawFilters []*taskFilter, includeNulls bool) (
...
-				// recreating the struct here to avoid modifying it when reusing the opts struct
-				field:      "reminder",
-				value:      f.value,
-				comparator: f.comparator,
-				isNumeric:  f.isNumeric,
-			}, includeNulls)
-			if err != nil {
-				return nil, err
+		subTableFilterParams, ok := subTableFilters[f.field]
+		if ok {
```
kolaente, line 174: Can you return earlier?

pkg/models/task_collection_test.go
vovochka404: Why? Tasks without labels fully comply with the condition "does not contain the specified label"

pkg/models/task_search.go
vovochka404: Fixed

pkg/models/task_search.go
vovochka404: Removed comment. It's not valid.

pkg/models/task_search.go
vovochka404: I don't see how to make it better.

pkg/models/task_collection_test.go
kolaente: That does make sense, I guess I was just confused by the change in assertions.

PR #705, opened 2025-04-25
fix(docs): improve swagger output by setting swaggertype and enums

capriolo:
Added `swaggertype:"string"` and `enums` tags to fields in ProjectView to ensure correct type representation in Swagger docs.

Affected fields:
- ViewKind
- Filter
- BucketConfigurationMode

### DIFF
--- a/pkg/models/project_view.go
+++ b/pkg/models/project_view.go
@@ -139,7 +139,7 @@ type ProjectView struct {
 	ViewKind ProjectViewKind `xorm:"not null" json:"view_kind" swaggertype:"string" enums:"list,gantt,table,kanban"`
 
 	// The filter query to match tasks by. Check out https://vikunja.io/docs/filters for a full explanation.
-	Filter *TaskCollection `xorm:"json null default null" query:"filter" json:"filter" swaggertype:"string"`
+	Filter *TaskCollection `xorm:"json null default null" query:"filter" json:"filter"`
 	// The position of this view in the list. The list of all views will be sorted by this parameter.
 	Position float64 `xorm:"double null" json:"position"`
 
--- a/pkg/models/project_view.go
+++ b/pkg/models/project_view.go
@@ -66,6 +66,10 @@ func (p *ProjectViewKind) UnmarshalJSON(bytes []byte) error {
 	return nil
 }
 
+// NOTE: When adding or changing enum values for ProjectViewKind,
+// make sure to update the corresponding `enums` tag in the ProjectView struct
+// to keep the OpenAPI documentation in sync.
+
 const (
 	ProjectViewKindList ProjectViewKind = iota
 	ProjectViewKindGantt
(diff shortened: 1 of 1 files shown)


pkg/models/project_view.go
```diff
@@ -128,15 +128,15 @@ type ProjectView struct {
 	// The project this view belongs to
 	ProjectID int64 `xorm:"not null index" json:"project_id" param:"project"`
 	// The kind of this view. Can be `list`, `gantt`, `table` or `kanban`.
-	ViewKind ProjectViewKind `xorm:"not null" json:"view_kind"`
+	ViewKind ProjectViewKind `xorm:"not null" json:"view_kind" swaggertype:"string" enums:"list,gantt,table,kanban"`
```
kolaente, line 139: Can you add a note to where the type is defined so that we'll remember to update it here when adding new cases?

pkg/models/project_view.go
```diff
@@ -128,15 +128,15 @@ type ProjectView struct {
...
 
 	// The filter query to match tasks by. Check out https://vikunja.io/docs/filters for a full explanation.
-	Filter *TaskCollection `xorm:"json null default null" query:"filter" json:"filter"`
+	Filter *TaskCollection `xorm:"json null default null" query:"filter" json:"filter" swaggertype:"string"`
```
kolaente: That's not a string, it's a subtype

pkg/models/project_view.go
```diff
@@ -128,15 +128,15 @@ type ProjectView struct {
...
 	// The position of this view in the list. The list of all views will be sorted by this parameter.
 	Position float64 `xorm:"double null" json:"position"`
 
 	// The bucket configuration mode. Can be `none`, `manual` or `filter`. `manual` allows to move tasks between buckets as you normally would. `filter` creates buckets based on a filter for each bucket.
-	BucketConfigurationMode BucketConfigurationModeKind `xorm:"default 0" json:"bucket_configuration_mode"`
+	BucketConfigurationMode BucketConfigurationModeKind `xorm:"default 0" json:"bucket_configuration_mode" swaggertype:"string" enums:"none,manual,filter,manual"`
```
kolaente, line 147: Same here as for the view kind

pkg/models/project_view.go
capriolo: @kolaente Hey, It seems to be a subtype but for some reason in API it returns a string:
image

pkg/models/project_view.go
capriolo: done

pkg/models/project_view.go
capriolo: done

pkg/models/project_view.go
kolaente: This has been changed a while ago, which version are you using?

Here's ther response from the demo (which runs unstable):

image

capriolo: Tested on the stable (v0.24.6) version.
I removed swaggertype for this field

kolaente (approved):

PR #744, opened 2025-05-08
feat: allow to write date via Quick Add Magic in DMY format with dots

Hudint:
Currently you can only add a date in MDY format (02/17/2021) via Quick Add Magic.
With this change, the DMY format is supported when using dots (e.g.: 17.02.2021)


frontend/src/helpers/time/parseDate.ts
```diff
@@ -138,46 +138,37 @@ const addTimeToDate = (text: string, date: Date, previousMatch: string | null):
...
+		/(^| )(?<foundText>(?<month>[0-9][0-9]?)\/(?<day>[0-9][0-9]?)(\/(?<year>[0-9][0-9]([0-9][0-9])?))?)($| )/gi,
+		/(^| )(?<foundText>(?<year>[0-9][0-9][0-9][0-9]?)\/(?<day>[0-9][0-9]?)\/(?<month>[0-9][0-9]))($| )/gi,
+		/(^| )(?<foundText>(?<year>[0-9][0-9][0-9][0-9]?)-(?<day>[0-9][0-9]?)-(?<month>[0-9][0-9]))($| )/gi,
+		/(^| )(?<foundText>(?<day>[0-9][0-9]?)\.(?<month>[0-9][0-9]?)\.(?<year>[0-9][0-9]([0-9][0-9])?))($| )/gi
+	]
+
+	// 1. Try parsing the text as a "usual" date, like 2021-06-24 or "06/24/2021" or "27/01" or "01/27"
+	for (let dateRegex of dateRegexes) {
+		const result = dateRegex.exec(text)
+		console.log(result)
```
kolaente: This seems like it should be removed

kolaente (changes requested):
Looks quite good already! Can you add more tests for the new format and adjust the existing ones?

frontend/src/helpers/time/parseDate.ts
Hudint: I forgot that :D

kolaente (approved):
Thanks!

Issue #728, opened 2025-05-06
iOS CalDAV works but breaks as soon as recurring task is created in Vikunja


harrystuart:
### Description

Hello, Vikunja makes clear that iOS CalDAV is not strongly supported. Support, however, would enable many users to sync with iOS Reminders which would be an absolute godsend.

I have successfully connected to and used Vikunja CalDAV with iOS Reminders. However, as soon as I create a recurring task in Vikunja, CalDAV completely breaks.

Below is the self-contained experiment I ran:

1. `docker-compose up | tee vikunja.log`
2. Logged into web app using Test
3. Connected to CalDAV with iOS
4. Created Test Web Task 1 in web app
5. Test Web Task 1 displayed in iOS Reminders
6. Created Test iOS Task 1 in iOS Reminders
7. Test iOS Task 1 displayed in web app
8. Created Test Web Recurring Task 1 with daily recurrence since completion (and later added due tomorrow) in web app
9. Test Web Recurring Task 1 not displayed in iOS Reminders
10. Created Test Web Task 2 in web app
11. Test Web Task 2 not displayed in iOS Reminders
12. Created Test iOS Task 2 in iOS Reminders
13. Test iOS Task 2 displayed in web app
14. Deleted CalDAV account from iOS
15. Restarted iOS
16. Connected to CalDAV with iOS
17. No tasks displayed in iOS Reminders

My `docker-compose.yml` looks like so:

```yaml
vikunja:
  image: vikunja/vikunja
  environment:
    VIKUNJA_DATABASE_TYPE: sqlite
    VIKUNJA_SERVICE_PUBLICURL: https://vikunja.hstu.net
    VIKUNJA_LOG_LEVEL: DEBUG
    VIKUNJA_LOG_EVENTS: stdout
    VIKUNJA_LOG_EVENTSLEVEL: DEBUG
  ports:
    - 3456:3456
  volumes:
    - /storage/enc/docker/vikunja/files:/app/vikunja/files
    - /storage/enc/docker/vikunja/db:/db
```

I have attached the logs output. I would greatly appreciate some assistance in parsing these log files and trying to arrest the offending backend logic - I am not a Go dev but am happy to do what I can if pointed in the right direction.

[vikunja.log](https://github.com/user-attachments/files/20073487/vikunja.log)

### Vikunja Version

v0.24.6

### Can you reproduce the bug on the Vikunja demo site?

No

### Screenshots

[...]

PR #748, opened 2025-05-09
Enable iOS CalDAV support for recurring tasks

harrystuart:
This PR updates how Vikunja stores tasks in its CalDAV server. Specifically, it maps `RRULE:FREQ=SECONDLY` to the appropriate `DAILY` or `WEEKLY` interval frequency where possible.

**Why?**

Vikunja currently stores all CalDAV recurring tasks with `FREQ=SECONDLY` in its `RRULE` field unless the "Repeat mode" is set to "Monthly". This is [permissible ](https://datatracker.ietf.org/doc/html/rfc5545#section-3.3.10) and logical as tasks recurring daily or weekly can be represented as recurring every X seconds. Unfortunately, [iOS does not support](https://developer.apple.com/documentation/eventkit/ekrecurrencefrequency) CalDAV tasks with a `SECONDLY` interval frequency.

If you try and connect to Vikunja CalDAV from iOS and create, for example, a daily recurring task, iOS Reminders will break and refuse to show subsequent updates to the task list.

Please see [this iseeu](https://github.com/go-vikunja/vikunja/issues/728) to understand my debugging process.

### DIFF
--- a/pkg/caldav/caldav.go
+++ b/pkg/caldav/caldav.go
@@ -168,10 +168,29 @@ PRIORITY:` + strconv.Itoa(mapPriorityToCaldav(t.Priority))
 		}
 
 		if t.RepeatAfter > 0 || t.RepeatMode == models.TaskRepeatModeMonth {
-			if t.RepeatMode == models.TaskRepeatModeMonth {
+			switch {
+		
+			case t.RepeatMode == models.TaskRepeatModeMonth:
 				caldavtodos += `
-RRULE:FREQ=MONTHLY;BYMONTHDAY=` + t.DueDate.Format("02") // Day of the month
-			} else {
+RRULE:FREQ=MONTHLY;BYMONTHDAY=` + t.DueDate.Format("02")
+		
+			case t.RepeatAfter%604800 == 0:
+				weeks := t.RepeatAfter / 604800
+				if weeks < 1 {
+					weeks = 1
+				}
+				caldavtodos += `
+RRULE:FREQ=WEEKLY;INTERVAL=` + strconv.FormatInt(weeks, 10)
+		
+			case t.RepeatAfter%86400 == 0:
+				days := t.RepeatAfter / 86400
+				if days < 1 {
+					days = 1
+				}
+				caldavtodos += `
+RRULE:FREQ=DAILY;INTERVAL=` + strconv.FormatInt(days, 10)
+		
+			default:
 				caldavtodos += `
 RRULE:FREQ=SECONDLY;INTERVAL=` + strconv.FormatInt(t.RepeatAfter, 10)
 			}


kolaente (changes requested):
Looks pretty good already! Can you add tests for the new cases as well?

harrydabble:
Hey @kolaente, I'm unfamiliar with this tech-stack and Go. I should hopefully get a chance later next week to sit down and figure out how to add tests - obviously very important. Sorry I am unable to add these right away! Cheers

harrystuart:
Hey @kolaente, I'm unfamiliar with this tech-stack and Go. I should hopefully get a chance later next week to sit down and figure out how to add tests - obviously very important. Sorry I am unable to add these right away! Cheers

harrystuart:
BTW @kolaente, I learned that when creating tasks in iOS with a due date (necessary for recurring tasks from iOS -> Vikunja), Vikunja expects `VTODO` to be the first element of `VCALENDAR`. I've made a quick change to fix this.

kolaente:
> BTW @kolaente, I learned that when creating tasks in iOS with a due date (necessary for recurring tasks from iOS -> Vikunja), Vikunja expects `VTODO` to be the first element of `VCALENDAR`. I've made a quick change to fix this.

Sounds good, but can you move this to a separate PR? That way the history stays clean.

kolaente:
Hey @harrystuart do you need more input to continue with this?

kolaente:
I went ahead and refactored your approach in https://github.com/go-vikunja/vikunja/pull/1230. Thanks for your work!

pkg/caldav/parsing.go
harrystuart, line 1: @kolaente see herre

harrystuart:
Hi @kolaente , apologies for the radio-silence. Been busy moving overseas and changing career. 

Thanks for deploying this change. It's been a while so I may misremember a couple of details but there were 2 things I needed to change for tasks to properly sync with iOS:

1. Change RRULE from `secondly` to supported frequencies.
2. Update the parsing logic to flexibly look for `VTODO` instead of asserting that it must be the first element (iOS sends other stuff before `VTODO` thus causing parsing to fail.)

For 2, it seems like the old logic is still [present](https://github.com/go-vikunja/vikunja/blob/26e5a36adce018af24f6e36a8e47a9aae9d4d17f/pkg/caldav/parsing.go#L277C2-L280C3). In this PR I suggested iterating elements to look for `VTODO` - I have tagged you in the corresponding logic. 

I haven't used Vikunja in a bit but I suspect the current module will still break iOS syncing. I'm afraid I'm quite busy at the moment and haven't got my local dev env setup to properly test such a change etc... I will just patch the file for myself quickly and return to raising a PR if i get some free time.

kolaente:
Thanks for the clarification!

I've cherry-picked the change to find the actual `VTODO` component in 62799c129b96173993827d74d09be2ebf10dbf67, please check with the next unstable build.

PR #753, opened 2025-05-12
fix: cypress selector

dpschen:


frontend/cypress/e2e/task/task.spec.ts
```diff
@@ -81,7 +81,7 @@ describe('Task', () => {
 		cy.visit('/projects/1/1')
 		cy.get('.project-is-empty-notice')
 			.should('not.exist')
-		cy.get('.input[placeholder="Add a task…"')
+		cy.get('.input[placeholder="Add a task…"]')
```
kolaente, line 84: Interesting how this has worked in the first place :sweat_smile:

kolaente (approved):

frontend/cypress/e2e/task/task.spec.ts
dpschen: I have no idea why

Issue #682, opened 2025-04-18
Vikunja keeps redirecting me to /password-reset


NihalNavath:
### Description

I'm facing an issue where Vikunja keeps redirecting me to /password-reset, even when I try to access the base domain at http://todo.domain.com/, navigate directly to /login, or visit any other path.
(I did not ask for a password reset and hadn't logged into vikunja for weeks)

In a bid to fix it, I followed the flow and attempted to reset the password, but I receive the following message:
'Invalid token to reset a user's password.'

To troubleshoot, I even used the CLI to manually reset the password, but the issue still persists.

Interestingly, everything works fine in an incognito window, but not in a regular session or when using the Edge-installed Progressive Web App (PWA). I don't have any extensions installed.

I have attached a video of the issue

### Vikunja Version

v0.24.6

### Browser and version

Edge135.0.3179.73 (Official build) (64-bit) 

### Can you reproduce the bug on the Vikunja demo site?

No

### Screenshots

![Image](https://github.com/user-attachments/assets/2d8f0716-287d-4990-bcb2-28d00b739353)

kolaente:
This sounds like you somehow got to that page once and then it will redirect you every time to that, because that's what you visited before. If it works in a private tab (as you described), that would confirm that.

Can you check what you have saved in local storage using your browser's dev tools?

I think there are a few ways to solve this:

1. Only redirect once, then remove the local storage item for the redirect 
2. Redirect using query parameter
3. Only redirect when the user is logged in and ignore the local storage item when logged out

@dpschen IIRC we talked about using a query parameter, what do you think?

NihalNavath:
"This sounds like you somehow got to that page once and then it will redirect you every time to that, because that's what you visited before"

Yup, you seem to have hit the nail on the head with this one. Vikunja force-redirects to `/password-reset` if there's a `"passwordResetToken"` key in localStorage and if that token has expired, you're basically stuck in a loop with no way out. Using query parameters would definitely help avoid this issue.

I had that key in my localStorage too, and manually removing it via the console resolved the issue for me.

kolaente:
Will be fixed by https://github.com/go-vikunja/vikunja/pull/770

PR #770, opened 2025-05-14
fix(auth): only use query parameters instead of local storage for password reset token

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/682

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,7 @@
         "editor.defaultFormatter": "dbaeumer.vscode-eslint"
     },
     "[typescript]": {
-        "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+        "editor.defaultFormatter": "vscode.typescript-language-features"
     },
 
     // https://eslint.vuejs.org/user-guide/#editor-integrations
--- a/frontend/cypress/factories/project.ts
+++ b/frontend/cypress/factories/project.ts
@@ -1,10 +1,18 @@
 import {Factory} from '../support/factory'
 import {faker} from '@faker-js/faker'
 
+export interface ProjectAttributes {
+	id: number | string; // Allow string for '{increment}'
+	title: string;
+	owner_id: number;
+	created: string;
+	updated: string;
+}
+
 export class ProjectFactory extends Factory {
 	static table = 'projects'
 
-	static factory() {
+	static factory(): Omit<ProjectAttributes, 'id'> & { id: string } { // id is '{increment}' before seeding
 		const now = new Date()
 
 		return {
--- a/frontend/cypress/factories/token.ts
+++ b/frontend/cypress/factories/token.ts
@@ -1,0 +1,29 @@
+import {faker} from '@faker-js/faker'
+import {Factory} from '../support/factory'
+
+export interface TokenAttributes {
+	id: number | string; // Allow string for '{increment}'
+	user_id: number;
+	token: string;
+	kind: number;
+	created: string;
+}
+
+export class TokenFactory extends Factory {
+	static table = 'user_tokens'
+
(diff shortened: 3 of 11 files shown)


frontend/cypress/factories/project.ts
```diff
@@ -1,10 +1,18 @@
 import {Factory} from '../support/factory'
 import {faker} from '@faker-js/faker'
 
+export interface ProjectAttributes {
+	id: number | string; // Allow string for '{increment}'
```
dpschen, line 5:
```ts
	id: number | '{increment}';

File: frontend/cypress/factories/token.ts [hunk 2]
```diff
@@ -0,0 +1,29 @@
+import {faker} from '@faker-js/faker'
+import {Factory} from '../support/factory'
+
+export interface TokenAttributes {
+	id: number | string; // Allow string for '{increment}'
```
dpschen, line 5: s.o.

frontend/cypress/factories/user.ts
```diff
@@ -2,16 +2,27 @@ import {faker} from '@faker-js/faker'
 
 import {Factory} from '../support/factory'
 
+export interface UserAttributes {
+	id: number | string;
```
dpschen, line 6:
```ts
	id: number | 'password';
```

frontend/src/router/index.ts
```diff
@@ -394,6 +394,17 @@ export async function getAuthForRoute(to: RouteLocation, authStore) {
 		return
 	}
 	
+	// Check if password reset token is in query params
+	const resetToken = to.query.userPasswordReset as string | undefined
+	if (resetToken) {
+		authStore.setPasswordResetToken(resetToken)
```
dpschen, line 400: I might misunderstand the reset flow, but now that we do use the url for the token is there even a need to store the token anywhere else? A comment might make this easier to understand

frontend/src/views/user/PasswordReset.vue
```diff
@@ -62,22 +65,32 @@ const credentials = reactive({
...
 const errorMsg = ref('')
 const successMessage = ref('')
 
 async function resetPassword() {
 	errorMsg.value = ''
-	
-	if(credentials.password === '') {
+	const token = route.query.token as string
+
+	if (!token) {
```
dpschen, line 80: Seems like this check should happen in beforeRouteEnter of this route now instead of the old global router hook.

.vscode/settings.json
```diff
@@ -11,7 +11,7 @@
         "editor.defaultFormatter": "dbaeumer.vscode-eslint"
     },
     "[typescript]": {
-        "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+        "editor.defaultFormatter": "vscode.typescript-language-features"
```
dpschen, line 14: Out of interest: why this change?

frontend/cypress/factories/user.ts
```diff
@@ -2,16 +2,27 @@ import {faker} from '@faker-js/faker'
...
+	language: string;
+	created: string;
+	updated: string;
+}
+
 export class UserFactory extends Factory {
 	static table = 'users'
 
-	static factory() {
+	static factory(): Omit<UserAttributes, 'id' | 'password'> & { id: string; password?: string } {
```
dpschen, line 19: why omit password if it's added again?

frontend/src/router/index.ts
kolaente: It probably is not required anymore.

.vscode/settings.json
kolaente: I think that was a change which happened automatically because I changed the settings in my vscode locally (using the gui). Not really intended.

PR #797, opened 2025-05-19
feat(auth): add ForceUserInfo option to OpenID provider

maggch97:
# Add ForceUserInfo option to OpenID provider

## Problem
When using Casdoor as an OpenID provider, there's an inconsistency between the user information in the JWT token and the UserInfo endpoint. The token contains the user's unique ID in the `name` field, while the UserInfo endpoint correctly returns the user's display name.

## Solution
This PR adds a new `ForceUserInfo` option to the OpenID provider configuration. When enabled, it forces the use of the UserInfo endpoint to retrieve user information instead of relying on claims from the ID token.

## Impact
- Default behavior remains unchanged (backward compatible)
- New option allows administrators to force using UserInfo endpoint data
- Particularly useful for providers like Casdoor that don't fully comply with OIDC standards

## Related
I've opened an issue in the Casdoor repository (https://github.com/casdoor/casdoor/issues/3806) to discuss the root cause. However, changing Casdoor's token structure might cause significant compatibility issues for existing integrations, so it's unclear if this can be fixed at the provider level. This PR provides a workaround in Vikunja that doesn't affect existing functionality.

Docs PR: https://github.com/go-vikunja/website/pull/107

### DIFF
--- a/config-raw.json
+++ b/config-raw.json
@@ -709,6 +709,11 @@
                                             "key": "emailfallback",
                                             "default_value": "false",
                                             "comment": "This option allows to look for a local account where the OIDC user's email match the Vikunja local email. Allowed value is either `true` or `false`. That option can be combined with `usernamefallback`.\nUse with caution, this can allow the 3rd party provider to connect to *any* local account and therefore potential account hijaking."
+                                        },
+                                        {
+                                            "key": "forceuserinfo",
+                                            "default_value": "false",
+                                            "comment": "This option forces the use of the OpenID Connect UserInfo endpoint to retrieve user information instead of relying on claims from the ID token. When set to `true`, user data (email, name, username) will always be obtained from the UserInfo endpoint even if the information is available in the token claims. This is useful for providers that don't include complete user information in their tokens or when you need the most up-to-date user data. Allowed value is either `true` or `false`."
                                         }
                                     ]
                                 }
--- a/pkg/modules/auth/openid/openid.go
+++ b/pkg/modules/auth/openid/openid.go
@@ -56,6 +56,7 @@ type Provider struct {
 	Scope            string `json:"scope"`
 	EmailFallback    bool   `json:"email_fallback"`
 	UsernameFallback bool   `json:"username_fallback"`
+	ForceUserInfo    bool   `json:"force_user_info"`
 	ClientSecret     string `json:"-"`
 	openIDProvider   *oidc.Provider
 	Oauth2Config     *oauth2.Config `json:"-"`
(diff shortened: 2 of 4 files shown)


pkg/modules/auth/openid/providers.go
```diff
@@ -180,6 +181,16 @@ func getProviderFromMap(pi map[string]interface{}, key string) (provider *Provid
 			usernameFallback = usernameFallbackTypedValue
 		}
 	}
+
+	var forceUserInfo = false
+	forceUserInfoValue, exists := pi["forceuserinfo"]
+	if exists {
+		forceUserInfoTypedValue, ok := forceUserInfoValue.(bool)
+		if ok {
```
kolaente, line 189: Maybe worth logging an error when the value is not a bool?

config-raw.json
```diff
@@ -709,6 +709,11 @@
...
                                             "comment": "This option allows to look for a local account where the OIDC user's email match the Vikunja local email. Allowed value is either `true` or `false`. That option can be combined with `usernamefallback`.\nUse with caution, this can allow the 3rd party provider to connect to *any* local account and therefore potential account hijaking."
+                                        },
+                                        {
+                                            "key": "forceuserinfo",
+                                            "default_value": "false",
+                                            "comment": "This option forces the use of the OpenID Connect UserInfo endpoint to retrieve user information instead of relying on claims from the ID token. When set to `true`, user data (email, name, username) will always be obtained from the UserInfo endpoint even if the information is available in the token claims. This is useful for providers that don't include complete user information in their tokens or when you need the most up-to-date user data. Allowed value is either `true` or `false`."
```
kolaente, line 716: Once this is merged, please add a note about this to https://github.com/go-vikunja/website/blob/main/src/content/docs/setup/openid.mdoc as well.

kolaente (commented):
Thanks for the PR! Looks quite good already.

Can you add a test for this?

pkg/modules/auth/openid/providers.go
maggch97: I added unit tests for merging `cl` and `cl2`. Because it's not easy to mock the `provider`, so I extract the code of merging 
 from `getClaims` to a new function `mergeClaims`, then test `mergeClaims`

config-raw.json
maggch97: Sure, will update after this PR is merged

pkg/modules/auth/openid/openid.go
```diff
@@ -288,6 +289,33 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
...
+	if (forceUserInfo && cl2.PreferredUsername != "") || cl.PreferredUsername == "" {
+		cl.PreferredUsername = cl2.PreferredUsername
+	}
+
+	if cl.PreferredUsername == "" && cl2.Nickname != "" {
+		cl.PreferredUsername = cl2.Nickname
+	}
+
+	if cl.Email == "" {
+		log.Errorf("Claim does not contain an email address for provider %s", providerName)
```
kolaente: Unrelated, but we can get rid of this line and thus the `providerName` parameter

maggch97: Done, the `providerName` parameter is removed and the error logging is moved back to `getCliams`
```go
		if err != nil {
			if user.IsErrNoEmailProvided(err) {
				log.Errorf("Claim does not contain an email address for provider %s", provider.Name)
			}
```

kolaente (approved):
Thanks!

pkg/modules/auth/openid/openid.go
kolaente: Great solution!

Issue #807, opened 2025-05-21
Chinese name from oauth will cause login failure


maggch97:
### Description

When I login with oauth and my account name is Chinese, there's no error message but go back to vikunja login page.
I repro this issue with local db and postgres, will do more debug and update more info here.

### Vikunja Version

latest unstable

### Can you reproduce the bug on the Vikunja demo site?

Please select

maggch97:
https://deepwiki.com/search/when-the-users-name-is-in-chin_42705e93-e257-41af-acea-18c8154c797a

The answer from deepwiki looks correct

The solution is to correct the regex replacement patterns in frontend/src/stores/auth.ts:
```
const base64 = jwt  
  .split('.')[1]  
  .replace(/-/g, '+')  // Correct regex syntax  
  .replace(/_/g, '/')  // Correct regex syntax
```

kolaente:
Duplicate of https://github.com/go-vikunja/vikunja/issues/388

maggch97:
> Duplicate of [#388](https://github.com/go-vikunja/vikunja/issues/388)

The fix of that issue has a mistake in regex syntax. You can check above answer from deepwiki. Please reopen this issue, I will push a fix later

PR #809, opened 2025-05-21
fix(auth): fix regex in JWT decoding causing login issues for Chinese/Japanese user names

maggch97:
This PR fixes https://github.com/go-vikunja/vikunja/issues/807 and I verified locally. It has no problem now.


kolaente (approved):
Thanks!

PR #813, opened 2025-05-22
feat(ui): Experimental redesign of Task View for improved structure and clarity

maggch97:
This PR introduces an experimental redesign of the Task View UI, aiming for a clearer and more structured layout, similar to common project management tools.

Key changes include a two-column layout, tabbed content for comments/attachments/related tasks, and removal of dynamic module loading in favor of a fixed structure.

I'm sharing this early for your feedback on this direction, especially considering Vikunja's diverse user base. One thought is whether a simpler view could co-exist if this more structured approach is deemed useful. **Perhaps this new, more complex view could be an optional feature, enabled via `config.yml`?**

Looking forward to your thoughts!

image
image
image


maggch97:
Make more changes
image


Add a new other tab, contains color, reminders and repeat because I didn't find a good place for them, and they are not commonly used by enterprise users
image


Show name after avatar, no overlap animation, make it easy to identity the user
image


Add an edit button to Assignees and Labels, only show editor after clicking edit buttons for all users
image


Add a move button to the right of project name, make the moving operation more intuitive
image
image

kolaente:
Hey thanks for the PR and the effort put into this!

To be honest, I'm really unsure about this change. It's not something I would merge in this state, but it's good to have something to discuss (and you seem to recognize this by marking the PR as draft).

My biggest issue is how this changes the general "feel" of Vikunja from a light to-do list to a big, clunky enterprise tool like Jira. That's not bad per se, but I think it is too much at once. Another thing is the small details like wrong spacing around icons and such, but that's easy to change.

> Perhaps this new, more complex view could be an optional feature, enabled via config.yml? This would, of course, add to the overall maintenance cost for the Task View

I don't think we should make this an option, since, as you said, it will increase the maintenance burden.

> they are not commonly used by enterprise users

These "non-enterprise" users make up a good chunk of Vikunja's users. That's not easy to disregard.

I feel like we need to take a step back here and make a concept first. What prompted you to make these changes? Do you have other applications which you've used as a base? What do you think should Vikunja "feel" like?

maggch97:
@kolaente I totally agree with you, I don't want this PR to be merged recently. I just want to use this PR to start a discussion about UI refactoring.

Before I found Vikunja, I tried several open source project management software, but I didn't use them because of some problems. I was looking for a project management software for a small team. When I found Vikunja, I found that it was an application that called itself a todo app, but it had many functions that only enterprises need, such as ldap, oauth, and perfect multi-person collaboration. I found that Vikunja almost completely met the needs of our small team.

But after checking the community, I found that many people use it as a to-do app. I think this is also the biggest difficulty of refactoring. The jira-like UI is too complicated for personal use.

Since most of users are todo users, the best way at present for me is to maintain the content of this PR as a fork. I'm amazed that Vikunja is almost entirely maintained by a single person. The code quality, documentation, and CI/CD pipeline are truly impressive. While it may have started as a simple todo application, among all the similar projects I've come across, I believe it can become the best open source project management software.

kolaente:
Thanks for the kind words!

> I found that Vikunja almost completely met the needs of our small team.

What do you think is missing, apart from the changes in the task detail view?

> The jira-like UI is too complicated for personal use.

That's exactly the struggle. 

I won't say the UI will always stay as it is today, but I probably won't change it in one go. I did think about it as well, but haven't found a good way to improve it yet. So, small changes it is.

kolaente:
> Other things like @mention users in comments

This is actually working, there's just no highlight in the editor for it.

kolaente:
I'll close this one for now as the direction of this redesign is not something that we'll do anytime soon. If you want to discuss redesigning the UI in general, feel free to open a new issue where we can talk about this.

PR #817, opened 2025-05-22
Fix/button boolean props

dpschen:


kolaente (commented):
I'm really unsure about this. I think the biggest thing here is consistency - why have some properties called `no-*` and others `has-*`? (for example `no-shadow` and `has-content`) I think we should have only ones with `has` and have good defaults. That's what we're doing with configuration as well, I don't like mixing these together.

kolaente:
I think it should be consistent and clear. IMHO having some properties with `no-` and some with `has-` prefixed makes it inconsistent.

What do you think about rather making everything explicit, so that all props can only be `false` by default and we'd need to set them to `true` everywhere?

dpschen:
> What do you think about rather making everything explicit, so that all props can only be false by default and we'd need to set them to true everywhere?

Unsure what this would look like, since this is exactly what I tried with my change. See the updated default props in the Button and Card components.

kolaente:
> Unsure what this would look like, since this is exactly what I tried with my change. See the updated default props in the Button and Card components.

I mean basically not having the default `true` anywhere (enabling the lint rule and then enforcing it) and then setting each affected prop to `true` manually.

kolaente:
I think I just don't like properties prefixed with `no`. Feels like a hack around the linter, instead of making it explicit and setting the prop everywhere.

PR #818, opened 2025-05-22
Feat/unify x button component name

dpschen:
~~Depends on https://github.com/go-vikunja/vikunja/pull/817~~

@kolaente I rebased this, so that it's not dependent on that PR anymore — should have been like that from the start, sry!


kolaente (approved):
Thanks!

PR #819, opened 2025-05-22
feat: duplicate tasks

HornedHeck:
### Why

I'm missing templates functionality. Since copying of task can be used as partial substitute and this feature does not need a lot of efforts, I decided to implement it.

### Feature requirements

1. As a user I want to create copies of task to simplify common flows.
   - Basic information about the task should be copied: title, description, progress, time constraints.
   - Related sub-tasks should be copied with all related fields and added as a sub-task to the duplicate.
   - Since this feature is positioned as a substitute for templates, do NOT add `Copied To` or `Copied From` relations to any of the tasks to avoid unneeded relations.
   - Do not copy fields that may harm UX: subscription, favorite status, reminders.
   - Do not copy attachments and comments, investigate later, if it is possible to link them.
2. As a user I want to open newly create task to make appropriate changes.
   - New task should be opened in the same style (modal or not modal) as the original task.
   - Navigation back from the new task should NOT return to the  original task but to the entry before it (e.g. to the Kanban board).

### Changes summary

1. Added duplicate task button to TaskDetailView.vue. This buttons creates a partial deep copy of the task that includes information about title, description, status, priority, time constrains, assignees, repeats, labels, project, sub-tasks. After process finished, details view for the new task is opened.
2. Added translations to accommodate UI changes.

### Screenshots
image

### P.S.
I am not very experienced with Vue and TypeScript, so feel free to propose related improvements and suggestions.


maggch97:
Not for this PR.

Currently the TaskDetailView has a simple ui pattern with all action buttons on the left. This works fine when this project is simple. But after more and more actions and components are added, it becomes difficult for users to read and find the required functions.

I strongly agree with the duplicate button. In my use case, this button is very useful for duplicating a task with a lot of tags and the description template. However, most project management softwares put this function in a menu.

kolaente:
Hey @HornedHeck thanks for the PR!

I think we can add the duplicate button here, the restructuring is a general thing and should be a different PR.

frontend/src/i18n/lang/ar-SA.json
```diff
@@ -600,7 +600,8 @@
 				"color": "تحديد اللون",
 				"delete": "حذف",
 				"favorite": "أضف إلى التفضيلات",
-				"unfavorite": "حذف من التفضيلات"
+				"unfavorite": "حذف من التفضيلات",
```
kolaente: Please do not add translations here, they will be synced from Crowdin. Only add the new source string in `en.json`.

Check out the docs: https://vikunja.io/docs/translations/

frontend/src/stores/tasks.ts
```diff
@@ -494,6 +494,38 @@ export const useTaskStore = defineStore('task', () => {
 		}
 	}
 	
+	async function duplicateTask(task : ITask) : Promise<ITask> {
```
kolaente: I think this should happen in the api layer. That will make it a lot easier to duplicate attachments and other related items as well.

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -946,6 +953,43 @@ function setRelatedTasksActive() {
 		}
 	}
 }
+import TaskRelationService from '@/services/taskRelation'
+const taskRelationService = shallowReactive(new TaskRelationService())
+
+import { RELATION_KINDS, RELATION_KIND, type IRelationKind } from '@/types/IRelationKind'
+import TaskRelationModel from '@/models/taskRelation'
```
kolaente: Please put the imports at the top.

frontend/src/stores/tasks.ts
HornedHeck: Thank you for the suggestion!

frontend/src/models/taskDuplicateModel.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default class TaskDuplicateModel extends AbstractModel<ITaskDuplicate> implements ITaskDuplicate {
+	taskId = 0
+	projectId = 0
+	duplicatedTask: ITask | null = null
+
+	constructor(data: Partial<ITaskDuplicate>) {
+		super()
+		this.assignData(data)
+		this.duplicatedTask = this.duplicatedTask ? this.duplicatedTask : null
```
kolaente: Is this required?

kolaente: Could also be simplified

```suggestion
		this.duplicatedTask = this.duplicatedTask ?? null
```

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -932,7 +939,7 @@ async function removeRepeatAfter() {
 	task.value.repeatAfter.amount = 0
 	task.value.repeatMode = TASK_REPEAT_MODES.REPEAT_MODE_DEFAULT
 	await saveTask()
-}
+}~
```
kolaente: This seems like it should not be here

```suggestion
}
```

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -946,6 +953,18 @@ function setRelatedTasksActive() {
 		}
 	}
 }
+
+async function duplicateTask() {
+	const clone = await taskStore.duplicateTask(task.value)
+	if (clone != null) {
```
kolaente: What could be a case where the clone returned is `null`? Can we handle that explicitly?

And please use `!==`:
```suggestion
	if (clone !== null) {
```

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+// Create duplicates of a task
+// @Summary Duplicate an existing task
+// @Description Copies the task, assignees, lables, subtasks from one task to a new one. The user needs read and write access in the project.
+// @tags task
+// @Accept json
+// @Produce json
+// @Security JWTKeyAuth
+// @Param taskID path int true "The task ID to duplicate"
+// @Param projectID path int true "The project ID of task to duplicate"
+
```
kolaente: This empty line should be removed

```suggestion
```

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+// @Success 201 {object} models.TaskDuplicate "The created task."
```
kolaente:
```suggestion
// @Success 201 {object} models.TaskDuplicate "The duplicated task."
```

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+// @Failure 400 {object} web.HTTPError "Invalid task duplicate object provided."
+// @Failure 403 {object} web.HTTPError "The user does not have access to the project."
+// @Failure 500 {object} models.Message "Internal error"
+// @Router /projects/{projectID}tasks/{taskID}/duplicate [put]
+func (td *TaskDuplicate) Create(s *xorm.Session, doer web.Auth) (err error) {
+	// Helper to copy a single task (without subtasks)
+	copyTask := func(orig *Task) *Task {
```
kolaente: Why the function?

You could either do it without the function or use `mergo` (third party package already used in other places).

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+			PercentDone: orig.PercentDone,
+		}
+	}
+
+	// Map from old task ID to new task pointer
+	idMap := map[int64]*Task{}
+
+	// Recursively duplicate a task and its subtasks
+	var duplicateTaskTree func(parentID, origID int64) (*Task, error)
+	duplicateTaskTree = func(parentID, origID int64) (*Task, error) {
```
kolaente: This function seems redundant as well, can you just inline it?

I feel like this would be clearer if you'd duplicate the original task first and then the related tasks.

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+		// Insert the new task
+		err = newTask.Create(s, doer)
+		if err != nil {
+			return nil, err
+		}
+		idMap[origTask.ID] = newTask
+
+		// Duplicate subtasks recursively
+		relations := []*TaskRelation{}
+		err = s.Where("task_id = ? AND relation_kind = ?", origTask.ID, RelationKindSubtask).Find(&relations)
```
kolaente: Why not duplicate all related tasks?

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+			return nil, err
+		}
+		err = addMoreInfoToTasks(s, map[int64]*Task{origTask.ID: &origTask}, doer, nil, nil)
+		if err != nil {
+			return nil, err
+		}
+		// Copy the task
+		newTask := copyTask(&origTask)
+		newTask.ProjectID = origTask.ProjectID
+		newTask.ID = 0 // Ensure new insert
```
kolaente: This *should* not be required, otherwise that's a bug in the task create function

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return nil
+}
+
+// CanCreate checks if a user has the right to duplicate a task
+func (td *TaskDuplicate) CanCreate(s *xorm.Session, a web.Auth) (canCreate bool, err error) {
+	// project Exists + user has read and write access to project
+	project := &Project{ID: td.ProjectID}
+	canRead, _, err := project.CanRead(s, a)
```
kolaente: If the user has write access to a project, they can also read it, no need to check read access separately.

pkg/models/task_duplicate.go
```diff
@@ -0,0 +1,166 @@
...
+// @Security JWTKeyAuth
+// @Param taskID path int true "The task ID to duplicate"
+// @Param projectID path int true "The project ID of task to duplicate"
+
+// @Success 201 {object} models.TaskDuplicate "The created task."
+// @Failure 400 {object} web.HTTPError "Invalid task duplicate object provided."
+// @Failure 403 {object} web.HTTPError "The user does not have access to the project."
+// @Failure 500 {object} models.Message "Internal error"
+// @Router /projects/{projectID}tasks/{taskID}/duplicate [put]
+func (td *TaskDuplicate) Create(s *xorm.Session, doer web.Auth) (err error) {
```
kolaente, line 51: I think this whole function could be refactored so that it works more like this:

1. Fetch the task to be duplicated
2. Create a copy of it
3. Copy all reminders
4. Copy all attachments
5. Copy all assignees
6. Copy other attributes in separate tables I've missed
7. For each related task, call the function recursively so that all attributes of these tasks are duplicated as well
8. Once the related task is duplicated, re-create the relation

Check the project duplication process, since we're duplicating all tasks with all attributes there as well.

pkg/models/task_duplicate_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTaskDuplicate_DeepCopy(t *testing.T) {
+	s := setupTestSession(t)
+	u := &user.User{ID: 1, Username: "testuser"}
+	auth := web.NewAuthUser(u)
+
+	// Create a root task
```
kolaente: Can you use an existing fixture here?

kolaente (changes requested):
Already better!

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -946,6 +953,18 @@ function setRelatedTasksActive() {
...
+		router.replace({
```
dpschen: I think this should be `router.push`, since it's a new task.

frontend/src/modelTypes/ITaskDuplicate.ts
```diff
@@ -0,0 +1,8 @@
+import type {IAbstract} from './IAbstract'
+import type {ITask} from './ITask'
+
+export interface ITaskDuplicate extends IAbstract {
+	taskId: number
+	projectId: number
+	duplicatedTask: ITask | null
+}
```
dpschen, line 8: @kolaente: First I wanted to note that the final new line is missing in the three new ts files, but then I realised that we have `insert_final_newline = false` in the `.editorconfig` file. Without final empty line GitHub displays it like this:

Screenshot 2025-05-23 at 11 04 06

kolaente: uhhh that should not be the case, I've changed it in dc6ef8b4be000b3dfe50c0e71e5009d764d31ef6

frontend/src/views/tasks/TaskDetailView.vue
HornedHeck: If the backed will return error, then clone will be null. There are error message displayed, so I do not think that it requires additional handling.

frontend/src/views/tasks/TaskDetailView.vue
HornedHeck: It was initially. However replace provides better experience when it comes to navigation. At least from my point of view. I do not have a desire to defend it until my death, so I'll leave the option that @kolaente  chooses.

dpschen: Can you explain, what made you feel, that this makes the navigation experience better?

kolaente: `router.push` seems like the better option since that means browser navigation will work.

kolaente:
Please also fix the lint and test errors

pkg/models/task_duplicate.go
HornedHeck: I am little bit confused with you comment.
The function already works very similar to the way you described, except skiping pp. 3,4 and 5. I will implement reminders copy (my miss), and p.5 seems to be excessive (assignee is copied during task creation successfully). 

What concerns p.4, I strongly against copying files and instead it will be better to link them to new request. Withing the existing implementation it seems to me to be impossible and changes to change the situation will be better to do in separate PR.

pkg/models/task_duplicate.go
HornedHeck: As far as I understand different relation kinds, it is not needed to duplicate most of the tickets with relations other than subtasks.
Relates, blocks, after, before will usually refer to some side features. E.g. between a story and a bug that blocks is or appeared during development. In this case new story should to be related to this bug, because it most likely will not be related to the same problem. The same goes for sub-tasks. 
The only relation that is missing and that I think worth to add is relation between task and its parent.

pkg/models/task_duplicate.go
HornedHeck: One more note regarding inspiration from project duplication. It is slightly different task. In the project duplication back always creates only one project, because there no such concept as nested projects. Meanwhile task can contains sub-tasks which I want to copy as wall.

HornedHeck:
> Please also fix the lint and test errors

Lints have been fixed. What concerns tests, there is password-reset cypress test is failing. This part is not relate to my changes and taking into account the fact, that it failed in only 1 of 3 runs, I think there are some problems with stability in this part of test cases.

frontend/src/services/taskDuplicateService.ts
kolaente, line 1: Please remove this empty file.

pkg/models/task_duplicate.go
kolaente: > The function already works very similar to the way you described, except skiping pp. 3,4 and 5. I will implement reminders copy (my miss), and p.5 seems to be excessive (assignee is copied during task creation successfully).

Reminders seem to be handled during task creation as well.

Can you add a test assertion for that?

> What concerns p.4, I strongly against copying files and instead it will be better to link them to new request. Withing the existing implementation it seems to me to be impossible and changes to change the situation will be better to do in separate PR.

That is correct, but not handling task attachments does not feel like a good solution either. Linking existing files to new attachments has it's own set of problems which need to be handled (what happens when the original is deleted? Should the link still exist? Could that lead to "ghost" files?)

> One more note regarding inspiration from project duplication. It is slightly different task. In the project duplication back always creates only one project, because there no such concept as nested projects. Meanwhile task can contains sub-tasks which I want to copy as wall.

There is a `duplicateTasks` function which duplicates all tasks with all task details. That could be adjusted so that it is possible to duplicate a single task (or the new duplicate task functionality could be called from there).

HornedHeck: I can suggest the following solution for attachments:
1. When user adds an attachment to task, create a file and make a link from this file to the task.
2. When duplicating the task (or any attachment containing entity), only copy the link.
3. When deleting any link, check if there are other links for this file. If no, then delete it.

HornedHeck: `duplicateTasks` in `project_duplicate.go` copies all the information about the task, while I want to exclude certain fields from the copying. Now I see only one way to join them: to introduce some kind of logic in the function, that will control the copying behavior.  I do not think that this is a solution, that worth implementing.

frontend/src/modelTypes/ITaskDuplicate.ts
dpschen: @HornedHeck the final new line is still missing

pkg/models/task_duplicate.go
kolaente: I would take one step back: Which fields do you want to exclude? And why? Maybe that's not required?

I think we should re-use as much code as we can. Similar code in different places doing almost the same thing is never a good idea.

kolaente:
Are you still interested in working on this? Do you need more information from me? If so, can you tell me where exactly you're missing something?

malys:
@HornedHeck  this PR is abandoned?

Issue #816, opened 2025-05-22
Feature Request: Sync avatar from OpenID providers


maggch97:
### Description

Latest unstable vikunja does not allow the OpenID user to change name because it's from provider, but avatar is allowed to be changed. I check the code, LDAP user's avatar is auto synced.
Would it be ok if I open a PR to do same thing for OpenID users, auto sync the avatar from provider?

### Vikunja Version

unstable

### Can you reproduce the bug on the Vikunja demo site?

Please select

maggch97:
> Sure, would love a PR!

I would like to discuss the implementation. The OpenID provider will return a picture url when a user login, here we have 3 method to deal with the url
1. Download image from the url, then process like ldap. To avoid download a same image multiple times, we need to add a new filed `externalAvatarUrl`. This method will cost more time to login.
2. Just save the picture url in a new field `externalAvatarUrl`, then redirect /api/v1/avatar to the external url. This method reduces the server load and does not cause any performace issue. But we cannot return an image of the specified size. 
3. Based on 2, We can add a new avatar API that does not allow for specifying the size and returns externalAvatarUrl. But for an avatar, this is a bit overly complicated.

Which method do you prefer?

kolaente:
Don't we always need to fetch the image to check if it changed? I wouldn't assume the url changes when the avatar changes.

Another option would be a cache of sorts. But in general, since we're talking about two servers talking to each other, I'd assume fetching the image is fast and does not incur much overhead. Did you check how much the overhead is?

maggch97:
> Don't we always need to fetch the image to check if it changed? I wouldn't assume the url changes when the avatar changes.
> 
> Another option would be a cache of sorts. But in general, since we're talking about two servers talking to each other, I'd assume fetching the image is fast and does not incur much overhead. Did you check how much the overhead is?

No, I didn't observe any performace issue with . If you don't have any concerns about this [implementation ](https://github.com/go-vikunja/vikunja/pull/821), I'm fine with it too. We can optimize it later if we get any user feedback about performance issues.

kolaente:
If there are no real performance implications, I think we're good with changing this later if we see problems in the wild.

PR #821, opened 2025-05-22
feat(auth): sync avatar from OpenID providers

maggch97:
For https://github.com/go-vikunja/vikunja/issues/816

This PR is still working in progresss.
The current change will download the same image every time.

### DIFF
--- a/pkg/modules/auth/openid/openid.go
+++ b/pkg/modules/auth/openid/openid.go
@@ -17,6 +17,7 @@
 package openid
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -28,7 +29,9 @@ import (
 	"code.vikunja.io/api/pkg/log"
 	"code.vikunja.io/api/pkg/models"
 	"code.vikunja.io/api/pkg/modules/auth"
+	"code.vikunja.io/api/pkg/modules/avatar/upload"
 	"code.vikunja.io/api/pkg/user"
+	"code.vikunja.io/api/pkg/utils"
 	"code.vikunja.io/api/pkg/web/handler"
 
 	"github.com/coreos/go-oidc/v3/oidc"
@@ -68,6 +71,7 @@ type claims struct {
 	PreferredUsername string                   `json:"preferred_username"`
 	Nickname          string                   `json:"nickname"`
 	VikunjaGroups     []map[string]interface{} `json:"vikunja_groups"`
+	Picture           string                   `json:"picture"`
 }
 
 func init() {
@@ -259,6 +263,8 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
 		fallbackMatchFound = err == nil // found if no error, not found if we reach it here despite an error
 	}
 
+	var isNewUser bool = false
+
 	if !alreadyCreatedFromIssuer && !fallbackMatchFound {
 
 		// If no user exists, create one with the preferred username if it is not already taken
@@ -270,7 +276,14 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
 			Issuer:   idToken.Issuer,
 			Subject:  idToken.Subject,
 		}
-		return auth.CreateUserWithRandomUsername(s, uu)
+		if cl.Picture != "" {
+			uu.AvatarProvider = "openid"
+		}
+		u, err = auth.CreateUserWithRandomUsername(s, uu)
+		if err != nil {
+			return nil, err
+		}
+		isNewUser = true
 	} else if alreadyCreatedFromIssuer {
 
 		// try updating user.Name and/or user.Email if necessary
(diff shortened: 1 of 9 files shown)


pkg/modules/auth/openid/openid.go
```diff
@@ -68,6 +71,7 @@ type claims struct {
 	PreferredUsername string                   `json:"preferred_username"`
 	Nickname          string                   `json:"nickname"`
 	VikunjaGroups     []map[string]interface{} `json:"vikunja_groups"`
+	Picture           string                   `json:"picture"`
```
kolaente, line 75: Is there a standard for what this is called? If not, this attribute name should be configurable.

pkg/modules/auth/openid/openid.go
```diff
@@ -280,13 +293,44 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
...
+			u.AvatarProvider = "openid"
+		}
 		u, err = user.UpdateUser(s, u, false)
 		if err != nil {
 			return nil, err
 		}
 	}
 
-	return
+	if cl.Picture != "" && (isNewUser || u.AvatarProvider == "openid" || u.AvatarProvider == "") {
```
kolaente: Shouldn't we always check this, regardless of `isNewUser`?

pkg/modules/auth/openid/openid.go
```diff
@@ -280,13 +293,44 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
...
+		log.Debugf("Found avatar URL for user %s: %s", u.Username, cl.Picture)
+
+		// Download avatar
+		avatarData, err := utils.DownloadImage(cl.Picture)
+		if err != nil {
+			log.Errorf("Error downloading avatar: %v", err)
```
kolaente: Maybe log the user id as well here so that the admin could investigate.

pkg/modules/auth/openid/openid.go
```diff
@@ -280,13 +293,44 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
...
+			// Continue, do not interrupt the authentication process due to avatar download failure
+		} else {
```
kolaente: Can you return early instead of the `else`?

pkg/modules/auth/openid/openid.go
```diff
@@ -280,13 +293,44 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
...
+			// Process avatar, ensure 1:1 ratio
+			processedAvatar, err := utils.CropAvatarTo1x1(avatarData)
+			if err != nil {
+				log.Debugf("Error processing avatar: %v", err)
+				// Continue, do not interrupt the authentication process due to avatar processing failure
+			} else {
```
kolaente: Same here about the early return

pkg/routes/api/v1/user_settings.go
```diff
@@ -32,7 +32,7 @@ import (
 
 // UserAvatarProvider holds the user avatar provider type
 type UserAvatarProvider struct {
-	// The avatar provider. Valid types are `gravatar` (uses the user email), `upload`, `initials`, `marble` (generates a random avatar for each user), `default`.
+	// The avatar provider. Valid types are `gravatar` (uses the user email), `upload`, `initials`, `marble` (generates a random avatar for each user), `ldap` (synced from LDAP server), `openid` (synced from OpenID provider), `default`.
```
kolaente, line 35: Good catch!

kolaente (changes requested):
This will also need an update to the docs once merged.

pkg/modules/auth/openid/openid.go
maggch97: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

`picture` is contained in OIDC standard, no need to be configurable

pkg/modules/auth/openid/openid.go
maggch97: return u, nil will immediately end the function, skipping any future code. Current design allows adding more code after avatar sync and will not be affected be avatar sync failure

pkg/modules/auth/openid/openid.go
```diff
@@ -270,7 +274,15 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
 			Issuer:   idToken.Issuer,
 			Subject:  idToken.Subject,
 		}
-		return auth.CreateUserWithRandomUsername(s, uu)
+		if cl.Picture != "" {
+			// This will not take effect because it will be overridden by the default provider
```
kolaente: If it has no effect, why add it?

pkg/modules/auth/openid/openid.go
```diff
@@ -280,13 +292,46 @@ func getOrCreateUser(s *xorm.Session, cl *claims, provider *Provider, idToken *o
 		if cl.Name != u.Name {
 			u.Name = cl.Name
 		}
+		if cl.Picture != "" && u.AvatarProvider == "" {
```
kolaente: Doesn't this mean if the user has already set their avatar to something else, their avatar from the openid provider won't get synced? Since `u.AvatarProvider` will be empty in that case.

pkg/modules/auth/openid/openid.go
kolaente: I mean, that's a problem when we're adding code after this in the future. After thinking about it, can you refactor this part out into a separate function? Then it should be no problem to return early.

pkg/modules/auth/openid/openid.go
maggch97: This line of code also has no effect, since the provider field will always have a default value and should never be empty.

The goal of my change is to avoid overwriting the avatars of OpenID users who have previously changed them.
Only new users or those who still use the default initials avatar will be affected.

If default avatar provider in config.yml is updated to another provider, all OpenID users in that instance will not be affected by this change.

I want to minimize impact on existing user data. Additionally, because of EmailFallback, we cannot guarantee that a current OpenID user signed up using OpenID

I will push a fix later

kolaente: > The goal of my change is to avoid overwriting the avatars of OpenID users who have previously changed them

Why? Shouldn't the provider always take precedence over what the user wants? That would be in line with not being able to change the avatar later in Vikunja.

> If default avatar provider in config.yml is updated to another provider, all OpenID users in that instance will not be affected by this change.

The default settings are only taken into account during user registration, existing accounts are never affected by a change there.

maggch97: Done, the provider check is removed

pkg/modules/auth/openid/openid.go
```diff
@@ -215,6 +219,39 @@ func getTeamDataFromToken(groups []map[string]interface{}, provider *Provider) (
 	return teamData
 }
 
+// trySyncUserAvatarFromOpenID attempts to download and store a user's avatar from an OpenID provider
+func trySyncUserAvatarFromOpenID(s *xorm.Session, u *user.User, pictureURL string) {
```
kolaente: Why try? Why not omit the `try`?

maggch97: Yes, it's better to let the caller handle the error itself. Updated.

PR #834, opened 2025-05-28
feat: allow sending reminders via webhooks

lelemm:
This PR creates two new cron jobs that check for reminders (normal reminder and overdue) just like the normal reminder cron, but instead of saving the information in the database or sending an email, it tries to call webhooks.

I'm unsure if this is a desirable feature, but it makes creating integrations with gotify and n8n easier.

### DIFF
--- a/pkg/models/user_webhook_setting.go
+++ b/pkg/models/user_webhook_setting.go
@@ -81,6 +81,28 @@ func AvailableWebhookNotificationTypes() []WebhookNotificationType {
 		{Type: "all", Description: "General webhook for all notification types (fallback)"},
 		{Type: "task.reminder", Description: "Task reminder notifications"},
 		{Type: "task.undone.overdue", Description: "Overdue task notifications"},
+		// Task notifications
+		// {Type: "task.comment", Description: "Task comment notifications"},
+		// {Type: "task.assigned", Description: "Task assigned notifications"},
+		// {Type: "task.deleted", Description: "Task deleted notifications"},
+		// {Type: "task.mentioned", Description: "User mentioned in task notifications"},
+		// Project notifications
+		// {Type: "project.created", Description: "Project created notifications"},
+		// Team notifications
+		// {Type: "team.member.added", Description: "Team member added notifications"},
+		// Data export notifications
+		// {Type: "data.export.ready", Description: "Data export ready notifications"},
+		// Migration notifications
+		// {Type: "migration.done", Description: "Migration completed notifications"},
+		// {Type: "migration.failed", Description: "Migration failed notifications"},
+		// {Type: "migration.failed.reported", Description: "Migration failure reported notifications"},
+		// User account notifications
+		// {Type: "totp.invalid", Description: "Invalid TOTP code notifications"},
+		// {Type: "password.account.locked.after.invalid.totp", Description: "Account locked after invalid TOTP notifications"},
+		// {Type: "failed.login.attempt", Description: "Failed login attempt notifications"},
+		// {Type: "user.deletion.confirm", Description: "Account deletion confirmation notifications"},
+		// {Type: "user.deletion", Description: "Account deletion scheduled notifications"},
+		// {Type: "user.deleted", Description: "Account deleted notifications"},
 	}
 }
 


kolaente:
> Instead of having two cron jobs, would it be better to change how these validations work?

Yes, these would need to be changed.

> I saw you saying about a new table where the user can set their webhooks instead of having the project webhook. Do you think this mandatory for this PR?

If you start with webhooks as a notification target here, the notification target should be a column in the users table, similar to how other settings are stored. If you think it should be possible to have multiple webhooks as notification targets, it should be a new table instead.

>  But if you prefer, I can add a new field to hold a user webhook and trigger it whenever we detect that the user is involved in that notification event.

I think it would be better to have this as a per-user setting, since the notifications are also a per-user thing.

> How would prefer to have this information saved?

The `webhooks` table already has a `created_by_id` column. I would use that and make the `project_id` nullable. Then it's a matter of filtering for the event and `created_by_id`.

lelemm:
Added the webhook URL to the user profile:
image

Now it triggers based on the user webhook. If the user doesn't have a webhook enabled or url is empty, it will skip the user

lelemm:
@kolaente , when you have time. See if anything else is needed here

```
VIKUNJA_SERVICE_ENABLEWEBHOOKREMINDERS=true ./vikunja
```
I added new config variable to enable this

pkg/models/task_reminder.go
```diff
@@ -243,12 +243,15 @@ func getTasksWithRemindersDueAndTheirUsers(s *xorm.Session, now time.Time) (remi
 				tzs[u.User.Timezone] = tz
 			}
 
-			actualReminder := r.Reminder.In(tz)
-			if (actualReminder.After(now) && actualReminder.Before(now.Add(time.Minute))) || actualReminder.Equal(now) {
+			actualReminder := r.Reminder.In(tz).Truncate(time.Minute)
```
lelemm, line 324: Btw, I changed how the reminder is calculated because it was missing some reminders for me while debugging. Truncating the date to remove the seconds made it more reliable.

kolaente:
Please make the webhook a notification target (similar to the existing mail or database targets). I'm afraid your current implementation makes things more complicated and harder to maintain.

lelemm:
@kolaente, I’m reviewing your request to use the notify method and add the webhook as a notification target. However, I noticed that the `notify` method is used in many places:

image

Instead of creating a separate option for "Send me reminders for tasks via webhook," I will implement a feature called "Enable webhook notifications." For now, I will check whether the notification is for a regular reminder or an overdue reminder. If it is, the webhook will be triggered. 

This approach will allow for greater flexibility when adding new webhook notifications in the future.

Does that sound good to you?

kolaente:
Sounds great! People have asked for getting general notifications via Apprise or Gotify in the past as well, having a webhook as notification target makes this easier.

ubergeek77:
Thanks so much for developing this! I know it's been a while, and this is still in development, but I really needed this feature quickly, so I gave this PR branch a try.

It works well for the most part! It's been reliably sending webhooks for half of this week. However, I had one webhook fail to fire today. n8n has no record of it, and since Vikunja doesn't log anything webhook related*, I wasn't able to diagnose it further. I can't see if the webhook was even sent, so I don't know if the cron randomly fails after high uptimes, or if it was a network interruption and Vikunja didn't attempt to retry sending it (my webhook handler sends a success response on receipt).

\* _Not that I could see anyway, the Vikunja Docker logs only show HTTP requests_

I thought I would mention this while this is still in development, maybe some logging and webhook retry settings could be added as well.

Thanks for your efforts on this! From what I can see, this has been a requested feature for many, many years!

lelemm:
@ubergeek77 , I've been VERY busy at work. Feel free to catch this PR from this point if you have time

lelemm:
At some point I will revisit this PR but I'm flooded with work lol

lelemm:
@kolaente , almost a year later, I refactored this.

lelemm:
new menu:

image

image
image

frontend/src/models/userSettings.ts
```diff
@@ -25,11 +25,13 @@ export default class UserSettingsModel extends AbstractModel<IUserSettings> impl
 		quickAddMagicMode: PrefixMode.Default,
 		colorSchema: 'auto',
 		allowIconChanges: true,
+		filterIdUsedOnOverview: null,
 		defaultView: DEFAULT_PROJECT_VIEW_SETTINGS.FIRST,
 		minimumPriority: PRIORITIES.MEDIUM,
 		dateDisplay: DATE_DISPLAY.RELATIVE,
 		timeFormat: TIME_FORMAT.HOURS_24,
 		defaultTaskRelationType: RELATION_KIND.RELATED,
+		backgroundBrightness: null,
```
kolaente: This seems like a rebase error

kolaente:
I've taken another look and I think it should reuse more of the existing webhooks infrastructure. Right now this PR re-introduces a lot of the same patterns but only partially, which makes this duplicate code and harder to maintain.

I'll rework this in two stages:

1. Add reminder events to the project-based webhooks - that's done in https://github.com/go-vikunja/vikunja/pull/2291
2. Allow users to configure these webhooks only for their account instead of per-user. This will then give a similar UX to what is proposed here in this PR.

kolaente:
Will be closed by https://github.com/go-vikunja/vikunja/pull/2295

lelemm:
Sorry to disappear again. Work and life are consuming me hard lol

kolaente:
No worries!

kolaente:
https://github.com/go-vikunja/vikunja/pull/2295 is done and merged, please check the next unstable build with the changes.

PR #845, opened 2025-06-01
fix: hide empty sidebar navs

dpschen:


kolaente:
Failure seems unrelated.

PR #848, opened 2025-06-01
feat(ci): pass cypress ci build id

dpschen:


kolaente:
How can I check if this works?

kolaente:
Seems like the secret variables are not passed to the runner:

```
Warning: It looks like you are trying to record this run from a forked PR.

The Record Key is missing. Your CI provider is likely not passing private environment variables to builds from forks.

These results will not be recorded.

This error will not affect or change the exit code.

dpschen:
> Seems like the secret variables are not passed to the runner:

I created a fix for that: https://github.com/go-vikunja/vikunja/pull/860

I think it should work now. I read in cypress docs, that you need to pass the ci build id in order for parallel workers to work effectively. From the timings it seems like 4 workers also do now the work of the 5 that we needed previously. The latter change was recommended in the cypress UI. For this I looked at a run of a merged PR, since as you correctly pointed out I can't record runs from PRs that are coming from forks.

Would be cool if I could add new branches in the main repo. I also saw that there are some old branches from me here that I can not force push to to update them

PR #860, opened 2025-06-04
Feat/cypress dont record forks

dpschen:


kolaente:
Seems like this does not work as expected

dpschen:
Yes, unsure if it would help if I would set all those flags conditionally or if I always need record.

dpschen:
~~@kolaente I think this should work now~~

Will never say this before tests running through :D

kolaente:
Since we've removed Cypress, this is no longer relevant.

PR #862, opened 2025-06-04
feat: mini tiptap improvements

dpschen:
The TipTap editor now uses Vue’s defineModel for its reactive value, simplifies editability control with watchEffect, and correctly types DOM image elements.

Summary

Imported watchEffect and switched to defineModel for the editor’s v-model, removing the old update:modelValue emit

Typed the retrieved image element as HTMLImageElement and removed debugging output from the paste handler

Replaced the watcher controlling editability with a simpler watchEffect and updated model handling to use the new model ref


dpschen:
> Frontend tests seem to fail now

Rebase error… Will check again

PR #863, opened 2025-06-04
ci: remove concurrency from test workflow

dpschen:


kolaente:
So pushes to main won't cancel previous runs but pushes to the same pr will?

dpschen:
Yes

PR #870, opened 2025-06-04
CI: remove duplicate sqlite-in-memory integration

dpschen:
## Summary
- change `test-api` matrix to run unit tests only with sqlite-in-memory
- run integration tests across all existing databases
- add `test-api-success` summary job so required checks can still pass
- remove sqlite-in-memory duplicate from integration job

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -162,20 +162,40 @@ jobs:
           chmod +x vikunja
           ./vikunja migrate
   
-  test-api:
+  test-api-unit:
+    runs-on: ubuntu-latest
+    needs:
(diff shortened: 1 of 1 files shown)


kolaente (commented):
unit and integration tests do not cover the same things. We should run both with all databases.

The real problem here is the wording, `unit` isn't really unit tests in the classical sense, maybe we should rename them. What we call unit here is more like a feature test, and integration tests are really http end to end tests.

dpschen:
I see! That makes sense.
For improving the runtime I still have a few other ideas that I want to try.
Will close this draft.

PR #871, opened 2025-06-04
Fix TypeScript issues in user and team views

kolaente:
## Summary
- fix Register view error handling
- type credentials in Login view
- type Teams listing
- refine team member management and imports
- handle PasswordReset casting

## Testing
- `pnpm run typecheck` *(fails: numerous remaining type errors)*

------
https://chatgpt.com/codex/tasks/task_e_684064a140208322be9922555b661bc9

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -39,7 +39,7 @@ import ContentLinkShare from '@/components/home/ContentLinkShare.vue'
 import NoAuthWrapper from '@/components/misc/NoAuthWrapper.vue'
 import Ready from '@/components/misc/Ready.vue'
 
-import {setLanguage} from '@/i18n'
+import {setLanguage, getBrowserLanguage} from '@/i18n'
 
 import {useAuthStore} from '@/stores/auth'
 import {useBaseStore} from '@/stores/base'
@@ -88,7 +88,7 @@ watch(userEmailConfirm, (userEmailConfirm) => {
 	router.push({name: 'user.login'})
 }, { immediate: true })
 
-setLanguage(authStore.settings.language)
+setLanguage(authStore.settings.language ?? getBrowserLanguage())
 useColorScheme()
 </script>
 
--- a/frontend/src/components/base/Expandable.vue
+++ b/frontend/src/components/base/Expandable.vue
@@ -83,47 +83,51 @@ function forceLayout(el: HTMLElement) {
 # see: https://vuejs.org/guide/built-ins/transition.html#javascript-hooks
 ###################################################################### */
 
-function beforeEnter(el: HTMLElement) {
-  el.style.height = '0'
-  el.style.willChange = 'height'
-  el.style.backfaceVisibility = 'hidden'
-  forceLayout(el)
+function beforeEnter(el: Element) {
+  const element = el as HTMLElement
+  element.style.height = '0'
+  element.style.willChange = 'height'
+  element.style.backfaceVisibility = 'hidden'
+  forceLayout(element)
 }
 
 // the done callback is optional when
 // used in combination with CSS
-function enter(el: HTMLElement) {
-  const height = getHeight(el) // Get the natural height
-  el.style.height = height // Update the height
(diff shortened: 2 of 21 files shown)


kolaente:
This was 100% done by Codex, not sure how much we can keep.

frontend/src/App.vue
```diff
@@ -88,7 +88,7 @@ watch(userEmailConfirm, (userEmailConfirm) => {
 	router.push({name: 'user.login'})
 }, { immediate: true })
 
-setLanguage(authStore.settings.language)
+setLanguage(authStore.settings.language ?? getBrowserLanguage())
```
dpschen, line 91: Maybe should have an initLanguage method or something like that instead. The `setLanguage` here made sense for me, since it waited for the authStore to init (reading from localstorage, not from server yet). Afaik we call setLanguage later with the correct user language, when logged in. Since the getBrowserLanguage is coming from i18n, we could also set getBrowserLanguage as default for the param?

frontend/src/components/base/Expandable.vue
```diff
@@ -83,47 +83,51 @@ function forceLayout(el: HTMLElement) {
 # see: https://vuejs.org/guide/built-ins/transition.html#javascript-hooks
 ###################################################################### */
 
-function beforeEnter(el: HTMLElement) {
-  el.style.height = '0'
-  el.style.willChange = 'height'
-  el.style.backfaceVisibility = 'hidden'
-  forceLayout(el)
+function beforeEnter(el: Element) {
+  const element = el as HTMLElement
```
dpschen, line 87: I think i would prefer here:
`@beforeEnter="(el) => beforeEnter(el as HTMLElement)"`.

frontend/src/views/teams/EditTeam.vue
```diff
@@ -331,22 +332,22 @@ async function save() {
 	}
 	showErrorTeamnameRequired.value = false
 
-	team.value = await teamService.value.update(team.value)
+       team.value = await teamService.value.update(team.value as ITeam)
```
dpschen, line 335: Indention

frontend/src/views/teams/EditTeam.vue
```diff
@@ -331,22 +332,22 @@ async function save() {
...
 	success({message: t('team.edit.success')})
 }
 
 async function deleteTeam() {
-	await teamService.value.delete(team.value)
+       await teamService.value.delete(team.value as ITeam)
```
dpschen, line 340: Probably better to check:

```
if (!team.value) {
  return // or throw error
}
```

frontend/src/views/teams/EditTeam.vue
```diff
@@ -393,16 +394,16 @@ async function findUser(query: string) {
 		return
 	}
 
-	const users = await userService.value.getAll({}, {s: query})
-	foundUsers.value = users.filter((u: IUser) => u.id !== userInfo.value.id)
+       const users = await userService.value.getAll(new UserModel(), {s: query})
```
dpschen, line 397: This fixes the type error, but seems wrong to me. The UserModel is not what getAll actually needs to send the request. We just didn't define a type for the request(s) yet.

dpschen:
> This was 100% done by Codex, not sure how much we can keep.

I just went through a bit. There are some interesting ideas, but in general we need to rework a lot. I think partly it could be better with the help of a better AGENTS.md

frontend/src/views/teams/EditTeam.vue
kolaente: Is that a problem of how we defined the types or how the logic is set up?

dpschen: In this case it's kind of both. Because getAll doesn't really need info for the model except the query the type that is expected by getAll originally failed (UserModel). But we don't need a UserModel class for the request. We have similar issues all over the other services. What we would need to do is separate output types and RequestTypes. Very often they are not the same.

PR #881, opened 2025-06-05
feat: improve edit project delete button design

dpschen:


kolaente:
Is the tertary action always delete? If not, maybe we shouldn't make it always red?

Issue #293, opened 2024-07-25
Checkboxes in task description require page refresh


cortl:
### Description

There are a few issues around checkboxes in descriptions, sometimes the boxes do not check/uncheck.

_Steps to Reproduce_
1. Add a task, create a description
```
[] test
```
2. Click save
3. Check `test`
4. Refresh page
5. `test` should be checked, but isn't

### Vikunja Version

v0.24.1

### Browser and version

Brave

### Can you reproduce the bug on the Vikunja demo site?

Yes

### Screenshots

You can see here where I checked a box, saved, then unchecked the box. The result should be that "0 of 2 tasks" and the text should not have strikethrough styling. Refreshing the page will also show this checkbox checked again.
![image](https://github.com/user-attachments/assets/4b4e9326-5ef5-4c95-8945-d5698d3f89a2)

kolaente:
https://community.vikunja.io/t/tasks-list-in-description-completely-broken/2490/2

PMP161316:
Dear kolaente,
Thank you very much for this helpful tool.
Unfortunately, I stumble across the bug described here almost every day and often lose information (checked checkbox) as a result.
Would it perhaps be conceivable - as a kind of work-around - to automatically reload/refresh the page after saving until this bug can be fixed?

kolaente:
I think that would be possible.

kolaente:
https://github.com/go-vikunja/vikunja/issues/563 has a little more info from me debugging this (but no fix, sadly)

kolaente:
https://github.com/go-vikunja/vikunja/issues/837 is another issue with the same root cause.

scosu:
Probably the same bug a bit differently:

1. Add task with checkboxes in the description
2. reload the page with the open task
3. Check the box
4. Uncheck the box (here the line still has a line through it though it shouldn't)
5. Reload the page with the open task
6. The box you unchecked is still checked

jasiekk913:
Hi, has anything changed with this issue? 

Also if you check/uncheck them in editing mode, everything works as expected (but number of completed tasks updates only when you save).

kolaente:
This should now be fixed as per the bot's comment, please check if it works for you.

PR #891, opened 2025-06-07
fix(frontend): mark only clicked task item

dpschen:


kolaente:
Does this resolve #293 ?

kolaente:
This seems oddly similar to https://github.com/go-vikunja/vikunja/pull/913 - does it fix https://github.com/go-vikunja/vikunja/issues/293 ?

kolaente (approved):
LGTM since it looks good and the test passes.

PR #900, opened 2025-06-08
fix: update sass-embedded and patch path

dpschen:


kolaente (changes requested):
Running the patch command now fails with

```
patchelf: getting info about '{}': No such file or directory

kolaente:
Still not working (same error)

PR #958, opened 2025-06-16
feat(frontend): highlight overdue tasks consistently

dpschen:


dpschen:
> why not use the class?

Good catch! I opted for a `data-` attribute so you can immediately see in the DOM in which state a component is; this separates state from styling, yet still allows styling based on that state. It also uses Vue’s boolean-attribute binding:

```vue
:data-is-overdue="isOverdue || undefined"
```


Didn’t test yet though. Because of that it’s draft

kolaente:
works, let's merge this

PR #960, opened 2025-06-16
fix: respect root path when reading configs

dpschen:


pkg/config/config.go
```diff
@@ -527,6 +528,29 @@ func setConfigFromEnv() error {
...
+	tmp.SetConfigName("config")
+	tmp.AddConfigPath(p)
+
+	if err := tmp.ReadInConfig(); err != nil {
+		var notFound viper.ConfigFileNotFoundError
+		if errors.As(err, &notFound) {
+			return
+		}
+		log.Warning(err.Error())
+		log.Warning("Using default config.")
```
kolaente, line 550: This is only true if no other config file could be read either.

pkg/config/config.go
```diff
@@ -527,6 +528,29 @@ func setConfigFromEnv() error {
 	return viper.MergeConfigMap(configMap)
 }
 
+func mergeConfigFromPath(p string) {
```
kolaente, line 535: Doesn't viper do this already?

kolaente:
closing as stale and conflicted, please ping if we should redo it

PR #961, opened 2025-06-16
fix: avoid xorm panic with tasks update

dpschen:


pkg/db/test.go
```diff
@@ -127,3 +127,63 @@ func AssertCount(t *testing.T, table string, where builder.Cond, count int64) {
 	require.NoErrorf(t, err, "Failed to assert count in db, error was: %s", err)
 	assert.Equalf(t, count, dbCount, "Found %d entries instead of expected %d in table %s", dbCount, count, table)
 }
+
+// BeginTx starts a transaction for a test and rolls it back when the test ends.
+func BeginTx(t *testing.T) *xorm.Session {
+	t.Helper()
+	s := NewSession()
+	require.NoError(t, s.Begin())
+	t.Cleanup(func() {
```
kolaente, line 136: TIL `t.Cleanup` is a thing!

pkg/db/test.go
```diff
@@ -127,3 +127,63 @@ func AssertCount(t *testing.T, table string, where builder.Cond, count int64) {
...
+		_ = s.Rollback()
```
kolaente, line 137: Rollback isn't required since setting up the test fixtures will throw away the database anyway

pkg/models/tasks_test.go
```diff
@@ -55,27 +55,22 @@ func TestTask_Create(t *testing.T) {
...
 			"bucket_id": 1,
 		}, false)
 
 		events.AssertDispatched(t, &TaskCreatedEvent{})
 	})
 	t.Run("with reminders", func(t *testing.T) {
-		db.LoadAndAssertFixtures(t)
-		s := db.NewSession()
-		defer s.Close()
+		s := db.BeginTx(t)
```
kolaente, line 73: Shouldn't we then use this everywhere?

pkg/db/test.go
```diff
@@ -127,3 +127,63 @@ func AssertCount(t *testing.T, table string, where builder.Cond, count int64) {
...
+		s.Close()
+	})
+	return s
+}
+
+// AssertExistsTx asserts that values exist within the given transaction.
+func AssertExistsTx(t *testing.T, s *xorm.Session, table string, values map[string]interface{}, custom bool) {
```
kolaente, line 144: Is there a way to have these functions with less duplication?

pkg/db/test.go
dpschen: If I got it correctly it's helpful for testing

pkg/models/tasks_test.go
dpschen: I might have more prs. Tried to limit the size to mage them easily mergeable. Maybe we should collect these kind of tasks, so that we don't forget them? Right now I (mis)use the history of codex for this.

kolaente: Maybe just create an issue for this? Or a task in Vikunja?

PR #903, opened 2025-06-10
Add timezone check on startup

kolaente:
## Summary
- validate configured time zone after loading configuration
- log a critical message if the zone cannot be parsed

## Testing
- `go test ./...` *(fails: could not read fixture files)*
- `go vet ./...` *(fails: struct field tag conflict)*
- `golangci-lint run`

------
https://chatgpt.com/codex/tasks/task_e_6848098398a08322ae8f9307d7f672ae

PR #962, opened 2025-06-16
fix: set default timezone when placeholder used

dpschen:
I guess this is just convenience. But foremost the DX should be fixed for this.


kolaente:
This should be validated instead, similar to https://github.com/go-vikunja/vikunja/pull/903

PR #1023, opened 2025-06-25
chore(deps): upgrade tailwind to v4

kolaente:


kolaente:
It seems like the problem is the prefix: I couldn't figure out how to set the prefix without importing the preflight at the same time. And having Bulma use the same layers as tailwind does not seem to make things simpler.

Given how little we're using tailwind, maybe it's easier to just remove it? Or remove bulma?

dpschen:
So preflight is necessary for prefix? That seems like a bug, no?

kolaente:
Looks like it, but it might be interfering with how we use Bulma.

frontend/src/styles/tailwind.css
```diff
@@ -0,0 +1,4 @@
+@layer theme, utilities;
+
+@import "tailwindcss/theme.css" layer(theme) prefix(tw);
+@import "tailwindcss/utilities.css" layer(utilities) prefix(tw);
```
dpschen, line 4: Why didn't you import base and components?

```css
/*
** see discussion under:
** [tailwindlabs/tailwindcss#17594 (comment)](https://github.com/tailwindlabs/tailwindcss/discussions/17594?utm_source=chatgpt.com#discussioncomment-12752302)
 */
@layer theme, base, components, utilities;
@import "tailwindcss/theme.css" layer(theme) prefix(tw);
@import "tailwindcss/utilities.css" layer(utilities);
```

kolaente: I tried, but it does not seem to make a difference

dpschen: What exactly is the issue?

kolaente: Looks broken when adding it, looks broken when not adding it.

PR #1001, opened 2025-06-24
feat: rebuild the gantt chart

kolaente:
## Things left to do

- [x] Use the `GanttBarPrimitive` and the styled variant - currently it is implemented directly into NewGanttChart
- [x] Rename `NewGanttChart` and remove the old implementation
- [x] Make the vertical lines in the background work
- [x] Adjust cypress tests
- [x] Clean up the structure of the new chart, remove everything that's not useful
- [x] Inline all prop type definitions for consistsency
- [x] Fix focus handling

Closes https://github.com/go-vikunja/vikunja/pull/1053
Resolves https://github.com/go-vikunja/vikunja/issues/855

## Summary by CodeRabbit

- New Features
  - Rebuilt interactive Gantt chart with drag-to-move, resize handles, and double‑click to open tasks
  - Keyboard navigation and focus management for rows/cells
  - Sticky timeline header with months and days (includes year) and vertical grid lines
  - Enhanced accessibility with ARIA labels and improved bar text contrast
  - Alternating row backgrounds for readability

- Translations
  - Added comprehensive Gantt-related strings (chart label, month/day labels, resize prompts, date types)

- Tests
  - Updated end-to-end tests to new DOM structure and pointer-based interactions

- Chores
  - Removed an unused Gantt dependency

PR #1053, opened 2025-07-01
fix(gantt): correct end date handling

dpschen:


kolaente:
Won't this be obsolete with #1001?

dpschen:
Yes. I‘m just going through all the stuff that I still had

kolaente:
I think since we have #1001, we should merge that first - what do you think?

PR #1054, opened 2025-07-01
fix: handle quotes in project filter titles

### LINKED ISSUE #743: special character in project name makes it impossible to filter on the project
### Description

Steps to reproduce:

* Name a project `Todo's`
* In a save filter add `project = Todo's`
* Get the following error when trying to save:

Error
Invalid task filter value. The task filter value 'Todo's' for field 'project_id' is invalid.

If I remove the `'`, everything works as expected.

### Vikunja Version

v0.24.1-1080-ea1a1843

### Browser and version

_No response_

Yes

### Screenshots

_No response_

dpschen:
## Summary
- allow quotes in filter regex
- adjust filter parsing logic
- test converting project titles with single quotes

Fixes #743

-------

The filter helper getFilterFieldRegexPattern() changed how it parses filter values.
Earlier versions matched optional quotes around the value and disallowed any quote characters inside it:

```ts
export const FILTER_OPERATORS_REGEX = '(<|>|<=|>=|=|!=|in)'

export function getFilterFieldRegexPattern(field: string): RegExp {
        return new RegExp('(' + field + '\\s*' + FILTER_OPERATORS_REGEX + '\\s*)([\\'\"]?)([^\\'\"&|()<]+\\1?)?', 'ig')
}
```

This meant values like project = Todo's failed to match because the single quote was interpreted as closing a quoted value.

The current code dropped the quote-specific groups and simply captures any characters until a filter operator or parenthesis:

```ts
export const FILTER_OPERATORS_REGEX = '(<|>|<=|>=|=|!=|not in|in)'

export function getFilterFieldRegexPattern(field: string): RegExp {
        return new RegExp('(' + field + '\\s*' + FILTER_OPERATORS_REGEX + '\\s*)([^&|()]+)', 'ig')
}
```

This mirrors the server-side regex used to parse filter expressions:

```go
// Regex pattern to match filter expressions
re := regexp.MustCompile(`(\w+)\s*(>=|<=|!=|~|\?=|\?!=|=|>|<)\s*([^&|()]+)`)
```

### Why it was like it was before:
The old pattern expected a value to be optionally wrapped in quotes and prohibited any internal quote characters. That avoided accidental splitting if quotes were used as delimiters, but it broke when project or tag names themselves contained a ' or ".

### Why it changed:
To support names that include quotes, the frontend regex now matches everything up to &, |, or parentheses—just like the backend. This allows titles such as "Todo's" to be parsed correctly, ensuring filters behave consistently across frontend and backend.

### DIFF
--- a/frontend/src/helpers/filters.test.ts
+++ b/frontend/src/helpers/filters.test.ts
@@ -187,6 +187,16 @@ describe('Filter Transformation', () => {
 
 			expect(transformed).toBe('project = 1')
 		})
+
+		it('should resolve projects with single quotes in the title', () => {
+			const transformed = transformFilterStringForApi(
+				"project = Todo's",
+				nullTitleToIdResolver,
+				(title: string) => title === "Todo's" ? 1 : null,
+			)
+
+			expect(transformed).toBe('project = 1')
+		})
 	})
 
 	describe('To API', () => {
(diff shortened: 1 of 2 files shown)


kolaente:
This seems like it will be made obsolete by https://github.com/go-vikunja/vikunja/pull/1046?

PR #1056, opened 2025-07-01
Feat/fix spelling mistakes and run lint/tests

dpschen:

PR #1057, opened 2025-07-01
Feat/update comments and fix misspellings

dpschen:


kolaente:
How's this different than https://github.com/go-vikunja/vikunja/pull/1056?

Issue #677, opened 2025-04-17
Editor URL Link should not extend beyond current word


nebula-it:
### Description

In the editor(both main body and comments) of task, if you add a hyperlink to a word by selecting the word and then pasting using `Ctrl+V` and then hitting the `space` and typing next word, the next words are hyperlinked as well. Which is not ideal the hyperlink should be dropped from formatting once you hit `space`.

### Vikunja Version

v0.24.6

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
This might be an upstream issue. Are you able to reproduce this here? https://templates.tiptap.dev/CGLrDQrzJd

nebula-it:
Nope, cannot reproduce there. It works as expected i.e after the space it does not add next word to hyperlink

kolaente:
Fixed in https://github.com/go-vikunja/vikunja/pull/912, please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

PR #1059, opened 2025-07-01
fix(editor): prevent links from extending after space

dpschen:
fixes: #677


dpschen:
I just saw https://github.com/go-vikunja/vikunja/pull/912/files . 
@kolaente Do you think this still makes sense?

kolaente:
Seems like you solution is simpler. Do you think you could remove the other one?

dpschen:
Done

kolaente (approved):
Thanks!

PR #1069, opened 2025-07-02
Update default database path

dpschen:


pkg/config/config.go
```diff
@@ -574,6 +574,12 @@ func InitConfig() {
 		log.Warningf("Failed to set config from environment variables: %s", err.Error())
 	}
 
+	// Update the database path to match a changed rootpath when it was not explicitly set
+	defaultDBPath := filepath.Join(getBinaryDirLocation(), "vikunja.db")
+	if DatabasePath.GetString() == defaultDBPath {
```
kolaente: This seems redundant since the default root path is the directory of the binary?

dpschen: If I understood it correctly, that's incorrect.

Because `database.path` is calculated before `service.rootpath` (from the user) is read.
=> gets “stuck” on the binary-directory path.

I replaced to code to make it clearer.

kolaente: But shouldn't the fix then be to initialize the rootpath before the database path?

PR #1080, opened 2025-07-03
mail: disable queue when mailer disabled

dpschen:


pkg/notifications/notify_disabled_test.go
```diff
@@ -0,0 +1,43 @@
+package notifications_test
```
kolaente, line 1: This uses the wrong package name

kolaente (changes requested):

kolaente:
Replaced by https://github.com/go-vikunja/vikunja/pull/2069

PR #1083, opened 2025-07-03
test(user): verify invalid deletion token

dpschen:
- return a clear error when confirming deletion with an invalid token
- add english translation for deletion token error
- test invalid token on account deletion confirmation


kolaente:
Already fixed by 4faf50a91

PR #1116, opened 2025-07-14
Fix CalDAV REPORT request to properly respond with VTODO objects

v-yarotsky:
First off, thanks for this awesome project!

Currently, when the following request is made:
```xml
<?xml version="1.0" encoding="utf-8"?>
<C:calendar-query xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav">
  <D:prop>
    <C:calendar-data/>
  </D:prop>
  <C:filter>
    <C:comp-filter name="VCALENDAR">
      <C:comp-filter name="VTODO">
      </C:comp-filter>
    </C:comp-filter>
  </C:filter>
</C:calendar-query>
```

Vikunja responds with
```xml
<?xml version="1.0" encoding="UTF-8"?>
<D:multistatus xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav" xmlns:CS="http://calendarserver.org/ns/">
  <D:response>
    <D:href>/dav/projects/1/6cbf88d0-1129-438e-a1a3-5b9c1f92bd96.ics</D:href>
    <D:propstat>
      <D:prop>
        <C:calendar-data>BEGIN:VCALENDAR
VERSION:2.0
METHOD:PUBLISH
X-PUBLISHED-TTL:PT4H
X-WR-CALNAME:
PRODID:-//Vikunja Todo App//EN
END:VCALENDAR</C:calendar-data>
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
  </D:response>
  <D:response>
    ...
  </D:response>
</D:multistatus>  
```

As you can see, it does not include the `VTODO` element.
I've stumbled upon this when integrating Vikunja with Home Assistant via their [CalDAV integration](https://www.home-assistant.io/integrations/caldav/).
When marking an item as completed via Home Assistant, I saw an error:
```
Failed to perform the action todo/update_item. 'CalendarObjectResource' object has no attribute 'set_due'
```
This has been filed as a bug with Home Assistant previously: https://github.com/home-assistant/core/issues/122346.

As it turned out, they [fetch the TODO item](
https://github.com/home-assistant/core/blob/bc07030304581d7529f0fe49a7f5ac33d2e864d5/homeassistant/components/caldav/todo.py#L151) before updating it. Because the response doesn't include the actual `VTODO`, the library used by Home Assistant fails to unmarshal the response properly, resulting in the aforementioned error.

With this change, Vikunja will actually include the `VTODO` in the response, fixing the issue.
This should also be consistent with the spec. I've followed the example from [RFC 4791 Section 7.8.9](https://datatracker.ietf.org/doc/html/rfc4791#section-7.8.9).

Updated response example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<D:multistatus xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav" xmlns:CS="http://calendarserver.org/ns/">
  <D:response>
    <D:href>/dav/projects/1/6cbf88d0-1129-438e-a1a3-5b9c1f92bd96.ics</D:href>
    <D:propstat>
      <D:prop>
        <C:calendar-data>BEGIN:VCALENDAR
VERSION:2.0
METHOD:PUBLISH
X-PUBLISHED-TTL:PT4H
X-WR-CALNAME:
PRODID:-//Vikunja Todo App//EN
BEGIN:VTODO
UID:6cbf88d0-1129-438e-a1a3-5b9c1f92bd96
DTSTAMP:20250708T063859Z
SUMMARY:Put out bins – Garbage/Compost
COMPLETED:20250708T063859Z
STATUS:COMPLETED
DUE:20250722T040000Z
CREATED:20250707T021855Z
PRIORITY:3
RRULE:FREQ=SECONDLY;INTERVAL=1209600
LAST-MODIFIED:20250708T063859Z
END:VTODO
END:VCALENDAR</C:calendar-data>
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
  </D:response>
  <D:response>
    ...
  </D:response>
</D:multistatus>  
```

### DIFF
--- a/pkg/webtests/caldav_test.go
+++ b/pkg/webtests/caldav_test.go
@@ -17,11 +17,14 @@
 package webtests
 
 import (
+	"encoding/xml"
 	"net/http"
+	"strings"
 	"testing"
 
 	"code.vikunja.io/api/pkg/routes/caldav"
 
+	ics "github.com/arran4/golang-ical"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


pkg/webtests/caldav_test.go
```diff
@@ -318,3 +321,76 @@ END:VCALENDAR`
 		assert.Contains(t, rec.Body.String(), "RELATED-TO;RELTYPE=PARENT:uid-caldav-test-parent-task-another-list")
 	})
 }
+
+func TestCaldavReport(t *testing.T) {
```
kolaente: Nit: Doesn't this test the _projects_ report response? Then the test method should be called accordingly.

kolaente (changes requested):
Looks pretty good! Only one nit, and please fix the lint issue.

PR #1119, opened 2025-07-14
fix: don't panic when using api token when not correctly put into context

kolaente:
## Summary
- prevent panic by loading the token owner into request context
- use the context user object in `GetCurrentUser`

## Testing
- `mage lint:fix`

------
https://chatgpt.com/codex/tasks/task_e_687577d0cec88322a07b3eabd96decf9

### DIFF
--- a/pkg/user/user.go
+++ b/pkg/user/user.go
@@ -419,10 +419,6 @@ func CheckUserPassword(user *User, password string) error {
 
 // GetCurrentUserFromDB gets a user from jwt claims and returns the full user from the db.
 func GetCurrentUserFromDB(s *xorm.Session, c echo.Context) (user *User, err error) {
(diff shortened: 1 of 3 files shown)


kolaente:
Should fix https://vikunja.sentry.io/share/issue/21d4215e5a464c4ab11b7ada11d00369/

pkg/user/user.go
```diff
@@ -419,6 +419,10 @@ func CheckUserPassword(user *User, password string) error {
 
 // GetCurrentUserFromDB gets a user from jwt claims and returns the full user from the db.
 func GetCurrentUserFromDB(s *xorm.Session, c echo.Context) (user *User, err error) {
+	if apiUser, ok := c.Get("api_user").(*User); ok {
```
kolaente: This seems redundant since `GetCurrentUser` will do the same

Issue #1086, opened 2025-07-03
Editing labels that do not have a color often fails


bernardosulzbach:
### Description

I apologize for the "often" in "often fails," but I swear it worked once or twice at some point.

1. Clear the color of a label and save it. \
I think this already fails sometimes, but I can at least bash it into working.
1. Change the name of the label and try to save this change.

Outcome: Struct is invalid. Invalid Data

Request:
```json
{
    "max_right": null,
    "id": 345,
    "title": "Colorless 2",
    "hex_color": "var(--grey-200)",
    "description": "",
    "created_by": {
        "max_right": null,
        "id": 1,
        "email": "",
        "username": "demo",
        "name": "",
        "exp": 0,
        "type": 0,
        "created": null,
        "updated": null,
        "settings": {
            "max_right": null,
            "name": "",
            "email_reminders_enabled": true,
            "discoverable_by_name": false,
            "discoverable_by_email": false,
            "overdue_tasks_reminders_enabled": true,
            "week_start": 0,
            "timezone": "",
            "language": "en",
            "frontend_settings": {
                "play_sound_when_done": true,
                "quick_add_magic_mode": "vikunja",
                "color_schema": "auto",
                "allow_icon_changes": true,
                "default_view": "first",
                "minimum_priority": 2
            }
        }
    },
    "project_id": 0,
    "text_color": "hsl(220, 13%, 91%)",
    "created": "2025-07-03T14:32:33.000Z",
    "updated": "2025-07-03T19:04:14.000Z"
}
```
Response:
```json
{
    "message": "Struct is invalid. Invalid Data"
}
```

Happens on my instance using v0.24.6 and on try using
> Frontend version: dev
> API version: v0.24.1-1529-efff6955

The `"hex_color": "var(--grey-200)"` is highly suspicious, but I didn't look at the backend code.

### Vikunja Version

v0.24.6

### Browser and version

Version 137.0.7151.120 (Official Build) (64-bit)

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
Fixed in https://github.com/go-vikunja/vikunja/pull/1124, please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

bernardosulzbach:
Yes, it seems fixed on try. Thank you.

PR #1124, opened 2025-07-15
fix(labels): correctly fall back to variable colors when no label color is set

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1086

------
https://chatgpt.com/codex/tasks/task_e_6867751462fc8322bd00be64aa159fd9

### DIFF
--- a/frontend/src/models/label.ts
+++ b/frontend/src/models/label.ts
@@ -28,15 +28,10 @@ export default class LabelModel extends AbstractModel<ILabel> implements ILabel 
 			this.hexColor = '#' + this.hexColor
 		}
 
-		if (this.hexColor === '') {
-			this.hexColor = 'var(--grey-200)'
-			this.textColor = 'var(--grey-800)'
-		} else {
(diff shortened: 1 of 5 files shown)


frontend/src/models/label.ts
```diff
@@ -28,15 +28,10 @@ export default class LabelModel extends AbstractModel<ILabel> implements ILabel
...
-		if (this.hexColor === '') {
-			this.hexColor = 'var(--grey-200)'
-			this.textColor = 'var(--grey-800)'
-		} else {
-			this.textColor = colorIsDark(this.hexColor)
-				// Fixed colors to avoid flipping in dark mode
-				? 'hsl(215, 27.9%, 16.9%)' // grey-800
-				: 'hsl(220, 13%, 91%)' // grey-200
-		}
+		this.textColor = colorIsDark(this.hexColor)
```
kolaente: This should only happen when no `hexColor` is set, and a fallback for the text color using var as well (because the text and hex color could be different in dark / light mode)

PR #1149, opened 2025-07-19
feat: replace axios with direct fetch usage

kolaente:


kolaente:
Might revisit later, but not for now

PR #1152, opened 2025-07-21
feat: rtl layout support for Arabic and Hebrew languages

kolaente:
This PR adds support for rtl layout in Arabic and Hebrew languages. The goal here is to get something working good enough quickly because this change touches a lot of files, causing many merge conflicts with other changes.

Things left to do:

- [x] mirror the text icon in task (for the description)
- [x] Fix rtl spacing in some cases (see comment below)
- [x] Are there equivalents for `left`/`right` css properties we could use?
- [x] keyboard shortcuts modal looks not quite right (and maybe others too?)
- [x] Fix the user menu
- [x] Form spacing
- [x] `has-icon-left` et al
- [x] icon buttons (at least on the task detail page)
- [x] task input: icon and spacing

### DIFF
--- a/frontend/src/components/tasks/partials/Comments.vue
+++ b/frontend/src/components/tasks/partials/Comments.vue
@@ -16,7 +16,7 @@
 		<div class="comments">
 			<span
 				v-if="taskCommentService.loading && saving === null && !creating"
-				class="is-flex is-align-items-center my-4 mis-2"
+				class="is-flex is-align-items-center mbs-4 mbe-4 mis-2"
 			>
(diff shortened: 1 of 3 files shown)


kolaente:
needs a `direction` clause somewhere

kolaente:
It seems like the "this is now rtl"-mode setting from the `dir` attribute does not work correctly - or I misunderstood something.

In rtl, it looks like this:

image

But in ltr, it looks like this:

image

kolaente:
well that last change was easier to fix than though - 599e49b11147e3c7482524c02bd695f24b932cc7

PR #1158, opened 2025-07-22
fix "null" in project views

CrazyWolf13:
Table project_views contains NULL which was decoded to weird UTF-8 characters and compledtely broke the app. 

grafik

This command fixed my DB:
```bash
sqlite3 vikunja.db "UPDATE project_views SET bucket_configuration = NULL;"
```

Adding this null check ensures "null" is treated as actual null.

Please again review thorougly, I'm not a go dev and that `nil` is only based on my research.

### DIFF
--- a/pkg/modules/dump/restore.go
+++ b/pkg/modules/dump/restore.go
@@ -228,7 +228,7 @@ func convertFieldValue(fieldName string, value interface{}, isFloat bool) (inter
 		if v == "null" {
 			return nil, nil
 		}
-		
+				
 		decoded, err := base64.StdEncoding.DecodeString(v)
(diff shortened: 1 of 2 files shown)


kolaente:
Unsure why the test is now failing, can you fix the lint issue and check again?

CrazyWolf13:
@kolaente was unsure too, but seems like they pass now.

Can you confirm, this `nil` should work? I don't really have an idea how to properly test this.

kolaente:
Added a handful of tests around this and made it case-insensitive. Should be good now.

CrazyWolf13:
awesome :)

PR #1300, opened 2025-08-16
fix: strip label syntax with parentheses from task title

### LINKED ISSUE #1299: Syntax for tags is not always stripped from the task title even though it is parsed
### Description

| Input      | Title      | Tag   |
|------------|------------|-------|
| a *"a a"   | a          | a a   |
| a *"a (a)" | a *"a (a)" | a (a) |

Image

The title should be "a" in both cases. It matters because some tag titles are repeated and need disambiguation, e.g., [Death (metal band)](https://en.wikipedia.org/wiki/Death_(metal_band)) vs [Death](https://en.wikipedia.org/wiki/Death).

### Vikunja Version

v0.24.6, v0.24.1-1816-c4889127

### Browser and version

Google Chrome Version 139.0.7258.127 (Official Build) (64-bit)

Yes

### Screenshots

_No response_

kolaente:
Resolves #1299

### DIFF
--- a/frontend/src/modules/parseTaskText.test.ts
+++ b/frontend/src/modules/parseTaskText.test.ts
@@ -643,7 +643,7 @@ describe('Parse Task Text', () => {
 			expect(result.labels).toHaveLength(1)
 			expect(result.labels[0]).toBe('label with space')
 		})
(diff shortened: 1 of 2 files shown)


kolaente:
Test failure is unrelated, merging.

Issue #1319, opened 2025-08-20
There is missing translation key for values : project.share.permission.*


zapp88:
### Description

The key project.share.permission.* is missing from all translations and as such cannot be translated via translation service .

### Vikunja Version

871861a

### Can you reproduce the bug on the Vikunja demo site?

Yes

### Screenshots

<img width="482" height="135" alt="Image" src="https://github.com/user-attachments/assets/0073bf30-f718-49bc-a89d-1ed3adafd1a6" />

PR #1320, opened 2025-08-20
fix(frontend): add missing permission translation keys

zapp88:
Added missing 'permission' section to both English and Polish translation files for project sharing permissions. This resolves the issue where permission labels were showing as raw keys (project.share.permission.readWrite) instead of translated text.

The permission section duplicates the existing 'right' section to maintain compatibility with components that reference either naming convention.

This PR fixes : https://github.com/go-vikunja/vikunja/issues/1319

The issue is resolved after fix :
obraz

### DIFF
--- a/frontend/src/i18n/lang/pl-PL.json
+++ b/frontend/src/i18n/lang/pl-PL.json
@@ -289,18 +289,6 @@
 				"addedSuccess": "{type} został pomyślnie dodany.",
 				"updatedSuccess": "{type} został pomyślnie zaktualizowany."
 			},
-			"right": {
-				"title": "Uprawnienia",
-				"read": "Tylko do odczytu",
-				"readWrite": "Odczyt i zapis",
-				"admin": "Administracja"
-			},
-			"permission": {
-				"title": "Uprawnienia",
-				"read": "Tylko do odczytu",
-				"readWrite": "Odczyt i zapis",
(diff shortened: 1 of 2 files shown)


kolaente:
The rights section isn't used anymore, you can just rename it to permission. 

Please don't add the polish translation in a PR, it will be overriden from crowdin.

zapp88:
> The rights section isn't used anymore, you can just rename it to permission.
> 
> Please don't add the polish translation in a PR, it will be overriden from crowdin.

Applied suggestions.

frontend/src/i18n/lang/pl-PL.json
```diff
@@ -289,7 +289,7 @@
 				"addedSuccess": "{type} został pomyślnie dodany.",
 				"updatedSuccess": "{type} został pomyślnie zaktualizowany."
 			},
-			"right": {
+			"permission": {
```
kolaente: Please don't change the polish translation in this PR, it will be overriden by Crowdin.

kolaente (changes requested):

frontend/src/i18n/lang/pl-PL.json
zapp88: Wasn't sure if Crowdin will pick it up since that wrong key name. Applied.

PR #1337, opened 2025-08-24
fix: mobile device  minimum width

NeoHuncho:
## Description of bug
The mobile breakpoint is set to an arbitrary value by dividing the tablet size by two (384.5px).However, many mainstream devices, such as the iPhone SE, have a smaller width. I suggest a minimum width of 320, which will fit 99% of mobile devices.  

## Before fix ( You have to scroll down a whole page to get to your tasks)
Screenshot 2025-08-24 at 12 11 17

## After fix
Screenshot 2025-08-24 at 12 02 29

### DIFF
--- a/frontend/src/styles/common-imports.scss
+++ b/frontend/src/styles/common-imports.scss
@@ -15,8 +15,8 @@ $family-sans-serif: 'Open Sans', Helvetica, Arial, sans-serif;
 // the default values get overwritten by the definitions above
 @import "bulma-css-variables/sass/utilities/_all";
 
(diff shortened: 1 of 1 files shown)


kolaente:
Sounds good, but the "Add" button should become the plus in any case (the text is visible in your "after" screenshot).

NeoHuncho:
This was due to an issue of my hot reload , i can confirm that the button remains the + button variant (the width has to go over the tablet width)


https://github.com/user-attachments/assets/e14b4980-d707-4ef4-9334-d444e69a7ca8

kolaente:
Thanks!

PR #1338, opened 2025-08-24
fix: stop event propagation for clicks inside the DeferTask popup

NeoHuncho:
## Description of bug
When using the DeferTask pop-up component, you are often redirected to the task because clicks propagate outside the components (see video below)

## Before fix
https://github.com/user-attachments/assets/47e2e3db-2ccc-4b04-8a72-e33bd0fa0432

## After fix
https://github.com/user-attachments/assets/7cb2e0ab-1d3b-4af4-bb7e-384de3783be0


kolaente (approved):
Thanks!

PR #1423, opened 2025-09-05
feat(auth): add AuthIssuer configuration for OpenID provider

Stijn-Timmer:
This pull request introduces a new configuration option that allows users to explicitly set the **OpenID Connect (OIDC) Issuer**.

---

### **Changes Introduced:**

- **Added `authissuer` configuration option**: A new, optional field has been added to the configuration file.
- **Default behavior**: If the `authissuer` is not explicitly set, the system will continue to use the OpenID discovery endpoint to automatically determine the issuer, maintaining backward compatibility.

### DIFF
--- a/pkg/modules/auth/openid/openid.go
+++ b/pkg/modules/auth/openid/openid.go
@@ -63,6 +63,7 @@ type Provider struct {
 	UsernameFallback    bool   `json:"username_fallback"`
 	ForceUserInfo       bool   `json:"force_user_info"`
 	RequireAvailability bool   `json:"-"`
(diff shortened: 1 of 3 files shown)


kolaente:
What is the use case for this?

IPv4Addr:
In fusionauth our tenant issuer differs from our auth server hostname causing a issuer mismatch, other tools also provide the option to set the issuer (rancher, grafana) so it seems to be no problem to add this small property.

IPv4Addr:
Turns out we are not using this configuration correctly in our authentication provider, sorry for wasting your time.

PR #1457, opened 2025-09-10
fix: reload list view when marking recurring task done

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/556


kolaente:
@codex fix comments

PR #1462, opened 2025-09-11
fix: guard saved filter requests

kolaente:
Fixes regression from https://github.com/go-vikunja/vikunja/pull/1457


kolaente:
@codex review

Issue #1467, opened 2025-09-11
After navigating to a task directly in kanban view, project is not loaded properly


hdev1:
### Description

Whenever I navigate to a task directly in the kanban view (i.e. https://try.vikunja.io/tasks/1011), it seems the project does not get loaded properly anymore. I can see [object Promise] behind the task in the kanban view, insetad of the usual component.

Clicking away seems to set the project ID to zero (i.e. https://try.vikunja.io/projects/0/756). This is reproducible in the demo site.

### Vikunja Version

v1.0.0-rc1

### Can you reproduce the bug on the Vikunja demo site?

Yes

### Screenshots

<img width="1202" height="359" alt="Image" src="https://github.com/user-attachments/assets/152bfd38-cd3c-4941-8bd6-787dfc5d16ac" />

kolaente:
Are you entering the url in a new tab into the browser or reloading? Because the kanban board in the background will only be loaded if you opened that task from the kanban view.

I was unable to verify your issue on the demo, bug I might have missed steps to reproduce it. I've tested directly entering the url in a new tab and reloadin.

hdev1:
I was reloading, instead of entering a new tab. I just tried to record a video to reproduce it, and it actually seems to happen inconsistently in the demo. You can see it takes me a few reloads to get the error to pop up. Weirdly enough, in my production instance, it happens every time without fail.

https://github.com/user-attachments/assets/095f6db7-e1d3-4199-9725-e4a5c3c36381

kolaente:
This sounds like a race condition.

kolaente:
Should be fixed in https://github.com/go-vikunja/vikunja/pull/1472, please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

hdev1:
Will give it a try. Thank you @kolaente

hdev1:
Sorry for the late response - was a bit busy the past few days. I've updated my instances to v1.0.0-rc2 and I can confirm this bug is now resolved!

PR #1472, opened 2025-09-11
fix(kanban): guard task modal race conditions

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1467


kolaente:
@codex review

Issue #1410, opened 2025-09-04
several UI Issues on Mobile/PWA and even Desktop


NeptunedGmbH:
### Description

First time I noticed was when I used the Vikunja Roadmap. It's impossible to navigate back to the Board after you opened a Task. There's no UI and the back button in the browser doesn't work as well.

I since had this issue several times once on the desktop, dont know where tho...

But in the PWA the lack of a back button gets you stuck a lot... When opening a task from the "Current Task" View for example. The only way back is the Burger menu.

Moving Buckets is basically impossible on mobile (iOS) since you dont have any control and even moving tasks to a list that is not adjacent is really hard to do. 

### Vikunja Version

Version: v0.24.6

### Browser and version

Arc 1.110.1 and Safari on iOS 26

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
Is this only a problem on mobile?

NeptunedGmbH:
> Is this only a problem on mobile?

I also have the issue with the roadmap on desktop. A few days ago I found a similiar issue in my instance, didin't document it sadly.

The main issue is mobile though

kolaente:
With https://github.com/go-vikunja/vikunja/pull/1475, there is now a back button which is shown on mobile and when opening the task full window (for example when coming from the list or table views). Please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

PR #1475, opened 2025-09-11
fix(task): provide back button when opening task detail

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1410


kolaente:
@codex review

Issue #1478, opened 2025-09-11
Inconsistent filter persistence across views (Kanban, Table, List)


Plong-Wasin:
### Description

There is inconsistent behavior in how filters and query parameters are handled across different views (Kanban, Table, and List).  

- **Kanban view:** Filters do not generate query parameters in the URL.  
- **Table view:** Filters generate query parameters in the URL, but after refreshing the page, the applied filters are lost.  
- **List view:** Filters generate query parameters in the URL, and after refreshing the page, the filters remain correctly applied.  

**Expected behavior:**  
Filter handling should be consistent across all views. If filters are reflected in the query parameters, refreshing the page should not remove them.  

**Actual behavior:**  
- Kanban: No query parameters at all when filters are applied.  
- Table: Query parameters appear, but filters disappear after refresh.  
- List: Works as expected — query parameters appear and filters persist after refresh.  

**Steps to reproduce:**  
1. Open a project in Kanban view and apply filters — observe no query parameters in the URL.  
2. Switch to Table view, apply filters, and refresh the page — filters disappear.  
3. Switch to List view, apply filters, and refresh the page — filters persist as expected.  

### Vikunja Version

v1.0.0-rc1-141-d14443d2

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
I would have guessed this is a regression introduced by https://github.com/go-vikunja/vikunja/commit/0dee1789a274a21c1e1cadfe7bd4e5d4a901c9ca but it seems to have been broken before as well.

kolaente:
Fixed in https://github.com/go-vikunja/vikunja/pull/1482, please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

Plong-Wasin:
@kolaente It still does not work in the table view.  
Here is a reproducible example on [try](https://try.vikunja.io/projects/1/247?filter=done+not+in+true,false&page=1) (v1.0.0-rc1-132-d147a01c).  

Should I open a new issue for this, or continue the discussion here?

kolaente:
We can continue here.

Does it work in the other views?

Plong-Wasin:
It works correctly in the other views.

PR #1482, opened 2025-09-11
fix(filters): persist url filter query across views

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1478


kolaente:
@codex review

PR #1509, opened 2025-09-15
Add S3 file storage support

### LINKED ISSUE #1469: Add support for S3 object storage
### Description

Hi there.

I'd like to suggest adding support for S3-compatible object storage services as an alternative to the local filesystem.

It could eventually be implemented as a plugin, but I haven't studied the plugin API yet so I'm not sure it's even feasible.

ghost:
Fixes #1469


frontend/package-lock.json
kolaente, line 1: Please don't commit this. This project uses pnpm.

frontend/src/indexes/index.ts
kolaente, line 1: Why the changes here?

kolaente (changes requested):
Please add tests for this.

kolaente:
On second thought, implementing S3 Support by implementing it as an afero driver is one way to do this, but then it should happen in a separate library instead. Vikunja does not need all features of Afero, and since they don't seem to implement S3 support any time soon, we'll have to implement the parts that are needed in Vikunja instead. But we don't want to maintain a bunch of dead code.

There are a few discussed here: https://github.com/spf13/afero/issues/11 (all of them haven't received updates for years)

kolaente:
Here is an up to date implementation: https://github.com/fclairamb/afero-s3

Please implement that or use the AWS sdk to support only the operations actually used in Vikunja.

frontend/src/indexes/index.ts
ghost: Fixed

frontend/package-lock.json
ghost: Fixed

pkg/files/s3_integration_test.go
```diff
@@ -0,0 +1,90 @@
...
+	config.FilesS3Region.Set("us-east-1")
+	config.FilesS3AccessKey.Set("test-access-key")
+	config.FilesS3SecretKey.Set("test-secret-key")
+
+	// This should not panic with valid configuration
+	assert.NotPanics(t, func() {
+		InitFileHandler()
+	})
+
+	// Test with missing S3 configuration
```
kolaente, line 59: Please make this two tests with t.Run

pkg/files/s3_integration_test.go
```diff
@@ -0,0 +1,90 @@
...
+	config.FilesS3Endpoint.Set("")
+	config.FilesS3Bucket.Set("")
+	config.FilesS3AccessKey.Set("")
+	config.FilesS3SecretKey.Set("")
+
+	// This should panic with missing configuration
+	assert.Panics(t, func() {
```
kolaente, line 66: Panicking is not a valid error handling method. Return an error and assert that.

pkg/files/s3_integration_test.go
```diff
@@ -0,0 +1,90 @@
...
+		InitFileHandler()
+	})
+}
+
+func TestInitFileHandler_LocalFilesystem(t *testing.T) {
```
kolaente, line 71: The same comments apply here as for the other test.

pkg/files/filehandling.go
```diff
@@ -46,8 +50,39 @@ func setDefaultConfig() {
 
 // InitFileHandler creates a new file handler for the file backend we want to use
 func InitFileHandler() {
-	fs = afero.NewOsFs()
-	afs = &afero.Afero{Fs: fs}
+	fileType := config.FilesType.GetString()
+	
+	if fileType == "s3" {
+		// Get S3 configuration
```
kolaente, line 56: Please refactor this (and the local fs usage) into a new function, then use a switch case to create the handler depending on the config

pkg/files/filehandling.go
```diff
@@ -46,8 +50,39 @@ func setDefaultConfig() {
...
+		endpoint := config.FilesS3Endpoint.GetString()
+		bucket := config.FilesS3Bucket.GetString()
+		region := config.FilesS3Region.GetString()
+		accessKey := config.FilesS3AccessKey.GetString()
+		secretKey := config.FilesS3SecretKey.GetString()
+
+		if endpoint == "" || bucket == "" || accessKey == "" || secretKey == "" {
+			log.Fatal("S3 configuration incomplete. Please set files.s3.endpoint, files.s3.bucket, files.s3.accesskey, and files.s3.secretkey")
```
kolaente, line 64: Why not check each file individually?

pkg/files/filehandling.go
```diff
@@ -46,8 +50,39 @@ func setDefaultConfig() {
...
+		}
+
+		// Create AWS session for afero-s3
+		sess, err := session.NewSession(&aws.Config{
+			Region:      aws.String(region),
+			Credentials: credentials.NewStaticCredentials(accessKey, secretKey, ""),
+			Endpoint:    aws.String(endpoint),
+			S3ForcePathStyle: aws.Bool(true), // Required for some S3-compatible services
```
kolaente, line 72: Why is this necessary? Should it be user-configurable?

zarelit: @kolaente (not the author) the force path style is commonly needed by selfhosters because serving buckets using path-style is really easier. This parameter is often overlooked

kolaente: Fine by me, but should it be configurable?

zarelit: Yes it should as some S3 services do only dns-style, some do only path-style.
I would probably change the the name to S3**Use**PathStyle instad of S3**Force**PathStyle like they [did in pulumi](https://github.com/pulumi/pulumi-aws/issues/3068) so that it doesn't look like one is forcing something bad, but that's nitpicking.

Issue #1545, opened 2025-09-25
[v1.0.0-rc2] Duplicated SQL condition proposed fix


Mattiam2:
I've noticed debugging the application in the last Vikunja version (v1.0.0-rc2) that when creating a filter with more than two conditions in && or ||, the generated SQL query duplicates the condition in the middle.

For example, if the filter is: _A && B && C_

The generated SQL condition in "Where" is: _A && **B && B** && C_

Editing the code I managed to fix this issue by rewriting the final filter construction in: 
https://github.com/go-vikunja/vikunja/blob/8c4fc4780ed1d10d667bb57f3c32ef472844254e/pkg/models/task_search.go#L228-L243

My proposed fix changes that code block to:

```
if len(dbFilters) > 0 {
	filterCond = dbFilters[0]
	if len(dbFilters) >= 1 {
		for i := range dbFilters {
			if len(dbFilters) > i+1 {
				switch rawFilters[i+1].join {
				case filterConcatOr:
					filterCond = builder.Or(filterCond, dbFilters[i+1])
				case filterConcatAnd:
					filterCond = builder.And(filterCond, dbFilters[i+1])
				}
			}
		}
	}
}
```

kolaente:
Do you want to send a PR?

Mattiam2:
Created the pull request

PR #1546, opened 2025-09-25
fix: prevent duplicated sql condition in filters

Mattiam2:
Proposing the fix as in #1545


kolaente:
Hey sorry for the long silence! If you're still interested in this, please check why the tests are failing so that we can merge this.

Thanks!

Issue #1549, opened 2025-09-26
Unable to confirmed email


Alfly-Alyx:
### Description

I got “Email address of the user not confirmed.” when trying to logging new user.

I create a new user, I well received the email, I clicked on the link, then I am redirected to the frontend logging page. I try to log-in but got this message.

Vikunja 1.0-RC2 server installed locally on Windows, and the client is 1.0-RC2.

### Vikunja Version

1.0-RC2

### Can you reproduce the bug on the Vikunja demo site?

Please select

kolaente:
What's the url you get in the email?

Alfly-Alyx:
This kind of URL :

http://myURL(example):3456/?userEmailConfirm=3ZzeAXfCiG4tVtDNuw7LlTAVTYrxjCvzyPfmAStWJ7nYU6sABzMvznGdId3H82ca

kolaente:
Fixed in https://github.com/go-vikunja/vikunja/pull/1575, please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

Alfly-Alyx:
Thanks, I will try it.

PR #1575, opened 2025-09-30
fix(user): race condition during email confirmation

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1549


kolaente:
@codex review

Issue #1494, opened 2025-09-12
File uploads/attachments fail at 32MB despite 500MB configuration


jeroenmarchand:
## Vikunja Version
`v1.0.0-rc2`
image: vikunja/vikunja:unstable
sha256:510cc702dfc385ed56295d4824648cd6f4b44b6095b266ce1aa7b1f62b6428d4

## Browser and version
N/A (affects web interface but also reproducible via API)

## OS and version
Docker (Linux containers)

## What is the expected behaviour?
Files should upload successfully up to the configured 500MB limit as indicated by the `/api/v1/info` endpoint.

## What is the actual behaviour?
File uploads fail with **HTTP 400 Bad Request** for files larger than exactly **33,554,432 bytes** (32MB), despite Vikunja showing a 500MB limit.

## Steps to reproduce the behaviour
1. Configure Vikunja with `VIKUNJA_FILES_MAXSIZE: 500MB`
2. Verify `/api/v1/info` endpoint shows `"max_file_size": "500MB"` 
3. Create test files:
   ```bash
   # Exactly 32MB (works)
   dd if=/dev/zero of=test-32MB.pdf bs=1048576 count=32
   # 32.5MB (fails)  
   dd if=/dev/zero of=test-32.5MB.pdf bs=1024 count=33280
   ```
4. Upload both files via web interface:
   - `test-32MB.pdf` (33,554,432 bytes) → Works
   - `test-32.5MB.pdf` (34,078,720 bytes) → Does not work

## Additional information
- **Exact boundary**: Files ≤ 33,554,432 bytes work, files > 33,554,432 bytes fail
- **32MB significance**: 33,554,432 equals `32 << 20` (common default for HTTP parsing)
- **No reverse proxy**: Direct access to Vikunja
- **Configuration verified**: `/api/v1/info` correctly shows 500MB limit

This suggests the issue may be related to HTTP framework multipart parsing limits that need to be aligned with Vikunja's configured file size limit.

### Vikunja Version

v1.0.0-rc2

### Browser and version

N/A (affects web interface but also reproducible via API)

### Can you reproduce the bug on the Vikunja demo site?

No

kolaente:
Will be fixed by https://github.com/go-vikunja/vikunja/pull/1577

kolaente:
Please check with the next unstable build (should be ready for deployment in ~30min, also on [try](https://try.vikunja.io)).

PR #1577, opened 2025-09-30
fix(attachments): extend upload file size to form data

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1494


kolaente:
@codex review

PR #1591, opened 2025-10-05
fix: correct case-sensitivity in duedate time parsing

### LINKED ISSUE #1590: Creating tasks at xPM are created at xAM
### Description

While the time parsing is correctly case insensitive here:
https://github.com/go-vikunja/vikunja/blob/5f1d684551fee62f8a2c299d206eb02045a5408d/frontend/src/helpers/time/parseDate.ts#L112-L113

Later in the function you are specifically checking for `pm` and not handling `PM`:
https://github.com/go-vikunja/vikunja/blob/5f1d684551fee62f8a2c299d206eb02045a5408d/frontend/src/helpers/time/parseDate.ts#L121

### Vikunja Version

`v0.24.6` and also on the demo site running `v1.0.0-rc2-75-5f1d6845`

### Browser and version

_No response_

Yes

### Screenshots

Consequently the following example: 
Image

Leads to
Image

Requiring the user to manually adjust the time from `05:00` back to `17:00` as desired.

mechanarchy:
This PR resolves the parsing of `PM` (in capitals) which was previously not handled correctly. Fixes #1590.

Also zeroes out the milliseconds value on due dates, for cleanliness.

### DIFF
--- a/frontend/src/helpers/time/parseDate.ts
+++ b/frontend/src/helpers/time/parseDate.ts
@@ -109,7 +109,7 @@ const addTimeToDate = (text: string, date: Date, previousMatch: string | null): 
 		}
 	}
 
-	const timeRegex = ' (at|@) ([0-9][0-9]?(:[0-9][0-9]?)?( ?(a|p)m)?)'
+	const timeRegex = ' (at|@) ([0-9][0-9]?(:[0-9][0-9])?( ?(a|p)m)?)'
 	const matcher = new RegExp(timeRegex, 'ig')
 	const results = matcher.exec(text)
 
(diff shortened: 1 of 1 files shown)


mechanarchy:
I added a second commit which will require users to give two digits for minutes, refer: https://github.com/go-vikunja/vikunja/issues/1590#issuecomment-3368671256

mechanarchy:
I'll have a crack at it, but I'm not a frontend developer by any means.

kolaente:
It's probably enough if you add more test cases here: https://github.com/go-vikunja/vikunja/blob/main/frontend%2Fsrc%2Fmodules%2FparseTaskText.test.ts#L88

kolaente:
Follow-up in https://github.com/go-vikunja/vikunja/pull/1613 due to CI issues.

Issue #28, opened 2022-05-16
Deleting Task Doesn't Update Timestamp


typkrft:
I may be wrong, but it seems that deleting a task doesn't update the tasklist `updated` timestamp. Creating a new task does.

I'm writing a script that uses the api to send notifications when a list has been `updated`. Maybe I'm missing something. I'm not a developer and this may be r/badcode lol.

example code:
```PYTHON
import requests
from time import sleep
import json


def get_token(url, username, password):
    credentials = {"username": f"{username}", "password": f"{password}"}
    token = requests.post(f"{url}/api/v1/login", json=credentials).json()["token"]
    return (url, token)


def get_tasks(auth):
    db = {}

    tasklists = requests.get(
        f"{auth[0]}/api/v1/lists", headers={"Authorization": f"Bearer {auth[1]}"}
    ).json()
    tasks = requests.get(
        f"{auth[0]}/api/v1/tasks/all", headers={"Authorization": f"Bearer {auth[1]}"}
    ).json()

    for tasklist in tasklists:
        db[tasklist["id"]] = dict(
            name=tasklist["title"], updated=tasklist["updated"], tasks={}, num_tasks=0
        )

    for task in tasks:
        db[task["list_id"]]["tasks"][task["id"]] = dict(
            name=task["title"],
            created=task["created"],
            done=task["done"],
            done_ts=task["done_at"],
            created_by=task["created_by"]["name"],
            updated=task["updated"],
            labels=task["labels"],
            priority=task["priority"],
            notes=task["description"],
        )

    for tasklist in tasklists:
        db[tasklist["id"]]["num_tasks"] = len(db[tasklist["id"]]["tasks"])

    return db


def main():
        auth = get_token("https://vikunja.tld", "user", "password")
    current_db = get_tasks(auth)

    while True:
        sleep(30)
        auth = get_token("https://vikunja.tld", "user", "password")
        new_db = get_tasks(auth)

        for tasklist in current_db.keys():

            # Get Deleted Tasks
            # FIXME: Should be able to check if tasklist has been updated first to avoid unnecesary runs
            deleted_tasks = [
                ("current_db", tasklist, task)
                for task in current_db[tasklist]["tasks"]
                if task not in new_db[tasklist]["tasks"]
            ]
            if len(deleted_tasks) > 0:
                print(deleted_tasks)

            # Check if tasklist has been updated
            if current_db[tasklist]["updated"] == new_db[tasklist]["updated"]:
                continue

[...]

PR #1638, opened 2025-10-10
fix(kanban): repeating tasks dates won't update when moved in done bucket

Biagio00:
This is my first pull request ever. Suggestions on how to do better are welcome :)

When moving a repeating task in done bucket it doesn't update start/due/end dates. In the code it seems that you were trying to create a copy of the task to be used as `oldTask` that has `Done` set to `false`. Because `task` is a pointer, doing `oldTask.Done = false` changes the done state in `task` too, causing the `updateDone` function call to exit without changing dates.
My fix just uses golang native deep copy that is enough for `updateDone` calculations that should not need any pointer-property of Task.

Example with new behaviour:
- Task configuration:
  immagine

- Task in kanban:
  immagine

- Task when moved to done column:
  immagine

Because the task doesn't go back to the original bucket immediately (it sticks in the done bucket until refresh), picking it up again from the done bucket and releasing it there causes the dates to update again. This might be a thing to consider but i feel it's unrelated from the fix in this pull request.

### DIFF
--- a/pkg/models/kanban_task_bucket_test.go
+++ b/pkg/models/kanban_task_bucket_test.go
@@ -142,7 +142,7 @@ func TestTaskBucket_Update(t *testing.T) {
 			ProjectID:     1, // In actual web requests set via the url
 		}
 
-		// Before running the TaskBucket Update we retrive the task and execute 
+		// Before running the TaskBucket Update we retrieve the task and execute
 		// an updateDone to obtain the task with updated start/end/due dates
 		// This way we can later match them with what happens after running TaskBucket Update
 		u := &user.User{ID: 1}
@@ -170,7 +170,7 @@ func TestTaskBucket_Update(t *testing.T) {
 			"task_id":   1,
 			"bucket_id": 1,
 		}, false)
-		
+
 		assert.Equal(t, updatedTask.DueDate.Unix(), tb.Task.DueDate.Unix())
 		assert.Equal(t, updatedTask.StartDate.Unix(), tb.Task.StartDate.Unix())
 		assert.Equal(t, updatedTask.EndDate.Unix(), tb.Task.EndDate.Unix())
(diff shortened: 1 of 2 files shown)


Biagio00:
I changed the current test for repeating task in kanban done bucket `TestTaskBucket_Update/moving_a_repeating_task_to_the_done_bucket`. Now it creates a copy of the task and passses it to `updateDone` (which is tested in another section so we assume is working), then it checks for equality of start, end and due dates between the outcome of `updateDone` and `TaskBucket.Update` functions.
I also needed to change some other tests because we now have `task #28` with `due_date`.

Biagio00:
I'm sorry i'm not very familliar with these tools. Can you help me understand what's wrong with those failing api-lint and e2e(1)?

kolaente:
The e2e test is probably flakey, I've started it again.

For the lint failure, you want to run them locally to fix the issues: https://vikunja.io/docs/magefile/#check (there are instructions in that page as well for how to set up the mage command)

Biagio00:
I should have fixed the lint issues it was reporting.

Biagio00:
I ran the tests on my pc to see if i can reproduce the issue. It seems that the result is pretty random. I'll attach here the screens with the successfull results (haven't changed a single line of code).
I see a lot of PRs getting slowed down by these flawed tests, I think it's important that they get fixed. Shall I open an issue on this topic? I've never used cypress before, maybe someone could fix this?
task spec ts-provides back navigation to the project in the kanban view on mobile
task spec ts-does not provide back navigation to the project in the kanban view on desktop
project-view-table spec ts

Biagio00:
It seems I can't re-run checks on my own. Can you please run them again? Shall I update the branch with latest changes? @kolaente

kolaente:
The test failure seems unrelated, I've merged this now. Thanks!

Issue #1637, opened 2025-10-10
saved Filter missing back to Project / Filter link and Filter cant be opened after opening task


NeptunedGmbH:
### Description

When opening a a task from a saved filter I think you implemented a back to project button a few weeks ago thats gone again and also the filter wont open again after opening a task. See in the loom attached
Issue is primarily with Kanban View
https://www.loom.com/share/71845d0cbc2f4cebaf4b9b17241f0f69?sid=2280ab24-09de-4cee-a5b7-92a231afd816

### Vikunja Version

Version: v1.0.0-rc2-129-cf9e5d08

### Browser and version

Arc  1.115.0

### Can you reproduce the bug on the Vikunja demo site?

Yes

PR #1642, opened 2025-10-11
fix(project): correctly set last project when navigating from a saved filter

kolaente:
Resolves https://github.com/go-vikunja/vikunja/issues/1637


kolaente:
@codex review

PR #1649, opened 2025-10-13
fix: show cross-project subtasks in List view

### LINKED ISSUE #782: list and table views do not show (sub)tasks if the parent task is in anothe project
### Description

If I move subtask from the project where the parent task is to another project, the subtask appears in the kaban and gantt views of that other project but not in the list or table view. It however still (usefully) remains in the list and table view of the original project where the parent task is.

I would expect (and it would be good) that the task appear in all views of the project it has been moved to (but I'd say it makes sense if it also stays in the list view of the project the parent task is).

For more on the topic of small inconsistencies I encountered in Vikunja in the logic and managements of the hierarchy between project/subprojects/tasks/subtasks, I finally found the time to write a little piece here: https://community.vikunja.io/t/improvement-suggestions/3732.

### Vikunja Version

0.24.6

### Browser and version

_No response_

Yes

### Screenshots

Image
Image
Image
Image

kolaente:
## Summary

Fixes #782 - Cross-project subtasks now remain visible in their own project's List view.

Previously, subtasks were hidden in List view if they had ANY parent task, regardless of which project the parent belonged to. This caused cross-project subtasks to disappear from their own project's List view, while they remained visible in Table and Kanban views.

## Changes

- **Root Cause:** The List view component filtered out all tasks with parent tasks, without checking if the parent was in the same project
- **Solution:** Updated filter logic to only hide subtasks when their parent task is also visible in the current view (same project)
- **Implementation:** 
  - Created `shouldShowTaskInListView()` composable with project-aware filtering logic
  - Updated `ProjectList.vue` to use the new filtering function
  - Added comprehensive unit tests (8 tests covering all scenarios)
  - Added E2E tests for cross-project and same-project subtask visibility

### Unit Tests (8 new tests)
- ✅ Hide subtasks when parent is in the same project
- ✅ Show subtasks when parent is in a different project
- ✅ Show tasks with no parents
- ✅ Handle undefined/empty relatedTasks
- ✅ Handle multiple levels of nesting within same project
- ✅ Show tasks with multiple parents when none are in view
- ✅ Hide tasks with multiple parents when at least one is in view

### E2E Tests
- ✅ Cross-project subtasks visible in their own project's List view
- ✅ Same-project subtasks hidden at top level (nested under parent)

## Verification

- ✅ All 706 unit tests pass
- ✅ Linting passes (ESLint + Stylelint)
- ✅ Frontend builds successfully
- ✅ No type errors in modified files

## Example Scenario

**Before:**
1. Create Task A in Project 1
2. Create Task B in Project 2
3. Make Task B a subtask of Task A
4. ❌ Task B disappears from Project 2's List view

**After:**
1. Create Task A in Project 1
2. Create Task B in Project 2
3. Make Task B a subtask of Task A
4. ✅ Task B remains visible in Project 2's List view
5. ✅ Task A shows Task B as nested subtask in Project 1's List view

### DIFF
--- a/frontend/cypress/e2e/project/project-view-list.spec.ts
+++ b/frontend/cypress/e2e/project/project-view-list.spec.ts
@@ -185,8 +185,12 @@ describe('Project View List', () => {
 
 		cy.get('.tasks')
 			.should('contain', 'Parent Task')
+		cy.get('.tasks')
+			.should('contain', 'Subtask Same Project')
 
-		cy.get('ul.tasks > div')
-			.should('have.length', 2)
+		cy.get('ul.tasks > div > .single-task')
+			.should('exist')
+		cy.get('ul.tasks > div > .subtask-nested')
+			.should('exist')
 	})
 })


kolaente:
@codex review

PR #1024, opened 2025-06-25
feat(list view): add sorting options

kolaente:
## Things left to do:

- [ ] SortPopup Background - use a model, not a delete modal
- [ ] Ascending/Descending is bad ux
- [ ] "Position" should be "manually"
- [ ] When sorting for position, always sort asc, not allow selecing
- [ ] Items in the dropdown should be sorted alphabetically
- [ ] Spacing for the sort options button to the left
- [ ] Text in the popup to explain what it does
- [x] Better layout for the input

Originally: https://chatgpt.com/codex/tasks/task_e_684fe97e3aa48322bb03dd5120a1977c

Resolves https://github.com/go-vikunja/vikunja/issues/1729#issuecomment-3472472918

PR #1652, opened 2025-10-13
feat: add sorting controls to Overview page

2ZZ:
## Summary
- Added ability to sort tasks by multiple fields on the Overview page
- Users can select from 6 sort options: Due Date, Priority, Title, Start Date, End Date, Done
- Users can toggle between ascending and descending order
- Sort preferences persist in URL query parameters

## Changes
- Added dropdown select and outlined button positioned to the right of "Current Tasks" title
- Dropdown allows choosing sort field
- Button toggles between Ascending and Descending order
- Clean design matching the rest of the UI
- Only visible on Overview page (not on date-filtered views)

## Technical Details
- Added `priority` to `TaskFilterParams` sort_by type options in `taskCollection.ts`
- Sort state managed with reactive refs in `ShowTasks.vue`
- Translation keys added: `task.show.sortBy`, `misc.ascending`, `misc.descending`
- E2E tests added to verify sorting functionality

## Test plan
- [x] Verify sort controls appear on Overview page
- [x] Test sorting by all available fields (due date, priority, title, start date, end date, done)
- [x] Test toggling between ascending and descending order
- [x] Verify sort preferences persist in URL and across page reloads
- [x] Added Cypress E2E tests for sorting functionality

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -18,6 +18,7 @@ coverage
 .vite/
 
 # Test files
+cypress/downloads
 cypress/screenshots
 cypress/videos
 
--- a/frontend/cypress/e2e/project/filter-persistence.spec.ts
+++ b/frontend/cypress/e2e/project/filter-persistence.spec.ts
@@ -66,14 +66,96 @@ describe('Filter Persistence Across Views', () => {
 	it('should handle URL sharing with filters', () => {
 		// Visit URL with pre-existing filter parameters
 		cy.visit('/projects/1/4?filter=done%3Dtrue&s=Test')
-		
+
 		// Verify URL parameters are preserved
 		cy.url().should('include', 'filter=done%3Dtrue')
 		cy.url().should('include', 's=Test')
-		
+
 		// Switch views and verify parameters persist
 		cy.visit('/projects/1/3?filter=done%3Dtrue&s=Test')
 		cy.url().should('include', 'filter=done%3Dtrue')
 		cy.url().should('include', 's=Test')
+	})
+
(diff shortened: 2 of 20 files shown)


kolaente:
This seems duplicate with https://github.com/go-vikunja/vikunja/pull/1024, can you explain what you did differently and why? What problem are you solving? Any issues this solves?

Please also add screenshots since this affects UI.

2ZZ:
> This seems duplicate with #1024, can you explain what you did differently and why? What problem are you solving? Any issues this solves?
> 
> Please also add screenshots since this affects UI.

I didn't know 1024 existed though this is for the Overview page rather than the list view.
I personally just want to be able to see all my tasks sorted by priority.

image

image

kolaente:
As you (or Claude) have implemented it, this needs a few modifications.

“Ascending” or “descending” is not exactly great ux. For people who know how databases work, it makes completely sense, but for everyone else, it’s difficult. Better would be to call this something like “low to high” or similar, ideally combined with the actual values this is sorting by. I’m sure you can find good examples of this with a little research.

Then, the sorting should be applied in all views, not only the overview. This also applies to saved filters.

Finally, the size of the button and drop-down is not the same.

How are other applications solving this?

jelmervdm:
Instead of a separate "high to low", "low to high" button, which doesn't make much sense if sorting alphabetically, maybe it would make sense to have the options in the drop down eg.:

- Sort A to Z
- Sort Z to A
- Sort by priority high to low
- Sort by priority low to high
- Sort by due date (Nearest first)
- Sort by due date (Farthest first)

etc.

PR #1688, opened 2025-10-20
feat: add S3 file storage support

maggch97:
Fixes https://github.com/go-vikunja/vikunja/issues/1469

This PR is based on https://github.com/go-vikunja/vikunja/pull/1509. The account of original author has been deleted, so I continue the work.

image

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -271,6 +271,7 @@ jobs:
           go-version: stable
       - name: test S3 file storage integration
         env:
+          VIKUNJA_TESTS_USE_CONFIG: 1
           VIKUNJA_DATABASE_TYPE: sqlite
           VIKUNJA_FILES_TYPE: s3
           VIKUNJA_FILES_S3_ENDPOINT: http://localhost:9000
--- a/pkg/files/s3_integration_test.go
+++ b/pkg/files/s3_integration_test.go
@@ -33,9 +33,6 @@ import (
 // with S3/MinIO storage backend. This test specifically validates S3 functionality
 // and will fail if S3 is not properly configured.
 func TestFileStorageIntegration(t *testing.T) {
-	// Initialize config to load environment variables
-	config.InitConfig()
-
 	// Ensure S3 is configured for this test
 	if config.FilesType.GetString() != "s3" {
 		t.Skip("Skipping S3 integration tests - VIKUNJA_FILES_TYPE must be set to 's3'")


maggch97:
Some optimization could be done in future PR:
- "github.com/aws/aws-sdk-go/aws" is deprecated: Deprecated: aws-sdk-go is deprecated. Use aws-sdk-go-v2.
See https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/.deprecated[default](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/deprecated)
- Replace server-proxied file downloads with S3 pre-signed URLs to eliminate bandwidth overhead and reduce latency by allowing direct client-to-S3 access.

maggch97:
@kolaente please help review

kolaente:
> "github.com/aws/aws-sdk-go/aws" is deprecated: Deprecated: aws-sdk-go is deprecated. Use aws-sdk-go-v2.

If it's already deprecated, please use the newer version.

pkg/files/filehandling.go
```diff
@@ -44,18 +50,79 @@ func setDefaultConfig() {
...
+func InitFileHandler() error {
+	fileType := config.FilesType.GetString()
+
+	var err error
+	switch fileType {
+	case "s3":
+		err = initS3FileHandler()
+	case "local":
+		initLocalFileHandler()
+	default:
```
kolaente, line 110: This should fail, because the configured file type is invalid.

pkg/files/s3_integration_test.go
```diff
@@ -0,0 +1,131 @@
+// Vikunja is a to-do list application to facilitate your life.
```
kolaente, line 1: Please add some end to end tests here for storing and retrieving files from s3 and locally, depending on the config. It should be possible to add minio to the CI, similar to how we're testing the ldap integration.

config-raw.json
```diff
@@ -484,6 +484,47 @@
                     "key": "maxsize",
                     "default_value": "20MB",
                     "comment": "The maximum size of a file, as a human-readable string.\nWarning: The max size is limited 2^64-1 bytes due to the underlying datatype"
+                },
+                {
+                    "key": "type",
+                    "default_value": "local",
+                    "comment": "The type of file storage backend. Supported values are 'local' and 's3'."
```
kolaente:
```suggestion
                    "comment": "The type of file storage backend. Supported values are `local` and `s3`."
```

kolaente (changes requested):

kolaente:
Looks pretty good already!

maggch97:
> If it's already deprecated, please use the newer version.

It depends on `afero-s3`, it still uses the v1 sdk. https://github.com/fclairamb/afero-s3/pull/385 There's an unmerged PR.


From https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/
> Existing applications that use AWS SDK for Go (v1) will continue to function as intended, unless there is a fundamental change to how an AWS service works. This is uncommon and would be broadly communicated if it happens. Between July 31, 2024 and end-of-support on July 31, 2025, the AWS SDK for Go (v1) will only receive critical bug fixes and security updates. The SDK will not be updated to support new AWS services, new service features, or changes to existing services.

Considering that S3 is a very stable protocol, v1 SDK should work for a long time.

pkg/files/s3_integration_test.go
maggch97: Sure, I will refer to the existing e2e and add a new test with minio

kolaente:
Ah I see, let's keep v1 of the SDK then.

pkg/files/s3_integration_test.go
maggch97: image

Compared with minio/minio, bitnami/minio adds the ability to create a default bucket. The test uses bitnamilegacy/minio because both Bitnami and MinIO recently stopped maintaining their docker images, and bitnami/minio was archived to bitnamilegacy/minio.
In future, we may need to switch to a more actively maintained S3-compatible project, but for now, we only use basic read/write functionality, so it‘s not a big issue.

.github/workflows/test.yml
```diff
@@ -240,6 +248,13 @@ jobs:
           VIKUNJA_AUTH_LDAP_BINDPASSWORD: password
           VIKUNJA_AUTH_LDAP_USERFILTER: "(&(objectclass=inetorgperson)(uid=%s))"
           VIKUNJA_SERVICE_PUBLICURL: http://127.0.0.1:3456
+          VIKUNJA_FILES_TYPE: s3
```
kolaente, line 276: Can this be local by default? Then in the s3 integration test, set it to `s3` and maybe fail when no endpoint is configured.

Right now it would be possible to run the s3 integration tests with local storage, which does not really make sense.

pkg/files/s3_integration_test.go
kolaente: Can we use [garage](https://garagehq.deuxfleurs.fr/documentation/cookbook/real-world/#get-a-docker-image) instead?

We will need to switch this at some point anyway and since we're integrating this now, better do it right directly.

.github/workflows/test.yml
maggch97: image

pkg/files/s3_integration_test.go
```diff
@@ -0,0 +1,329 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestFileStorageIntegration tests end-to-end file storage and retrieval
+// with S3/MinIO storage backend. This test specifically validates S3 functionality
+// and will fail if S3 is not properly configured.
+func TestFileStorageIntegration(t *testing.T) {
+	if os.Getenv("VIKUNJA_TESTS_S3") != "1" {
```
kolaente: please use `config.FilesType.GetString() != "s3"` and then in the ci set the `VIKUNJA_TESTS_USE_CONFIG` environment variable

pkg/files/s3_integration_test.go
```diff
@@ -0,0 +1,329 @@
...
+		t.Skip("Skipping S3 integration tests - VIKUNJA_FILES_TYPE must be set to 's3'")
+	}
+
+	// Validate S3 configuration is present
+	if config.FilesS3Endpoint.GetString() == "" {
+		t.Fatal("S3 integration test requires VIKUNJA_FILES_S3_ENDPOINT to be set")
+	}
+
+	// Save original config values
+	originalType := config.FilesType.GetString()
```
kolaente: is this needed?

pkg/files/s3_integration_test.go
```diff
@@ -33,16 +33,11 @@ import (
 // with S3/MinIO storage backend. This test specifically validates S3 functionality
 // and will fail if S3 is not properly configured.
 func TestFileStorageIntegration(t *testing.T) {
-	if os.Getenv("VIKUNJA_TESTS_S3") != "1" {
-		t.Skip("Skipping integration tests - set VIKUNJA_TESTS_S3=1 to run")
-	}
-
 	// Initialize config to load environment variables
 	config.InitConfig()
```
kolaente: This will be done by the config init function (triggered by `VIKUNJA_TESTS_USE_CONFIG`)

kolaente (approved):
Thanks for all the effort you put in!

kolaente:
Test failure seems unrelated.

PR #1754, opened 2025-11-05
feat: enable user mentions in task description & comments

maggch97:
image
image

Resolves https://github.com/go-vikunja/vikunja/issues/572


maggch97:
fix: https://github.com/go-vikunja/vikunja/issues/572

frontend/src/components/input/editor/MentionList.vue
```diff
@@ -0,0 +1,178 @@
...
+						@{{ item.username }}
+					</p>
+				</div>
+			</button>
+		</template>
+		<div
+			v-else
+			class="mention-item no-results"
+		>
+			No users found
```
maggch97: It should be a localized string

frontend/src/components/input/editor/MentionList.vue
```diff
@@ -0,0 +1,178 @@
...
+				@click="selectItem(index)"
+			>
+				<img
+					:src="item.avatarUrl"
+					alt=""
+					class="mention-avatar"
+				>
+				<div class="mention-info">
+					<p class="mention-name">
+						{{ item.label }}
```
kolaente, line 18: If the user did not set a name, the label will be the same as the username, maybe we can just hide it in that case?

frontend/src/components/input/editor/mentionSuggestion.ts
```diff
@@ -0,0 +1,190 @@
...
+import type {Editor} from '@tiptap/core'
+
+import MentionList from './MentionList.vue'
+import ProjectUserService from '@/services/projectUsers'
+import {fetchAvatarBlobUrl, getDisplayName} from '@/models/user'
+import type {IUser} from '@/modelTypes/IUser'
+
+interface MentionItem {
+	id: number
+	label: string
```
kolaente, line 12: Since this will always be the user's display name, we might as well call it like that.

frontend/src/components/input/editor/mentionSuggestion.ts
```diff
@@ -0,0 +1,190 @@
...
+let cacheProjectId: number | null = null
+
+async function fetchUsersForProject(projectId: number): Promise<MentionItem[]> {
+	// Return cached users if we're asking for the same project
+	if (cacheProjectId === projectId && cachedUsers.length > 0) {
+		return cachedUsers
+	}
+
+	const projectUserService = new ProjectUserService()
+	const users = await projectUserService.getAll({projectId}, {}) as IUser[]
```
kolaente: This is paginated by default, if you cache it, you'll need to fetch all pages.

Another option would be to not cache it and use the search functionality from the server to look up users.

frontend/src/components/input/editor/TipTap.vue
```diff
@@ -804,6 +822,21 @@ watch(
 		border-radius: $radius;
 	}
 
+	// Mention styles
+	.mention {
+		background-color: var(--primary-light, #e3f2fd);
+		color: var(--primary, #1976d2);
```
kolaente: I think this should look differently. Right now the text is hardly readable:

Image

Please use the `grey-200` for the background and maybe `grey-800` for the text, normal font weight. Then, please replace the `@` with the user avatar - right now, the `@` implies a username but the display name is shown when the user has set one. Using the avatar makes it clear this is a user.

pkg/models/mentions.go
```diff
@@ -17,25 +17,68 @@
...
 	}
 
-	usernames := []string{}
-	for _, match := range matches {
-		usernames = append(usernames, strings.TrimPrefix(match, "@"))
+	return user.GetUsersByIDs(s, userIDs)
+}
+
+// extractMentionedUserIDs parses HTML content and extracts user IDs from mention spans.
+// It looks for <span class="mention" data-id="123"> elements and returns the user IDs.
```
kolaente, line 39: Wouldn't the old approach still work?

kolaente (changes requested):

pkg/models/mentions.go
maggch97: No, do you think we still need to support the regex match? The reason that I made this PR is that the previous regex match does not work if the username contains Chinese characters or symbols like @

When a user types @, he will see the mention suggestion menu. It's foreseeable for the user that the old way does not work now.

kolaente: Gotcha, then your approach makes more sense. Let's keep it. We would need a migration for this though.

frontend/src/components/input/editor/MentionList.vue
maggch97: image

Do you mean this?

kolaente: yes, this looks better!

frontend/src/components/input/editor/mentionSuggestion.ts
maggch97: It's required by `MentionNodeAttrs` from `@tiptap/extension-mention`
I updated the type to define to explicit extends
```
interface MentionItem extends MentionNodeAttrs {
	id: string
	label: string
	username: string
	avatarUrl: string
}
```

kolaente: Ah, that makes sense then.

frontend/src/components/input/editor/TipTap.vue
maggch97: image
image

kolaente: Looks better! Can you add the avatar as well?

pkg/models/mentions.go
maggch97: I prefer not to do the migration for previous mention:
1. The old data was based on regex matches rather than structured fields, database query is required to reduce false positives. Also, the mentions used to reference usernames, but now it shows display names — so the content has changed.
2. From a user experience perspective, mentions mainly exist to trigger notifications. The notifications for the old data have already been sent, so the migration doesn’t seem particularly valuable.

kolaente: I would counter that: The highlighting of usernames would be valuable to users.

Nevertheless, the feature is pretty much hidden so I'd expect not many people have used it anyways. Let's continue without the migration.

frontend/src/components/input/editor/TipTap.vue
```diff
@@ -480,6 +486,18 @@ const extensions : Extensions = [
 	PasteHandler,
 ]
 
+// Add mention extension if enabled
+if (props.enableMentions && props.mentionProjectId > 0) {
+	extensions.push(
+		Mention.configure({
```
kolaente, line 492: Can you change the rendering here so that it shows the user avatar?

https://tiptap.dev/docs/editor/extensions/nodes/mention#renderhtml

maggch97: @kolaente for almost all apps I'm using, mention will not show the avatar. I think we can follow the common pattern first.

kolaente: For me it's the other way around :) but okay, let's integrate the avatar later.

kolaente (commented):
Please resolve the conflicts, then we're ready to merge.

kolaente:
Thanks again!

PR #1823, opened 2025-11-15
add migrator for nextcloud deck app

kermorgant:
This PR is the result of asking claude code to write a migrator for nextcloud deck. I'm afraid it might hurt the eyes when reviewing, but thought it might still be worth sharing.

I managed to successfully import a dozen of boards from my self hosted nextcloud 30 instance, with attachments and comments.

I should be able to fix things in the near future, but can't promise any long term commitment.

# UX 

## Picking the migrator

(it is always enabled : there's no need for the admin to configure anything in the config.yml, so maybe it can be shown here unconditionally ?)
image

## Configuring 

(the redirect url is probably useless, since it seems to be enforced by nextcloud)

image

### DIFF
--- a/pkg/modules/migration/create_from_structure.go
+++ b/pkg/modules/migration/create_from_structure.go
@@ -459,6 +459,25 @@ func createProjectWithEverything(s *xorm.Session, project *models.ProjectWithTas
 			}
 			log.Debugf("[creating structure] Created new comment %d", comment.ID)
 		}
+
+		// Assignees
+		if len(t.Assignees) > 0 {
+			log.Debugf("[creating structure] Creating %d assignees for task %d", len(t.Assignees), t.ID)
+		}
+		for _, assignee := range t.Assignees {
+			if assignee == nil || assignee.ID == 0 {
+				continue
+			}
+			taskAssignee := &models.TaskAssginee{
+				TaskID: t.ID,
+				UserID: assignee.ID,
+			}
+			_, err = s.Insert(taskAssignee)
+			if err != nil {
+				return
+			}
+			log.Debugf("[creating structure] Assigned user %d to task %d", assignee.ID, t.ID)
+		}
 	}
 
 	// All tasks brought their own bucket with them, therefore the newly created default bucket is just extra space
--- a/pkg/modules/migration/deck/deck.go
+++ b/pkg/modules/migration/deck/deck.go
@@ -437,7 +437,7 @@ func (m *Migration) mapUser(s *xorm.Session, nextcloudUsername string, fallbackU
 	return fallbackUser
 }
 
-// resolveUserMappings resolves Nextcloud usernames stored in comments and attachments to Vikunja users
+// resolveUserMappings resolves Nextcloud usernames stored in comments, attachments, and assignees to Vikunja users
 // This is called after building the project structure but before insertion
 func (m *Migration) resolveUserMappings(projects []*models.ProjectWithTasksAndBuckets, fallbackUser *user.User) error {
 	// Create a database session for user lookups
@@ -448,9 +448,33 @@ func (m *Migration) resolveUserMappings(projects []*models.ProjectWithTasksAndBu
 	commentsTotal := 0
 	attachmentsMapped := 0
 	attachmentsTotal := 0
+	assigneesMapped := 0
+	assigneesTotal := 0
 
 	for _, project := range projects {
 		for _, task := range project.Tasks {
+			// Resolve task assignees
+			resolvedAssignees := []*user.User{}
+			for _, assignee := range task.Assignees {
+				assigneesTotal++
+				if assignee != nil && assignee.Username != "" && assignee.ID == 0 {
+					// This is a temporary user object with Nextcloud username
+					ncUsername := assignee.Username
+					vikunjaUser := m.mapUser(s, ncUsername, fallbackUser)
+
+					// Only add if we found a mapped user (not fallback)
+					// We don't want to assign tasks to the migrating user for unmapped users
+					if vikunjaUser.ID != fallbackUser.ID {
+						resolvedAssignees = append(resolvedAssignees, vikunjaUser)
+						assigneesMapped++
+						log.Debugf("[Deck Migration] Resolved assignee: NC user '%s' -> Vikunja user '%s' (ID: %d)", ncUsername, vikunjaUser.Username, vikunjaUser.ID)
+					} else {
+						log.Debugf("[Deck Migration] No mapping found for assignee NC user '%s', keeping in description only", ncUsername)
+					}
+				}
+			}
+			task.Assignees = resolvedAssignees
+
 			// Resolve comment authors
 			for _, comment := range task.Comments {
 				commentsTotal++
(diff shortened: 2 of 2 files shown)


pkg/modules/migration/handler/handler.go
```diff
@@ -55,6 +56,16 @@ func (mw *MigrationWeb) RegisterMigrator(g *echo.Group) {
 // AuthURL is the web handler to get the auth url
 func (mw *MigrationWeb) AuthURL(c echo.Context) error {
 	ms := mw.MigrationStruct()
+
+	// For POST requests, bind the request body to the migrator struct
+	// This allows migrators that need server URL to receive it
+	if c.Request().Method == http.MethodPost {
```
kolaente, line 65: What's the point of binding the request body and then not doing anything with it?

pkg/modules/migration/create_from_structure.go
```diff
@@ -307,7 +308,45 @@ func createProjectWithEverything(s *xorm.Session, project *models.ProjectWithTas
 		t.ProjectID = project.ID
 		originalBucketID := t.BucketID
 		t.BucketID = 0
-		err = t.Create(s, user)
+
+		// Debug logging for datetime preservation
```
kolaente: this should be removed

pkg/routes/routes.go
```diff
@@ -714,6 +715,17 @@ func registerMigrations(m *echo.Group) {
 		microsoftTodoMigrationHandler.RegisterMigrator(m)
 	}
 
+	// Nextcloud Deck
+	// Note: Unlike other OAuth migrators, Deck is always registered because it requires
```
kolaente: If it's alwasy registered, the config should default to true

pkg/swagger/docs.go
```diff
@@ -765,6 +765,125 @@ const docTemplate = `{
                 }
             }
         },
+        "/migration/deck/auth": {
```
kolaente: The changes to the docs should be removed, they are automatically generated by CI

config-raw.json
```diff
@@ -606,6 +606,31 @@
...
+                {
+                    "key": "deck",
+                    "children": [
+                        {
+                            "key": "enable",
+                            "default_value": "false",
+                            "comment": "Whether to enable the Nextcloud Deck migrator."
+                        },
+                        {
+                            "key": "clientid",
```
kolaente: But the point of the new structure is to add the nextcloud login url in the UI? Why do I need to provide a client id and secret in the config?

kolaente (changes requested):
Thanks for being upfront about Claude. This does need more work if we want to integrate it.

Another option would be to rework this so that it can handle a nextcloud export? That would be possible with less changes to the migrator structure.

kolaente:
Can you explain how the flow works in general?

kermorgant:
> Thanks for being upfront about Claude. This does need more work if we want to integrate it.

Sure, that was the least I could do. I took a deliberate position of not trying to understand the codebase when starting that attempt (I was ready to give up if that did not work, and I have limited time & energy for such side work, just to give a bit of my context). But I got something working and thanks to your feedback, it feels like this could be worth a bit more effort :-) 

> Another option would be to rework this so that it can handle a nextcloud export? That would be possible with less changes to the migrator structure.

That was something I considered at the beginning but Claude warned against this being a dead end for getting attachments. Not sure how much truth there's in that, but that led me where we are here :-)

pkg/routes/routes.go
kermorgant: That was a mistake. The config has a boolean to enable the migrator, so I'll change the code

config-raw.json
kermorgant: True, this was not thought through. I guess this PR would be simpler if we put everything in the config ? I'll probably give this a shot unless you see this differently ?

kolaente: I like the idea of letting users migrate from any instance. That would mean users should be able to enter everything themselves, or just provide an exported file.

kermorgant: agree, that's better. I'll try. But I give up with exported file. I figured I had to install an app (on NC), but once installed there was some issue...

frontend/src/views/migrate/MigrationHandler.vue
```diff
@@ -28,7 +28,7 @@
 						{{ $t('migrate.upload') }}
 					</XButton>
 				</template>
-				<template v-else-if="migrator.requiresServerURL">
+				<template v-else-if="migrator.requiresCustomOAuth">
```
kermorgant: fields shown below are very "deck" specific, so this wording is too broad. `migrator.isNextcloudDeck` would feel better

frontend/src/views/migrate/MigrationHandler.vue
```diff
@@ -207,9 +294,19 @@ useTitle(() => t('migrate.titleService', {name: migrator.value.name}))
 async function startOAuthFlow() {
 	try {
 		migrationError.value = ''
-		// Store server URL in session storage so we can retrieve it after OAuth callback
+		// Store OAuth config in session storage so we can retrieve it after OAuth callback
 		sessionStorage.setItem('deckServerUrl', serverUrl.value)
-		const {url} = await migrationService.getAuthUrl(serverUrl.value)
+		sessionStorage.setItem('deckClientId', clientId.value)
+		sessionStorage.setItem('deckClientSecret', clientSecret.value)
```
kermorgant, line 291: not a good idea I feel, I'll remove that line

frontend/src/views/migrate/MigrationHandler.vue
```diff
@@ -222,12 +319,19 @@ async function initMigration() {
...
-	// For migrators that require server URL, restore it from session storage
-	if (migrator.value.requiresServerURL) {
+	// For migrators that require custom OAuth, restore config from session storage
+	if (migrator.value.requiresCustomOAuth) {
 		const storedServerUrl = sessionStorage.getItem('deckServerUrl')
-		if (storedServerUrl) {
-			serverUrl.value = storedServerUrl
-		}
+		const storedClientId = sessionStorage.getItem('deckClientId')
+		const storedClientSecret = sessionStorage.getItem('deckClientSecret')
```
kermorgant, line 316: same

frontend/src/views/migrate/MigrationHandler.vue
```diff
@@ -222,12 +319,19 @@ async function initMigration() {
...
+		const storedRedirectUrl = sessionStorage.getItem('deckRedirectUrl')
+		const storedUserMappings = sessionStorage.getItem('deckUserMappings')
+
+		if (storedServerUrl) serverUrl.value = storedServerUrl
+		if (storedClientId) clientId.value = storedClientId
+		if (storedClientSecret) clientSecret.value = storedClientSecret
```
kermorgant, line 321: same

pkg/routes/api/v1/info.go
```diff
@@ -147,7 +147,8 @@ func Info(c echo.Context) error {
 		m := &microsofttodo.Migration{}
 		info.AvailableMigrators = append(info.AvailableMigrators, m.Name())
 	}
-	if config.MigrationDeckEnable.GetBool() {
+	// Deck migrator is always available (no config needed - OAuth credentials provided via UI)
```
kermorgant: remove this comment ?

kolaente:
hey @kermorgant what's the status here? Asking because it's still marked as a draft but seems done, is this ready to review?

kermorgant:
> hey @kermorgant what's the status here? Asking because it's still marked as a draft but seems done, is this ready to review?

Hi, I'm sorry, it's unfortunately on hold due to lack of time. Too many things going on at the moment so I'm still using nextcloud deck at the moment but I haven't abandoned the idea (hoping to resume that during 2026, but feel free to close it for now if you like).

As far as I remember, I was indeed pretty close to get an acceptable outcome. I iterated on details such as user mappings and keeping original timestamps but I believe there was some detail (perhaps minor) I wanted to address.

kolaente:
Thanks for the update! I'll close this in the meantime, happy to accept a follow-up if anyone wants to implement this.

Issue #1670, opened 2025-10-16
Comment preview in task card


xeruf:
### Description

As we are using Vikunja as primary project management and cutting out duplicating the work updates on another tool, one small lacking feature has become apparent: The ability to see whether a task has comments.

Github has these nice little icons on issues, and they would really help:

<img width="90" height="216" alt="Image" src="https://github.com/user-attachments/assets/cb4965c8-bea1-4498-92ea-5fc9825e3bcc" />

Because for me I would like to see at a glance:
Where did something change, is there a question in a comment on a task I should respond to?
Ideally this could include unread tracking - if you haven't opened a task since the last comment was made, there could be a blue dot in this notification bubble.

This should be available in all views: Table, List, Kanban

### Which alternatives did you consider using instead?

clicking at each task to see if there's something new urgh
or writing an update in our chat tool saying: "please check task xy" - which is really the kind of overhead we want to eliminate ;)

NeptunedGmbH:
+1

mithileshgupta12:
Hi! I'd like to work on this feature. My plan is to:
- Add a comment count indicator to task cards across Kanban, List, and Table views
- Include unread tracking (visual indicator for tasks with new comments since last viewed)

Before I start, any guidance on:
- Existing patterns for tracking "last viewed" on tasks?
- Preferred icon/badge placement and styling?

Happy to discuss the approach further!

kolaente:
@mithileshgupta12 great!

I would split this in two PRs:

1. Comment count:
    * In the api, use an `expand` query on `/tasks` endpoints (look at how this is done for subtasks) to make comment count optional
    * Check if db indexes to make this fast exist
    * Once the api can optionally return the comments count, adjust the frontend to include the parameter when querying and show it
    * For styling, I would look at the "this task has a description" icon and what GitHub is doing and then do something similar
2. Unread tracking:
    * In the api, this should be a new endpoint to have more control over when a task is marked as read. Since it is an action that has side-effects, it should not happen as part of the GET request that happens when fetching the task.
    * We'll need to see how/if this interacts with marking notifications as read, maybe we can even reuse the existing notification things?

mithileshgupta12:
Ok got it. I'll work on this

mithileshgupta12:
I added the pr for the first task

https://github.com/go-vikunja/vikunja/pull/1771

kolaente:
[...]

PR #1857, opened 2025-11-21
feat: task unread tracking

mithileshgupta12:
Resolves https://github.com/go-vikunja/vikunja/issues/1670

### DIFF
--- a/frontend/src/services/task.ts
+++ b/frontend/src/services/task.ts
@@ -7,7 +7,7 @@ import LabelService from './label'
 import {colorFromHex} from '@/helpers/color/colorFromHex'
 import {SECONDS_A_DAY, SECONDS_A_HOUR, SECONDS_A_WEEK} from '@/constants/date'
 import {objectToSnakeCase} from '@/helpers/case'
-import { AuthenticatedHTTPFactory } from '@/helpers/fetcher'
+import {AuthenticatedHTTPFactory} from '@/helpers/fetcher'
 
 const parseDate = date => {
 	if (date) {


pkg/models/task_comments.go
```diff
@@ -97,6 +97,46 @@ func (tc *TaskComment) CreateWithTimestamps(s *xorm.Session, a web.Auth) (err er
 		}
 	}
 
+	project, err := GetProjectSimpleByTaskID(s, tc.TaskID)
```
kolaente: Can you do this in an event listener?

pkg/models/task_comments.go
```diff
@@ -97,6 +97,46 @@ func (tc *TaskComment) CreateWithTimestamps(s *xorm.Session, a web.Auth) (err er
...
+	if err != nil {
+		return err
+	}
+
+	projectUsers := []*ProjectUser{}
+	err = s.Where("project_id = ?", project.ID).
+		Find(&projectUsers)
```
kolaente: This won't work for users who have access through a parent project or team sharing. Please use the `ListUsersFromProject` function to find all users who should be notified.

Or (this is up for discussion): Only users who are subscribed to a project?

pkg/models/task_comments.go
```diff
@@ -97,6 +97,46 @@ func (tc *TaskComment) CreateWithTimestamps(s *xorm.Session, a web.Auth) (err er
...
+		}
+
+		exists, err := s.Where("task_id = ? AND user_id = ?", tc.TaskID, userIDToNotify).
+			Exist(&TaskUnreadStatus{})
+		if err != nil {
+			return err
+		}
+
+		if !exists {
+			_, err := s.Insert(&TaskUnreadStatus{
```
kolaente: This can be optimized further. Ideally use something like `ON CONFLICT DO NOTHING` but I don't know if xorm supports that. If it doesn't at least use only one `Insert` (you can pass a slice of structs and it will insert all of them at once).

pkg/models/tasks.go
```diff
@@ -611,6 +635,11 @@ func addMoreInfoToTasks(s *xorm.Session, taskMap map[int64]*Task, a web.Auth, vi
 		projectIDs = append(projectIDs, i.ProjectID)
 	}
 
+	err = addIsUnreadToTasks(s, taskIDs, taskMap, a)
```
kolaente: This should be behind an `?expand`

pkg/routes/api/v1/tasks.go
```diff
@@ -0,0 +1,55 @@
...
+	"net/http"
+	"strconv"
+
+	"code.vikunja.io/api/pkg/db"
+	"code.vikunja.io/api/pkg/models"
+	"code.vikunja.io/api/pkg/modules/auth"
+	"github.com/labstack/echo/v4"
+)
+
+func MarkTaskAsRead(c echo.Context) error {
```
kolaente: Please use the `CRUD` interface.

pkg/models/tasks.go
```diff
@@ -1855,3 +1890,9 @@ func triggerTaskUpdatedEventForTaskID(s *xorm.Session, auth web.Auth, taskID int
 	})
 	return err
 }
+
+func (t *Task) MarkTaskAsRead(s *xorm.Session, a web.Auth) error {
```
kolaente: This should be moved to the `task_unread_statuses.go` file.

mithileshgupta12:
@kolaente I did all the changes you mentioned.

frontend/src/components/tasks/partials/CommentCount.vue
```diff
@@ -41,6 +46,30 @@ const tooltip = computed(() => t('task.attributes.comment', props.task.commentCo
...
 	}
+
+	&.is-unread {
+		font-weight: 600;
+		color: var(--primary);
+
+		.unread-indicator {
+			display: inline-block;
+			inline-size: 6px;
+			block-size: 6px;
```
kolaente: Please use `rem` here.

frontend/src/components/tasks/partials/CommentCount.vue
```diff
@@ -41,6 +46,30 @@ const tooltip = computed(() => t('task.attributes.comment', props.task.commentCo
...
+			background-color: var(--primary);
+			border-radius: 50%;
+			margin-inline-start: 2px;
+			animation: pulse 2s infinite;
+		}
+	}
+}
+
+@keyframes pulse {
```
kolaente: Really unsure about the animation - I would prefer if it was not animated, because there isn't really much else in the UI that is animated by itself (transitions are a different thing).

frontend/src/services/task.ts
```diff
@@ -123,5 +124,15 @@ export default class TaskService extends AbstractService<ITask> {
 
 		return transformed as ITask
 	}
+
+	async markTaskAsRead(taskId: ITask['id']): Promise<void> {
+		const cancel = this.setLoading()
+	
+		try {
+			await AuthenticatedHTTPFactory().post(`/tasks/${taskId}/read`, {} as ITask)
+		}finally {
```
kolaente:
```suggestion
		} finally {
```

frontend/src/services/task.ts
```diff
@@ -123,5 +124,15 @@ export default class TaskService extends AbstractService<ITask> {
 
 		return transformed as ITask
 	}
+
+	async markTaskAsRead(taskId: ITask['id']): Promise<void> {
+		const cancel = this.setLoading()
+	
+		try {
+			await AuthenticatedHTTPFactory().post(`/tasks/${taskId}/read`, {} as ITask)
```
kolaente, line 132: something like `this.post` should work as well, but please check if that's the case.

pkg/models/listeners.go
```diff
@@ -1181,3 +1182,78 @@ func (s *HandleUserDataExport) Handle(msg *message.Message) (err error) {
...
 }
+
+type MarkTaskUnreadOnComment struct {
+}
+
+func (s *MarkTaskUnreadOnComment) Name() string {
+	return "task.comment.mark.unread"
+}
+
+func (s *MarkTaskUnreadOnComment) Handle(msg *message.Message) (err error) {
```
kolaente, line 1193: Can you add a test for this?

You might need to extract the logic into a new private function and then test that.

pkg/models/task_unread_statuses.go
```diff
@@ -0,0 +1,65 @@
...
+// @Accept json
+// @Produce json
+// @Security JWTKeyAuth
+// @Param projecttask path int true "Task ID"
+// @Success 200 {object} models.TaskUnreadStatus "The task unread status object."
+// @Failure 403 {object} web.HTTPError "The user does not have access to the task"
+// @Failure 500 {object} models.Message "Internal error"
+// @Router /tasks/{projecttask}/read [post]
+func (t *TaskUnreadStatus) Update(s *xorm.Session, a web.Auth) error {
+	err := markTaskAsRead(s, t.TaskID, a)
```
kolaente: you can shorten this to just `return markTaskAsRead(s, t.TaskID, a)`

frontend/src/services/task.ts
mithileshgupta12: `this.post` goes through the axios interceptor which calls `beforeUpdate()` → `processModel()`. Since we're passing an empty object `{}`, `processModel()` fails when trying to call `new Date(undefined).toISOString()` on the missing `created`/`updated` fields.

kolaente: gotcha, then it makes sense the way you did it.

mithileshgupta12:
@kolaente I did all the changes mentioned.

frontend/src/services/task.ts
```diff
@@ -7,6 +7,7 @@ import LabelService from './label'
 import {colorFromHex} from '@/helpers/color/colorFromHex'
 import {SECONDS_A_DAY, SECONDS_A_HOUR, SECONDS_A_WEEK} from '@/constants/date'
 import {objectToSnakeCase} from '@/helpers/case'
+import { AuthenticatedHTTPFactory } from '@/helpers/fetcher'
```
kolaente:
```suggestion
import {AuthenticatedHTTPFactory} from '@/helpers/fetcher'
```

just for consistency

PR #1845, opened 2025-11-19
fix(editor): preserve consecutive whitespace in comments in TipTap

maggch97:
fix: https://github.com/go-vikunja/vikunja/issues/1844
Ensure multiple spaces in comment content are no longer collapsed when editing/saving by:
- Adding SetContentOptions with parseOptions.preserveWhitespace = 'full'
- Applying those options to all setContent calls (initial load, exit edit mode, post-upload cleanup)
- Enabling preserveWhitespace in editor parseOptions

Previously, repeated spaces were normalized away after setContent(false), causing comments with deliberate spacing to be altered unexpectedly.

No behavioral changes beyond whitespace retention; renders identical except space fidelity.

Issue #1844, opened 2025-11-19
All comments containing user mention will be automatically edited.


maggch97:
### Description

<img width="666" height="485" alt="Image" src="https://github.com/user-attachments/assets/8b7c86c3-f577-4126-9d42-0641422ba74c" />

### Vikunja Version

latest

### Can you reproduce the bug on the Vikunja demo site?

Please select

maggch97:
<img width="1120" height="263" alt="Image" src="https://github.com/user-attachments/assets/d5c93ed0-64ec-475f-8b43-fef1565c1f01" />

<img width="1109" height="268" alt="Image" src="https://github.com/user-attachments/assets/8d72d25e-a427-421f-ae76-50171414d24c" />

PR #1893, opened 2025-11-27
fix(editor): preserve consecutive whitespace in comments in TipTap

kolaente:
Fixes https://github.com/go-vikunja/vikunja/issues/1844
Closes https://github.com/go-vikunja/vikunja/pull/1845

Ensure multiple spaces in comment content are no longer collapsed when editing/saving by:
- Adding SetContentOptions with parseOptions.preserveWhitespace = 'full'
- Applying those options to all setContent calls (initial load, exit edit mode, post-upload cleanup)
- Enabling preserveWhitespace in editor parseOptions

Previously, repeated spaces were normalized away after setContent(false), causing comments with deliberate spacing to be altered unexpectedly.

No behavioral changes beyond whitespace retention; renders identical except space fidelity.

### DIFF
--- a/frontend/tests/e2e/user/login.spec.ts
+++ b/frontend/tests/e2e/user/login.spec.ts
@@ -37,15 +37,14 @@ async function login(page: Page): Promise<void> {
 }
 
 test.describe('Login', () => {
-	test.beforeEach(async ({apiContext}) => {
+	test.beforeEach(async ({page, apiContext}) => {
 		await UserFactory.create(1, {username: credentials.username})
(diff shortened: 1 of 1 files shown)


kolaente:
Explanation of the fix by claude:

> The 'full' mode preserves all whitespace including between block elements, which caused the HTML parser to create extra empty list items from newlines between `<li>` tags. Using 'true' preserves whitespace in inline content only, which still achieves the goal of preserving consecutive spaces in text while not creating spurious nodes from formatting whitespace.

PR #1142, opened 2025-07-17
Filter background brightness to 60%

Anduin2017:
Before:

image

I bet it's a little too bright that can't focus on cards.

After:

image

Or simply allow user to edit it.

PR #1915, opened 2025-12-01
feat: Added background brightness setting

mithileshgupta12:
Closes https://github.com/go-vikunja/vikunja/pull/1142

### DIFF
--- a/frontend/src/components/home/ContentAuth.vue
+++ b/frontend/src/components/home/ContentAuth.vue
@@ -17,7 +17,7 @@
 				class="app-container-background background-fade-in d-print-none"
 				:style="{
 					'background-image': background && `url(${background})`,
-					'filter': backgroundBrightness !== null && `brightness(${backgroundBrightness}%)`
+					'filter': backgroundBrightness && `brightness(${backgroundBrightness}%)`
 				}"
 			/>
 			<Navigation class="d-print-none" />


frontend/src/components/home/ContentAuth.vue
```diff
@@ -15,7 +15,10 @@
 			<div
 				:class="{'is-visible': background}"
 				class="app-container-background background-fade-in d-print-none"
-				:style="{'background-image': background && `url(${background})`}"
+				:style="{
+					'background-image': background && `url(${background})`,
+					'filter': `brightness(${backgroundBrightness}%)`
```
kolaente: This should only be shown when a background brightness is set, similar to how it's done for the background image.

frontend/src/views/user/settings/General.vue
```diff
@@ -314,9 +314,26 @@
 					{{ $t('user.settings.general.allowIconChanges') }}
 				</label>
 			</div>
+			<div class="field">
+				<label class="two-col">
+					<span>
+						{{ $t('user.settings.backgroundBrightness.title') }}
+					</span>
+					<div class="select">
+						<select v-model.number="settings.frontendSettings.backgroundBrightness">
```
kolaente: Can this be a number input instead?

frontend/src/components/home/ContentAuth.vue
mithileshgupta12: In the auth store's `loadSettings` function,`backgroundBrightness` always has a default value of 100, so it's guaranteed to never be empty or undefined. Additionally, when the user clears the input, the value is reset to 100. This means the filter will always be applied with a valid value.

That said, if you prefer to only apply the filter when it differs from the default to avoid unnecessary CSS processing, I can do that as well.

kolaente: I would prefer if this was `null` if not set and then the css property would not be added.

mithileshgupta12:
@kolaente I did all the changes requested

frontend/src/components/home/ContentAuth.vue
```diff
@@ -15,7 +15,10 @@
 			<div
 				:class="{'is-visible': background}"
 				class="app-container-background background-fade-in d-print-none"
-				:style="{'background-image': background && `url(${background})`}"
+				:style="{
+					'background-image': background && `url(${background})`,
+					'filter': backgroundBrightness !== null && `brightness(${backgroundBrightness}%)`
```
kolaente: If this is undefined, it will not work (because `undefined !== null`)

Better:

```suggestion
					'filter': backgroundBrightness && `brightness(${backgroundBrightness}%)`
```

PR #1943, opened 2025-12-08
feat(openid): add emailclaim config for custom user identifier

RyanHecht:
Add `emailclaim` configuration option to openid auth providers to specify an alternative claim to use as the user's email address. This is useful for environments like Azure AD where the standard email claim may not be available or configurable per-app.

The emailclaim option can be set to use claims like 'upn' (User Principal Name) or any other claim containing a unique identifier in email format.

### DIFF
--- a/config-raw.json
+++ b/config-raw.json
@@ -750,6 +750,11 @@
                                             "comment": "The scope necessary to use oidc.\nIf you want to use the Feature to create and assign to Vikunja teams via oidc, you have to add the custom \"vikunja_scope\" and check [openid.md](https://vikunja.io/docs/openid/).\ne.g. scope: openid email profile vikunja_scope"
                                         },
                                         {
+                                            "key": "emailclaim",
+                                            "default_value": "",
+                                            "comment": "This option allows specifying an alternative claim to use as the user's email address. By default, Vikunja uses the standard `email` claim. In environments like Azure AD where users may not have email addresses configured or the email claim cannot be modified per-app, you can set this to use a different claim such as `upn` (User Principal Name) or any other claim that contains a unique identifier in email format. The claim value should be a valid email-like string that can be used to identify the user."
+                                        },
+                                        {
                                             "key": "usernamefallback",
                                             "default_value": "false",
                                             "comment": "This option allows to look for a local account where the OIDC Issuer match the Vikunja local username. Allowed value is either `true` or `false`. That option can be combined with `emailfallback`.\nUse with caution, this can allow the 3rd party provider to connect to *any* local account and therefore potential account hijaking."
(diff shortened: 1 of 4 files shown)


RyanHecht:
I run an organization that uses a free tier Azure AD/Entra ID tenant. Not everybody has the `email` claim filled in, since by default this is only done for those with Outlook licenses. 

Vikunja requires the `email` claim be provided, else users receive an error when they log in.

I cannot modify the `email` claim to point to the `upn` claim for the Vikunja application registration, because `email` is a "restricted claim" that cannot be modified by an app, as explained in [this Stack Exchange post](https://serverfault.com/questions/1106738/azure-ad-app-registration-possible-to-modify-or-transform-email-claim-provided).

So, I (with the assistance of Claude Opus 4.5 via [GitHub Copilot CLI](https://github.com/github/copilot-cli)) added a new option to the OpenID auth provider config: `emailclaim`. This allows

In my deployment, I've set `emailclaim: upn` to allow users without email addresses in Azure AD/Entra ID to still be able to log into Vikunja without getting an error.

I'm not married to the name `emailclaim`, I would accept any changes. 

I'll also note that I'm not a Go developer by day, and Copilot (via Opus) did most of the heavy lifting.

pkg/modules/auth/openid/openid_test.go
```diff
@@ -417,3 +417,139 @@ func TestMergeClaims(t *testing.T) {
 		assert.ErrorAs(t, err, &expectedErr)
 	})
 }
+
+// TestExtractCustomEmailClaim tests extracting email from a custom claim
+func TestExtractCustomEmailClaim(t *testing.T) {
```
kolaente, line 422: These tests just duplicate the logic in the test and are therefore not very useful since they don't test the actual logic. Please modify them accordingly.

config-raw.json
```diff
@@ -749,6 +749,11 @@
                                             "default_value": "openid email profile",
                                             "comment": "The scope necessary to use oidc.\nIf you want to use the Feature to create and assign to Vikunja teams via oidc, you have to add the custom \"vikunja_scope\" and check [openid.md](https://vikunja.io/docs/openid/).\ne.g. scope: openid email profile vikunja_scope"
                                         },
+                                        {
+                                            "key": "emailclaim",
+                                            "default_value": "",
+                                            "comment": "This option allows specifying an alternative claim to use as the user's email address. By default, Vikunja uses the standard `email` claim. In environments like Azure AD where users may not have email addresses configured or the email claim cannot be modified per-app, you can set this to use a different claim such as `upn` (User Principal Name) or any other claim that contains a unique identifier in email format. The claim value should be a valid email-like string that can be used to identify the user."
```
kolaente, line 755:
```suggestion
                                            "comment": "This option allows specifying an alternative claim to use as the user's email address if the default claim is empty or not present. By default, Vikunja uses the standard `email` claim. In environments like Azure AD where users may not have email addresses configured or the email claim cannot be modified per-app, you can set this to use a different claim such as `upn` (User Principal Name) or any other claim that contains a unique identifier in email format. The claim value should be a valid email-like string that can be used to identify the user."
```

kolaente:
Closing this as inactive, please ping if you want to pick it up again.

PR #1974, opened 2025-12-12
fix(menu): alignment of projects and subprojects in sidebar

vonalbert:
This PR changes the width of the collapse-project-button-placeholder element in order to match the width of the collapse button that has 0.25 + 0.5 rem of horizontal padding and the svg element inside has 1.25em of width.

As you can see from the screenshots, the first and third projects are misaligned compared to the second one. The same happens with sub-subprojects.

image
image

Here there are the same projects with this fix applied. The alignment seems to be preserved even with deeply nested projects.

image

image

### DIFF
--- a/frontend/src/styles/theme/navigation.scss
+++ b/frontend/src/styles/theme/navigation.scss
@@ -85,11 +85,12 @@
         svg {
           transition: all $transition;
           color: var(--grey-400);
+          width: 1.25rem;
         }
       }
 
       .collapse-project-button-placeholder {
-        width: calc(.75rem + 1.25em); 
+        width: 2rem;
         flex-shrink: 0;
       }
 


frontend/src/styles/theme/navigation.scss
```diff
@@ -89,7 +89,7 @@
       }
 
       .collapse-project-button-placeholder {
-        width: 1.625rem;
+        width: calc(.75rem + 1.25em);
```
kolaente: Good catch!

Does this work with `rem` only (no `calc`)?

vonalbert: Yes it works. There are 2 commits in the PR: in the first one I implemented the fix using rem only (it was 2rem), but while reviewing the fix I noticed the SVG element has a width of 1.25em, hence I changed the fix accordingly.

In this specific case, em and rem should result in the same value, so either approach would work. Another solution could be to change the SVG style to use rem for consistency with button padding and placeholder width. What do you think?

kolaente: I think the second approach is cleaner, so let's go with that.

vonalbert: I changed the units of the svg to rem, keeping the same value as before, and replaced the calc with a more readable 2rem. Then I realized that I may have misunderstood what you meant by the second approach.
If I’m mistaken, please let me know.

PR #1988, opened 2025-12-15
fix: ensure API consistency for /tasks and empty array responses

### LINKED ISSUE #1984: Inconsistent APIs: tasks/all vs projects & labels returns null
### Description

Bottom-line: Highlighting the "get-all" type API calls aren't the same URL (`/tasks/all` vs `/projects`) and the empty response is `[]` for some and `null` for others.

Caveat: This is either beyond fix with 1.x.x API on RC3, or there's a bigger design choice I've not seen. Or maybe this is a feature request rather than a bug.

First, the "get all the tasks a user has access to" API utilizes `all` at the end (`https://try.vikunja.io/api/v1/tasks/all`) while other "get everything" APIs do not, such as projects (`https://try.vikunja.io/api/v1/projects`). This seems like an inconsistency.

Second, while working with the APIs, I noticed that requesting tasks using pagination parameters that are beyond the total number of tasks returns an empty array (`[]`). For example, 55 tasks exist and page 2 with 100 tasks per page is requested: empty array is returned. However, not so for the labels API.

For the labels API, if 55 labels exist and page 2 is requested with 100 labels per page is requested the API returns `null`. This is inconsistent with what I would expect based on the tasks API. The `teams` API also returns an empty array for this scenario.

For both items (`/tasks/all` and `/labels` --> `null`), the <https://try.vikunja.io/api/v1/docs> documentation doesn't specify the different forms of the results when there is nothing to be returned by the paginated results.

Finally, wonderful project, use this thing all the time. If all of this is beyond API freeze or whatever, close it out and move on. Thanks!

### Vikunja Version

v1.0.0-rc3

### Browser and version

_No response_

Please select

### Screenshots

_No response_

kolaente:
## Summary

- Renames the `/tasks/all` endpoint to `/tasks` for consistency with other collection endpoints like `/projects` and `/labels`
- Returns `[]` instead of `null` for empty pagination results across all list endpoints
- Updates the frontend service to use the new endpoint path
- Updates API token tests to use the new endpoint path

The empty array fix is implemented in the generic `ReadAllWeb` handler using reflection to detect nil slices wrapped in `interface{}` (a Go gotcha where an interface containing a nil slice is not equal to `nil` itself).

Fixes #1984

## Test plan

- [x] Verify `/api/v1/tasks` returns all tasks the user has access to
- [x] Verify empty results return `[]` not `null` for all endpoints:
  - `/tasks`, `/labels`, `/teams`, `/notifications`, `/tokens`
  - `/projects/:id/shares`, `/projects/:id/tasks`, `/projects/:id/teams`, `/projects/:id/users`, `/projects/:id/webhooks`
- [x] Run existing tests to ensure no regressions

### DIFF
--- a/pkg/web/handler/read_all.go
+++ b/pkg/web/handler/read_all.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"math"
 	"net/http"
+	"reflect"
 	"strconv"
 
 	vconfig "code.vikunja.io/api/pkg/config"
@@ -126,6 +127,13 @@ func (c *WebHandler) ReadAllWeb(ctx echo.Context) error {
 	err = s.Commit()
 	if err != nil {
 		return HandleHTTPError(err)
+	}
+
+	// Ensure we return an empty array instead of null when there are no results.
+	// We need to use reflection here because a nil slice wrapped in an interface{}
+	// is not equal to nil (the interface contains a nil value but is not nil itself).
+	if result == nil || (reflect.ValueOf(result).Kind() == reflect.Slice && reflect.ValueOf(result).IsNil()) {
+		result = []interface{}{}
 	}
 
 	err = ctx.JSON(http.StatusOK, result)


shamoon:
Totally I realize it's a release candidate and it's "not your problem", but this is a complete breaking change, it would be lovely to surface these kinda things in the release notes. Cheers

https://github.com/gethomepage/homepage/discussions/6232

PR #2027, opened 2025-12-27
Change Done status condition from '1' to '2'

red-avtovo:
Hi! Thanks for the application.
I just installed the Vikunja 1.0.0-rc3 and tried to import my data from TickTick.
I noticed that completed tasks are set to 2 in the backup CSV file and I see, that you're mapping it to 1 instead. It seems like  TickTick changed its status mapping.

As you can see at the screenshot, there are no `status=1` at all. The file itself is about 700 rows, and statuses have only 0s or 2s
image

### DIFF
--- a/pkg/modules/migration/ticktick/ticktick_test.go
+++ b/pkg/modules/migration/ticktick/ticktick_test.go
@@ -86,7 +86,7 @@ func TestConvertTicktickTasksToVikunja(t *testing.T) {
 			ParentID:    0,
 			ProjectName: "Project 2",
 			Title:       "Test task 4",
-			Status:      "0",
+			Status:      "2",
 			Order:       -109951627777,
 		},
 	}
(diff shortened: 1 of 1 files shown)


kolaente:
In the csv export, what does the status explanation say at the top of the file?

red-avtovo:
ooohhh... I can't read

"Status: 
0 Normal
1 Completed
2 Archived"

Although, the tool treats 2 as not completed

kolaente:
Technically, just because a task is archived does not mean it's completed. I guess the semantics of it are similar, but it's not the same.

Since Vikunja does not have a way to properly archive tasks, we could handle them the same. Do you want to adjust your PR?

kolaente:
Please also adjust the tests

kolaente:
Hey @red-avtovo are you still interested in this?

red-avtovo:
Sorry, no, not anymore 😢

PR #2031, opened 2025-12-28
fix(api): preserve IsFavorite for saved filters in ReadOne

### LINKED ISSUE #1989: Saved filters render twice in sidebar on initial load
### Bug:
Saved filters appear in both `savedFilterProjects` section AND main projects list on page load. Clicking any duplicate removes it. Reappears on refresh.

### Console on load:
[debug] Loading project, projectId: -4, loadedProjectId = 0, currentProject = undefined

### API data is correct:
```json
{"id": -3, "title": "@WORK", "parent_project_id": 0}
{"id": -4, "title": "@IN-PROGRESS", "parent_project_id": 0}
{"id": -5, "title": "@PERSONAL", "parent_project_id": 0}
```

### Hypothesis: 
Race condition in Navigation.vue - Sidebar renders before projectsArray populated. Computed properties (notArchivedRootProjects filtering p.id > 0) return stale data initially. Possibly triggered by data volume (22 projects, 298 tasks, 3 saved filters).

### Vikunja Version

v1.0.0-rc3

### Browser and version

Brave 143.1.85.111 && Firefox 146.0

No

### Initial Load:

Image 

### After selecting each filter once:

Image

IAMSamuelRodda:
## Summary
- Saved filters' `IsFavorite` field was not being properly returned when fetched as pseudo-projects via `/projects/{id}`
- This caused favorited filters to appear in both the Favorites and Filters sections initially, but then disappear from Favorites after clicking on them (navigating to the filter)

## Root Cause
Two issues in `pkg/models/project.go` `ReadOne()`:
1. `IsFavorite` wasn't being copied from the `SavedFilter` struct to the `Project` struct
2. The `isFavorite()` lookup was then overwriting the value with `false` (since saved filters don't use the regular project favorites table)

## Fix
- Copy `sf.IsFavorite` to `p.IsFavorite` when loading a saved filter as a project
- Skip the `isFavorite()` lookup for saved filters since they manage their favorite status independently via the `saved_filters` table

## Test plan
- [x] Create/mark a saved filter as favorite (via API since no UI exists)
- [x] Refresh page - filter appears in both Favorites and Filters sections (intended)
- [x] Click on the filter to navigate to it
- [x] Verify filter still appears in Favorites section (before fix: it disappeared)
- [x] Verify `/api/v1/projects/-{id}` returns `is_favorite: true` matching `/api/v1/filters/{id}`

Fixes #1989

### DIFF
--- a/pkg/models/project.go
+++ b/pkg/models/project.go
@@ -295,6 +295,7 @@ func (p *Project) ReadOne(s *xorm.Session, a web.Auth) (err error) {
 		}
 		p.Title = sf.Title
 		p.Description = sf.Description
+		p.IsFavorite = sf.IsFavorite
 		p.Created = sf.Created
 		p.Updated = sf.Updated
 		p.OwnerID = sf.OwnerID
@@ -334,9 +335,13 @@ func (p *Project) ReadOne(s *xorm.Session, a web.Auth) (err error) {
 		}
 	}
 
-	p.IsFavorite, err = isFavorite(s, p.ID, a, FavoriteKindProject)
-	if err != nil {
-		return
+	// For saved filters, IsFavorite was already set from the SavedFilter struct.
+	// Don't overwrite it with the project favorites lookup.
+	if !isFilter {
+		p.IsFavorite, err = isFavorite(s, p.ID, a, FavoriteKindProject)
+		if err != nil {
+			return
+		}
 	}
 
 	subs, err := GetSubscriptionForUser(s, SubscriptionEntityProject, p.ID, a)


kolaente:
Still the question remains: there is no way a saved filter can be marked as favorite. How did you manage to do that leading to it showing up multiple times?

IAMSamuelRodda:
is_favorite can be set via the api for a filter. i've an mcp server for vikunja and was migrating from todoist attempting to create my setup from there (they have a gui option to favorite filters). I found it more efficient to setup labels and filters via the mcp server. I never questioned that there was no GUI option, though I have changed that for future with an enhanced sidebar i'm working on. Not sure if its relavent for the main build though. Calling it "enhanced sidebar mode" but really it is just suppose to make me feel better and a little more like todoist with the layout, task counters and favorite options in the dot menus.
image 
https://github.com/IAMSamuelRodda/vikunja/tree/feat/sidebar-enhanced-mode

kolaente:
We should expose the option to mark filters as favorites in the UI though

PR #2053, opened 2026-01-05
feat(files): validate file storage is writable on startup

kolaente:
## Summary
- Adds startup validation that checks if the configured file storage is writable
- Creates and removes a temporary `.vikunja-check-{timestamp}` file during `InitFileHandler()`
- Fails startup with a clear error if storage is not writable

## Changes
- Extract `writeToStorage()` internal function from `Save()` for reuse
- Add `ValidateFileStorage()` function to test storage writability
- Call validation after backend initialization in `InitFileHandler()`

## Test plan
- [ ] Start Vikunja with valid local storage path - should start normally
- [ ] Start Vikunja with read-only local storage path - should fail with clear error
- [ ] Start Vikunja with valid S3 configuration - should start normally
- [ ] Start Vikunja with invalid S3 credentials - should fail with clear error

### DIFF
--- a/pkg/files/files.go
+++ b/pkg/files/files.go
@@ -158,7 +158,6 @@ func writeToStorage(path string, content io.Reader, size uint64) error {
 		return afs.WriteReader(path, content)
 	}
 
-	// S3 path - use PutObject with Content-Length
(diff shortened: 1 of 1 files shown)


kolaente:
auggie review

kolaente:
@codex review

PR #971, opened 2025-06-17
feat: merge duplicat notifications

dpschen:

PR #2056, opened 2026-01-06
feat: merge duplicate notifications

kolaente:
closes https://github.com/go-vikunja/vikunja/pull/971

## Summary
Incorporates the changes from https://github.com/go-vikunja/vikunja/pull/971 with improvements:

- Enables duplicate notification merging using the `ignoreDuplicates` prop from vue3-notification
- Shows a count (×N) when notifications are merged
- **Fix**: Uses string text instead of arrays so duplicate detection actually works (vue3-notification uses strict equality)
- **Fix**: Moves `actions` under `data` in error notifications to match template's `item.data?.actions`
- **Fix**: Updates template to handle both string and array text for backwards compatibility
- Replaces Cypress test with Playwright e2e test

## Test plan
- [x] Playwright test verifies duplicate notifications are merged and show correct count

### DIFF
--- a/frontend/src/components/misc/Notification.vue
+++ b/frontend/src/components/misc/Notification.vue
@@ -21,11 +21,16 @@
 					{{ item.title }}
 				</div>
 				<div class="notification-content">
-					<template
-						v-for="(t, k) in item.text"
-						:key="k"
-					>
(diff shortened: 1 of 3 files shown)


kolaente:
auggie review

PR #2083, opened 2026-01-11
feat: add configurable gravatar-compatible base URL

### LINKED ISSUE #2082: Implement Federated Avatars
### Description

It would be useful to implement the ability to use federated avatars ( https://libravatar.org/ , a libre Gravatar-like service ), similar to the way Gitea/Forgejo does it. They use this library: https://github.com/gogs/go-libravatar .

Once the library is imported, you can use it like so:

```golang
import "strk.kbt.io/projects/go/libravatar"
avt := libravatar.New()
avatar_url := avt.FromEmail("person@example.com")
```

My use case is I run my own libravatar server for the services that run inside my SSO network. It's working with Gitea & several other services I run. Other people may use the public libravatar service.

As an alternative, since libravatar is compatible with Gravatar-format HTTP requests, you can enable a configurable gravatar/libravatar URL. As a bonus, libravatar can fallback to Gravatar itself.

For instance, currntly you have www.gravatar.com hardcoded

```golang
req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "https://www.gravatar.com/avatar/"+utils.Md5String(strings.ToLower(user.Email))+"?s="+sizeString+"&d=mp", nil)
```

But you could have it as a variable, so the user could configure the grvatar/libravatar URL to be any compatible gravatar/libravatar service, with default of www.gravatar.com.

https://www.gravatar.com

https://libravatar.org

https://avatars.example.com

```yaml
avatar:
  gravatarexpiration: 3600
  # Possible new setting.
  gravatarbaseurl: "https://avatars.example.com"
```

rhclayto:
This adds the ability to set a base URL for a Gravatar-compatible avatar service (Gravatar itself, or Libravatar, for instance). The default will be www.gravatar.com, so nothing will change from current behavios unless the user explicitly configures another URL.

Resolve #2082

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -187,7 +187,7 @@ const (
 	CorsMaxAge  Key = `cors.maxage`
 
 	AvatarGravaterExpiration Key = `avatar.gravatarexpiration`
-	AvatarGravatarBaseURL Key = `avatar.gravatarbaseurl`
+	AvatarGravatarBaseURL    Key = `avatar.gravatarbaseurl`
 
 	BackgroundsEnabled               Key = `backgrounds.enabled`
 	BackgroundsUploadEnabled         Key = `backgrounds.providers.upload.enabled`


config-raw.json
```diff
@@ -616,6 +616,11 @@
                     "key": "gravatarexpiration",
                     "default_value": "3600",
                     "comment": "When using gravatar, this is the duration in seconds until a cached gravatar user avatar expires"
+                },
+                {
+                    "key": "gravatarbaseurl",
+                    "default_value": "https://www.gravatar.com",
+                    "comment": "If you use a Gravatar-compatible service other than gravatar.com, you may configure the base URL for the service here.\nFor instance, gravatarbaseurl: 'https://libravatar.org'. The default is https://www.gravatar.com.\nDo not include a trailing slash."
```
kolaente: Can you add a check so that the slash is automatically removed in the config init if present?

kolaente (changes requested):

config-raw.json
rhclayto: Yes, I will, but that will definitely be tomorrow. Thank you for looking at this.

kolaente: No pressure, take your time!

kolaente: No pressure, take your time!

rhclayto: OK, I did it.

pkg/config/config.go
```diff
@@ -605,6 +607,8 @@ func InitConfig() {
 
 	readConfigValuesFromFiles()
 
+	AvatarGravatarBaseURL.Set(strings.TrimRight(AvatarGravatarBaseURL.GetString(), "/"))
```
kolaente, line 614: Please also validate the url - should be enough to pass it through `url.ParseRequestURI` and check if that returns an error.

rhclayto: OK. I see a couple lines down it has `parsed, err := url.Parse(ServicePublicURL.GetString())`. Should I use `url.Parse` to be consistent with that or is `url.ParseRequestURI` better?

Edit: I just used `url.ParseRequestURI` as you suggested.

kolaente (approved):
Thanks!

kolaente:
Please fix the reported lint issues

rhclayto:
> Please fix the reported lint issues

Done.

kolaente:
Thanks again!

PR #1889, opened 2025-11-27
feat: allow users to customize keyboard shortcuts

kolaente:
## Summary by CodeRabbit

* **New Features**
  * Added customizable keyboard shortcuts management system.
  * New keyboard shortcuts settings page with intuitive editor UI.
  * Ability to customize individual shortcuts, reset to defaults, or reset all at once.
  * Real-time conflict detection to prevent duplicate shortcuts.
  * Shortcut changes persist across sessions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2116, opened 2026-01-18
feat: non-latin keyboard layout compatible shortcuts

seorgiy:
### Problem
The keyboard shortcuts requires using a Latin-based keyboard layout.
For example, `Ctrl+k` on the Russian layout would result in `Ctrl+л`.

### Assumptions
- We want to stay with [github/hotkeys](https://github.com/github/hotkey) library
- We don't want to use `event.code` instead of `event.key`, as it discussed [here](https://github.com/github/hotkey/issues/34)
- We don't want to mess with the `KeyboardEvent` due to possible browser-specific troubles and the security reasons
- Hotkeys should work simultaneously in a Latin and non-Latin variant

### Solution
Duplicate all the hotkeys for language-specific variants based on provided language key maps such as `{ k: 'л' }` for the Russian example. This PR contains the Russian hotkey map and the functionality to add more.

### Concerns
- Still not sure we shouldn't switch for `event.code` approach
- If we add a lot of key maps, it may lead to some conflicts, so we probably should make them user-controllable like an option «add keyboard support for: [multiple select from allowed keyboards]». But it doesn't make sense until we agree on the approach
- In general this PR conflicts with [custom shortcuts](https://github.com/go-vikunja/vikunja/pull/1889) but it's mergeable


kolaente:
Hey thanks for the PR! I'll do a proper review later, as the 1.0 release is right around the corner.

> If we add a lot of key maps, it may lead to some conflicts, so we probably should make them user-controllable like an option «add keyboard support for: [multiple select from allowed keyboards]». But it doesn't make sense until we agree on the approach

Can't we auto-detect the keyboard that's used and switch based on that? Another option would be to switch based on the selected language, but that might not be the best indicator.

seorgiy:
It's worth mentioning that I'm not a front-end guy at the first place, so please don't full trust me on this. 

But the more I researched the issue, the more convinced I became that we should consider switching to ```event.code``` specifically in the Vikunja case, and then the hotkey reassignment would also start makes sense. For now even if I set myself Ctrl+л instead of Ctrl+k, then the English layout will not work for me. While I constantly change the language back and forth, like any other it guys, in order to describe the task in my native language with code inclusions, for example. So I expect hotkeys to stay persistent, as in some game, not in the way like «**A** goes for **Add Task**, so in Russian it should be **Д** for **Добавить задачу**, which is direct "Add Task" translation». On that topic I can refer to this [article](https://medium.com/@AlexanderObregon/handling-keyboard-input-with-javascript-keyboardevent-properties-8f558597a853).

> Can't we auto-detect the keyboard that's used and switch based on that? 

There is banch of security reasons why you can't get full info about the user's keyboard before they type anything, so at best case you can create something like pop-up which goes like «We see you typing in Chinese, do you want to enable Chinese-hotkeys support?». And this UX sucks.

Not sure I can promise you a PR with full conversion to `event.code` approach, but for now we wait for 1.0 release anyway. So please do a quick research on the topic in the mean time and mention me at the moment you will be ready to considering conversion, may be I will try to offer at least some help, dependce on my work load at the moment🫠

kolaente:
Sorry for the late reply!

I've checked your approach and it seems like for the use case that we have your idea is the better one. I'll check what we would need to implement this with `event.code` instead.

kolaente:
Here's a draft PR that implements this: https://github.com/go-vikunja/vikunja/pull/2310

PR #2122, opened 2026-01-19
feat: add filtering and sorting by task bucket name

### LINKED ISSUE #2101: Filter on task bucket name
### Description

Hi,

Thanks for the great app!

As far as I can see on the Filter Syntax page, it is not possible to filter tasks on the name of their bucket.

A use case is to exclude from lists tasks that are started but not actively worked on, for example by moving them in a "On Hold" bucket.

A bucket name filter would support the string operators (`=`, `!=`, `like`, `in`, `not in` ...).

KR,

Bernard

_No response_

mithileshgupta12:
Resolves https://github.com/go-vikunja/vikunja/issues/2101


pkg/models/task_collection_filter.go
```diff
@@ -350,6 +350,18 @@ func getNativeValueForTaskField(fieldName string, comparator taskFilterComparato
 		return nil, valueSlice, nil
 	}
 
+	if realFieldName == "Bucket" {
+		if comparator == taskFilterComparatorIn || comparator == taskFilterComparatorNotIn {
```
kolaente: Why the check here? How is it different from the assignees filter above?

pkg/models/task_search.go
```diff
@@ -80,6 +80,12 @@ var subTableFilters = SubTableFilters{
 		FilterableField: "parent_project_id",
 		AllowNullCheck:  false,
 	},
+	"bucket": {
+		Table:           "buckets",
+		BaseFilter:      "task_buckets.bucket_id = buckets.id",
+		FilterableField: "title",
```
kolaente, line 86: TL;DR: leave to filter by title, but we might want to add filtering by id later.

I'm unsure if this should be id instead (just rambling, let me know what you think).

Filtering by id would be more consistent with other fields (except assignees, which has a unique username instead). 

On the other hand, filtering by title would be more in line with how people use it and it would not require a translation between title and id. But it may result in duplicates when the same name is used in multiple different views (though that could be solved by filtering on the project or view).

Maybe let's leave it at what it is right now, we can always add filtering by id later.

kolaente (changes requested):
Thanks for the PR!

Please add tests for the filtering and make sure the highlighting works in the frontend filter input.

pkg/models/task_collection_filter.go
mithileshgupta12: Good catch! You're right - there's no need for the comparator check. I've simplified it to match the assignees filter exactly. Both now just split the value on commas and return a string slice. The comparator handling happens later in the filter processing code.

pkg/models/task_search.go
mithileshgupta12: I agree with keeping it as title filtering for now. Here's my thinking:

1. **User-friendliness**: Filtering by name (`bucket = "Doing"`) is much more intuitive than remembering bucket IDs, especially since bucket names are what users see in the UI.

2. **Consistency with assignees**: As you mentioned, this follows the assignees pattern which also filters by a human-readable field (username) rather than ID.

3. **The duplicate name issue is manageable**: If someone uses the same bucket name across different views and needs to disambiguate, they can combine it with a view or project filter (e.g., `bucket = "On Hold" && project = 123`).

4. **Easy to extend later**: We can always add `bucket_id` filtering later for advanced use cases, similar to how we have both `labels` and `label_id`.

The use case from the original issue (filtering out "On Hold" tasks) works perfectly with name-based filtering and doesn't require knowing any IDs.

mithileshgupta12:
@kolaente All issues are fixed. Please check

kolaente (commented):
Looks pretty good already, can you also add tests for the actual filtering? Not just the expression parsing.

mithileshgupta12:
Hi @kolaente I am unable to figure out why the test issue is happening. Can you please check?

kolaente:
This test is failing:

```
    --- FAIL: TestTaskCollection_ReadAll/filter_bucket_like_pattern (0.01s)
```

you should be able to run them locally with `mage test:feature | grep FAIL`

hajiboy95:
any chance of this being picked up? would be great feature! @mithileshgupta12

PR #2178, opened 2026-01-29
feat: use yaegi for plugins instead of go-native

kolaente:
## Summary

Proof-of-concept for a Yaegi-based plugin system as an alternative to Go's native `plugin` package (`.so` shared libraries).

### Why Yaegi instead of Go's native plugins?

Go's native plugin system (`-buildmode=plugin`) has significant limitations:
- Plugins must be compiled with the **exact same Go version** and **identical dependency versions** as the host binary
- Only works on Linux and macOS (no Windows support)
- No way to unload plugins once loaded
- Build toolchain constraints make distribution painful

[Yaegi](https://github.com/traefik/yaegi) is a Go interpreter that loads `.go` source files at runtime, avoiding all of the above. The trade-off is slower execution (interpreted, not compiled), but for plugin hooks this is generally acceptable.

### Key difference: typed factory functions

With native Go plugins, you export a single `NewPlugin() plugins.Plugin` factory and then type-assert the result to discover additional capabilities:

```go
// Native plugin: single factory, type assertions work
func NewPlugin() plugins.Plugin { return &MyPlugin{} }

// Host code can do:
// p := NewPlugin()
// if rp, ok := p.(plugins.AuthenticatedRouterPlugin); ok { ... }
```

**Yaegi wraps interpreted values per their declared return type**, which breaks sub-interface type assertions. So with Yaegi, plugins must export **separate typed factory functions** for each capability:

```go
// Yaegi plugin: one factory per interface
func NewPlugin() plugins.Plugin                                       { return &MyPlugin{} }
func NewAuthenticatedRouterPlugin() plugins.AuthenticatedRouterPlugin { return &MyPlugin{} }
func NewUnauthenticatedRouterPlugin() plugins.UnauthenticatedRouterPlugin { return &MyPlugin{} }
```

Only `NewPlugin()` is required. The others are optional and discovered by the loader.

### Plugin capabilities

| Capability | Interface | Description |
|---|---|---|
| Lifecycle | `Plugin` | `Init()` / `Shutdown()` hooks |
| Authenticated routes | `AuthenticatedRouterPlugin` | Routes under `/api/v1/plugins/*` (behind auth) |
| Unauthenticated routes | `UnauthenticatedRouterPlugin` | Routes under `/plugins/*` (no auth, rate-limited) |
| Database migrations | `MigrationPlugin` | xormigrate migrations appended to core |
| Event listeners | Via `events.RegisterListener()` in `Init()` | React to domain events (e.g. `TaskCreatedEvent`) |

Plugins also have access to Vikunja internals: `db.NewSession()`, `user.GetCurrentUserFromDB()`, `log.Infof()`, all model types, etc.

### Writing a plugin

See `examples/plugins/example/main.go` for a complete example. The key points:

1. Package must be `main`
2. Define a struct implementing `plugins.Plugin` (required) plus any optional interfaces
3. Export typed factory functions (see above)
4. In `Init()`, register event listeners via `events.RegisterListener()`
5. In route registration methods, add Echo handlers to the provided group

### Configuration

```yaml
plugins:
  enabled: true
  dir: ./plugins    # directory containing plugin source folders or .so files
  loader: native    # "native" (default) for .so files, "yaegi" for Go source dirs
```

The `plugins.loader` value is validated at startup — invalid values cause a fatal error.

- **`native`** (default): scans `plugins.dir` for `.so` shared library files. Plugins must be compiled with the exact same Go version and dependencies as the Vikunja binary.
- **`yaegi`**: scans `plugins.dir` for subdirectories containing `.go` source files. Plugins are interpreted at runtime — no compilation step needed.

### Known limitations

- **Single-file plugins only (for now):** The yaegi loader evaluates `.go` files individually via `Eval()` rather than using yaegi's `EvalPath()`. This means multi-file plugins may hit order-dependency issues if cross-file declarations depend on filename ordering. `EvalPath()` would solve this, but it doesn't support `main.NewPlugin` symbol lookup for `package main` — yaegi resolves the package name from the import path rather than the `package` declaration. A future fix could either use a non-main package convention or find a workaround for `EvalPath` symbol resolution.
- **Typed factory functions required:** Due to yaegi's interface wrapping, plugins cannot use a single factory with type assertions. Each capability needs its own exported factory function (see above).
- **Interpreted performance:** Yaegi-loaded plugins run interpreted, not compiled. This is fine for hooks and route handlers but would be a concern for hot-path code.

### Files

- `pkg/plugins/interfaces.go` — plugin interface definitions
- `pkg/plugins/manager.go` — plugin manager, loads plugins via configured loader
- `pkg/plugins/yaegi/loader.go` — Yaegi-based interpreter loader
- `pkg/yaegi_symbols/` — generated symbol tables exposing Vikunja internals to interpreted code
- `examples/plugins/example/main.go` — example plugin

### DIFF
--- a/pkg/plugins/yaegi/events_test.go
+++ b/pkg/plugins/yaegi/events_test.go
@@ -1,0 +1,48 @@
+// Vikunja is a to-do list application to facilitate your life.
+// Copyright 2018-present Vikunja and contributors. All rights reserved.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package yaegi
+
+import (
+	"testing"
+
+	"code.vikunja.io/api/pkg/log"
+)
+
+func TestPluginEventListener(t *testing.T) {
+	log.InitLogger()
+
+	loaded, err := LoadPluginFull(examplePluginDir)
+	if err != nil {
+		t.Fatalf("LoadPluginFull failed: %v", err)
+	}
+
+	// Call Init() — this registers the TestListener for TaskCreatedEvent.
+	// If the Listener interface boundary is broken, this will panic with a
+	// reflection error when calling events.RegisterListener.
+	err = loaded.Plugin.Init()
+	if err != nil {
+		t.Fatalf("plugin Init failed: %v", err)
+	}
+	t.Log("Init() succeeded — events.RegisterListener accepted the interpreted Listener")
+
+	// Verify Shutdown works too
+	err = loaded.Plugin.Shutdown()
+	if err != nil {
+		t.Fatalf("plugin Shutdown failed: %v", err)
+	}
+	t.Log("Shutdown() succeeded")
+}
--- a/pkg/plugins/yaegi/loader_test.go
+++ b/pkg/plugins/yaegi/loader_test.go
@@ -1,0 +1,70 @@
+// Vikunja is a to-do list application to facilitate your life.
+// Copyright 2018-present Vikunja and contributors. All rights reserved.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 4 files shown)


pkg/plugins/yaegi/events_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+package yaegi
+
+import (
+	"testing"
+
+	"code.vikunja.io/api/pkg/log"
+)
+
+func TestPluginEventListener(t *testing.T) {
```
kolaente, line 25: Is this a smoke test?

pkg/config/config.go
```diff
@@ -483,6 +484,7 @@ func InitDefaultConfig() {
 	// Plugins
 	PluginsEnabled.setDefault(false)
 	PluginsDir.setDefault(ResolvePath("plugins"))
+	PluginsLoader.setDefault("yaegi")
```
kolaente: should default to native

pkg/plugins/manager.go
```diff
@@ -99,19 +111,31 @@ func (m *Manager) loadPlugins(paths []string) error {
...
+				}
+			case "yaegi":
+				if !e.IsDir() {
+					continue
+				}
+				if err := m.loadYaegiPlugin(full); err != nil {
+					log.Errorf("Failed to load yaegi plugin %s: %s", e.Name(), err)
+				}
+			default:
+				return errors.New("invalid plugins.loader value: must be \"yaegi\" or \"native\"")
```
kolaente: This should be checked in the config loading

kolaente:
auggie review

PR #2186, opened 2026-02-01
fix(files): require io.ReadSeeker for S3 uploads to prevent permission denied errors

### LINKED ISSUE #2185: Vikunja 1.0.0 S3 Avatar Upload fails on permission denied
### Pre-submission checklist

### Description

running Vikunja 1.0.0 in a docker container, using Wasabi S3

when using the S3 file type attachments are made perfectly fine, however avatars get hit with a permission denied on a localpath `./vikunja-s3-upload-stringofnumbers`, I myself am not sure where it's trying to dump this file that it wouldn't have permission over, as the files path should be perfectly writable, and the rest is owned by the docker container it runs in.  
when files is set to local instead of S3 avatar upload works perfectly fine.

most of my config is defaults except for openID being set, and filetype being S3.  

### Vikunja Version

1.0.0

### Browser and version

_No response_

No

### Screenshots

_No response_

kolaente:
When uploading avatars (and background images), the resized image was passed as a `*bytes.Buffer` to the file storage layer. Since `*bytes.Buffer` does not implement `io.ReadSeeker`, the S3 upload code fell back to creating temporary files (`vikunja-s3-upload-*`) on the local filesystem. In Docker containers with restrictive permissions, this temp file creation failed. Regular file attachments worked because `multipart.File` already implements `io.ReadSeeker`.

**Fix:** Changed the file storage API (`Create`, `CreateWithMime`, `CreateWithMimeAndSession`, `Save`) to require `io.ReadSeeker` instead of `io.Reader`. This:
- Eliminates the temp file fallback entirely — no more local filesystem writes during S3 uploads
- Enforces at the **type level** that all callers provide seekable readers, preventing this class of bug from recurring
- Updated all callers (avatar upload, background upload, unsplash, dump restore, migration) to provide `*bytes.Reader` or equivalent seekable readers

Fixes #2185

### DIFF
--- a/pkg/files/s3_test.go
+++ b/pkg/files/s3_test.go
@@ -367,7 +367,7 @@ func TestFileSave_S3_ReturnsErrorOnPutObjectFailure(t *testing.T) {
 	assert.Contains(t, err.Error(), "failed to upload file to S3")
 }
 
-func TestFileSave_S3_LogsWarnOnSizeMismatch(t *testing.T) {
+func TestFileSave_S3_UsesActualReaderSizeOnMismatch(t *testing.T) {
 	originalClient := s3Client
 	originalBucket := s3Bucket
 	t.Cleanup(func() {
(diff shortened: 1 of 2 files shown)


kolaente:
auggie review

kolaente:
@codex review

kolaente:
auggie review

PR #2193, opened 2026-02-04
refactor(db): extract testable ResolveDatabasePath function

### LINKED ISSUE #2189: SQLite absolute database path ignored in 1.0.0, only relative path works
### Pre-submission checklist

### Description

When configuring an absolute path for the SQLite database in config.yml, Vikunja 1.0.0 ignores it and uses the default path instead. Only relative paths work correctly.
Environment:

Vikunja version: 1.0.0
Database: SQLite
Installation method: .deb package on Debian Trixie

Configuration (does NOT work):

```yaml
database:
  type: sqlite
  path: "/var/lib/vikunja/vikunja.db"  # Absolute path
```

**Result:**
```
level=INFO msg="Using SQLite database at: /opt/vikunja/vikunja.db"
```

Config is read (confirmed via logs), but the absolute path is ignored.
Workaround (works):

```yaml
service:
  rootpath: "/var/lib/vikunja/"
database:
  type: sqlite
  path: "vikunja.db"  # Relative path
```

**Result:**
```
level=INFO msg="Using SQLite database at: /var/lib/vikunja/vikunja.db"
```

Expected behavior:
According to [the documentation](https://vikunja.io/docs/config-options/#1-database-path), absolute paths should work: "Can be an absolute path or relative path."

### Vikunja Version

1.0.0

### Browser and version

_No response_

No

### Screenshots

_No response_

kolaente:
Add DatabasePathConfig struct and ResolveDatabasePath function that takes all dependencies as parameters, making it easier to test path resolution logic in isolation. Should also fix the reported cases.

Resolves #2189

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -220,7 +220,7 @@ func resolveDatabasePath(cfg DatabasePathConfig, getUserDataDir func() (string, 
 	default:
 		dataDir, err := getUserDataDir()
 		if err != nil {
-			log.Debugf("Could not get user data directory, falling back to rootpath: %v", err)
+			log.Warningf("Could not get user data directory, falling back to rootpath: %v", err)
 			path = filepath.Join(cfg.RootPath, cfg.ConfiguredPath)
 		} else {
 			path = filepath.Join(dataDir, cfg.ConfiguredPath)


kolaente:
auggie review

PR #2201, opened 2026-02-06
fix: prevent login form flash in authenticated app shell

kolaente:
## Summary

- After a successful login or registration, the login form briefly flashed (~250ms) inside the authenticated app shell (navbar + sidebar) before navigating to the home page
- Root cause: `App.vue` switches to the authenticated layout via `v-if="authStore.authUser"` as soon as auth state becomes truthy, but Vue's reactivity flush runs before the `await` continuation in `submit()` can navigate away from `/login`
- Fix: add a route check in `App.vue` so the authenticated layout only renders when the current route is not an auth page (login, register, password reset, openid). The `NoAuthWrapper` stays visible until `redirectIfSaved()` navigates away, then the authenticated layout renders cleanly
- Also add `redirectIfSaved()` calls after successful login/registration in `submit()` to ensure navigation happens, and fix a race condition in `onBeforeMount` that could clear the saved redirect URL

## Test plan

- [ ] Log in and verify the login form no longer flashes inside the navbar/sidebar shell
- [ ] Register a new user and verify no flash occurs
- [ ] Navigate to a protected route while logged out, log in, and verify redirect back to that route
- [ ] New E2E test `Should not show login form inside authenticated app shell after login` passes (uses MutationObserver to detect simultaneous presence of login form + navbar)
- [ ] All existing login E2E tests pass

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -58,7 +58,7 @@ const baseStore = useBaseStore()
 
 const route = useRoute()
 
-const showAuthLayout = computed(() => authStore.authUser && !AUTH_ROUTE_NAMES.has(route.name as string))
+const showAuthLayout = computed(() => authStore.authUser && typeof route.name === 'string' && !AUTH_ROUTE_NAMES.has(route.name))
 
 useBodyClass('is-touch', isTouchDevice())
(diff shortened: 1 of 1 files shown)


kolaente:
auggie review

PR #2207, opened 2026-02-08
fix(auth): remove unnecessary fields from JWT token payloads

kolaente:
## Summary

- Remove `email`, `name`, `emailRemindersEnabled`, and `isLocalUser` claims from user JWT tokens
- Remove `isLocalUser` claim from link share JWT tokens
- Simplify `GetUserFromClaims` to only extract `id` and `username`
- Remove the now-unnecessary email override in the frontend's `refreshUserInfo`

These fields are never used from the token — the backend always fetches the full user from the database by ID, and the frontend fetches user data from the `GET /user` API endpoint immediately after login. Existing tokens with the old claims continue to work since extra fields are simply ignored.

## Test plan

- [x] Backend web tests pass
- [x] `go vet` and `go build` pass on changed packages
- [x] Frontend lint passes
- [ ] Manual login/logout flow works correctly
- [ ] Link share access works correctly

### DIFF
--- a/frontend/src/stores/auth.ts
+++ b/frontend/src/stores/auth.ts
@@ -325,7 +325,6 @@ export const useAuthStore = defineStore('auth', () => {
 			const newUser = new UserModel({
 				...response.data,
 				...(info.value?.type && {type: info.value?.type}),
-				...(info.value?.email && {email: info.value?.email}),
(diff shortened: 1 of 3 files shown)


kolaente:
auggie review

PR #2210, opened 2026-02-09
feat: Idiomatic Go test behavior with Magefile

JohnStarich:
I started work on [urgency / weighted sort](https://community.vikunja.io/t/introduce-configurable-urgency-mechanism-akin-to-taskwarrior/4007/5) for tasks and ran into some surprises while building, testing, and linting:

* `go test ./...` did not run for me successfully without additional runtime environment variables. I fixed this by `go:embed`ing test fixtures and initializing config variables where necessary.
    - Please spot check me here! This one I had a hard time spotting usage of globals across tasks and package inits. I _think_ I got them all.
    - Both Mage and `go test` implicitly set the current working directory before running tasks or testing packages. I took advantage of this to simplify file paths.
* With `go test` busted for me, I tried Mage. Mage test tasks always resulted in high verbosity test output, which was difficult to read when identifying just failing test results. I've now linked Mage's verbosity to the test verbosity, so we can run `mage -v` to enable when desired.
* Some common task failures resulted in `os.Exit()`, rather than returning contextual errors in tests or other mage tasks. This made troubleshooting pretty tricky. I returned errors instead and deferred print and exit logic to Mage.
* Some mage tasks performed special `go list all` and `find` calls to carefully select sections of packages and files. This worked, but also tested all sorts of packages outside the current Go module, like standard library packages.
    - For the package lists, this appeared to only split by (predominantly) faster tests and slower tests. There's an app for that! I used `go test -short ./...` to only run faster tests, and left the single package selected for slow tests to run on their own.
    - For the file lists, I switched to a Go native `filepath.Walk` to pick up on the same files without the slightly less obvious syntax of `find`. This one wasn't too big a deal, `find` is pretty common.

Surprises aren't always bad, and this may totally be intended behavior, but I wanted to run this set of changes by you all. I believe adopting some common Go idioms like these will ease others' contributions as well as my own.

Feel free to reject this in part or as a whole. I carefully split my commits for a piecemeal revert if we need it or to more easily split into separate, smaller PRs. Just give the word, I'll hack it into smaller bits!

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -330,7 +330,7 @@ func Fmt() error {
 		return nil
 	})
 	if err != nil {
-		panic(err)
+		return err
 	}
 	args := append([]string{"-s", "-w"}, goFiles...)
 	return runAndStreamOutput("gofmt", args...)


pkg/modules/migration/todoist/todoist_test.go
```diff
@@ -48,7 +45,7 @@ func TestConvertTodoistToVikunja(t *testing.T) {
 	dueTimeWithTime = dueTimeWithTime.In(config.GetTimeZone())
 	nilTime, err := time.Parse(time.RFC3339Nano, "0001-01-01T00:00:00Z")
 	require.NoError(t, err)
-	exampleFile, err := os.ReadFile(config.ServiceRootpath.GetString() + "/pkg/modules/migration/testimage.jpg")
+	exampleFile, err := os.ReadFile("../testimage.jpg")
```
kolaente, line 48: Will this work when the test context is something else? Jetbrains IDEs set it to some temp directory iirc.

magefile.go
```diff
@@ -373,45 +316,54 @@ func printSuccess(text string, args ...interface{}) {
 	fmt.Printf(InfoColor+"\n", text)
 }
 
-// Formats the code using go fmt
-func Fmt() {
+// Fmt formats the code using go fmt
```
kolaente, line 319: I get that the comments are not really idiomatic go, but the way mage parses them for use in cli help makes it double if we're starting the function here. Not sure if that's a good reason to keep them as they were though.

pkg/modules/migration/todoist/todoist_test.go
JohnStarich: Yep! It should always use the same directory as its source files (the package). This is from `go help testflag`:

>When 'go test' runs a test binary, it does so from within the
corresponding package's source code directory.

JohnStarich: I believe the temporary directory might be where the compiled test binaries are placed and executed, however their current working directory is set back to their source files' package.

magefile.go
JohnStarich: I thought about that too, but it looks like Mage detects this and strips out the word. Running `mage` locally shows me this:

```
$ mage
Targets:
  build:build                  builds a vikunja binary, ready to run
  build:clean                  cleans all build, executable and bindata files
  build:saveVersionToFile
  check:all                    runs golangci and the swagger test in parallel
  check:golangci
  check:golangciFix
  check:gotSwag                checks if the swagger docs need to be re-generated from the code annotations
  check:translations           checks if all translation keys used in the code exist in the English translation file
  dev:makeEvent                create a new event.
  dev:makeListener             create a new listener for an event.
  dev:makeMigration            creates a new bare db migration skeleton in pkg/migration.
  dev:makeNotification         create a new notification.
  dev:prepareWorktree          creates a new git worktree for development.
  dev:tagRelease               creates a new release tag with changelog.
  fmt                          formats the code using go fmt
  generate:configYAML          create a yaml config file from the config-raw.json definition
  generate:swaggerDocs         generates the swagger docs from the code annotations
  plugins:build                compiles a Go plugin at the provided path.
  release:check                creates sha256 checksum files for each binary in dist/release/
  release:compress             compresses the built binaries in dist/binaries/ to reduce their filesize
  release:copy                 copies all built binaries to dist/release/ in preparation for creating the os packages
  release:darwin               builds binaries for darwin
  release:dirs                 creates all directories needed to release vikunja
  release:linux                builds binaries for linux
  release:osPackage            creates a folder for each
  release:packages             creates deb, rpm and apk packages
  release:prepareNFPMConfig    prepares the nfpm config
  release:release              runs all steps in the right order to create release packages for various platforms
  release:reprepro             creates a debian repo structure
  release:windows              builds binaries for windows
  release:xgo
  release:zip                  creates a zip file from all os-package folders in dist/release
  test:all
  test:coverage                runs the tests and builds the coverage html file from coverage output
  test:feature                 runs the feature tests
  test:filter
  test:web                     runs the web tests
```

Is that sorta what you had in mind?

pkg/modules/migration/todoist/todoist_test.go
kolaente: All right, thanks!

magefile.go
kolaente: Yes! Great to see this should work like that.

kolaente:
Rebased and fixed the conflicts, we should be good to merge when the CI passes

PR #2226, opened 2026-02-11
feat(auth): allow LDAP authentication with anonymous bind

Quiwy:
As discussed on Matrix, Vikunja currently prevents users from using LDAP authentication if the server allows anonymous binds (common in local environments like YunoHost). The application would previously trigger a `log.Fatal` if `AuthLdapBindDN` or `AuthLdapBindPassword` were left empty in the configuration.

#### **How this fixes the problem:**

* **Validation:** Removed the strict requirement for Bind credentials in `InitializeLDAPConnection`.
* **Connection Logic:** Updated `ConnectAndBindToLDAPDirectory` to attempt an `UnauthenticatedBind` from the `go-ldap` library when no credentials are provided.
* **Safety:** If a Bind DN is provided, the behavior remains unchanged (authenticated bind).

#### **Testing:**

* Tested manually on a **YunoHost** instance by replacing the binary.
* Confirmed that Vikunja now successfully starts and authenticates users via the local LDAP (localhost) without requiring a service account.
* Added a basic unit test in `pkg/modules/auth/ldap/ldap_test.go` to ensure the initialization logic doesn't crash with empty credentials.

*Note: This is my first contribution to a Go project (assisted by an LLM for syntax). Feedback on code style is more than welcome!*

### DIFF
--- a/pkg/modules/auth/ldap/ldap.go
+++ b/pkg/modules/auth/ldap/ldap.go
@@ -51,12 +51,6 @@ func InitializeLDAPConnection() {
 	if config.AuthLdapBaseDN.GetString() == "" {
 		log.Fatal("LDAP base DN is not configured")
 	}
-	if config.AuthLdapBindDN.GetString() == "" {
-		log.Fatal("LDAP bind DN is not configured")
-	}
-	if config.AuthLdapBindPassword.GetString() == "" {
-		log.Fatal("LDAP bind password is not configured")
-	}
 	if config.AuthLdapUserFilter.GetString() == "" {
 		log.Fatal("LDAP user filter is not configured")
 	}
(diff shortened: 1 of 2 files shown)


pkg/modules/auth/ldap/ldap.go
```diff
@@ -99,10 +93,16 @@ func ConnectAndBindToLDAPDirectory() (l *ldap.Conn, err error) {
...
 	}
 
-	err = l.Bind(
-		config.AuthLdapBindDN.GetString(),
-		config.AuthLdapBindPassword.GetString(),
-	)
+	bindDN := config.AuthLdapBindDN.GetString()
+	bindPassword := config.AuthLdapBindPassword.GetString()
+
+	if bindDN != "" {
```
kolaente: This should check if the `bindPassword` is empty as well

pkg/modules/auth/ldap/ldap.go
```diff
@@ -99,10 +93,16 @@ func ConnectAndBindToLDAPDirectory() (l *ldap.Conn, err error) {
...
+		// Standard authentication
+		err = l.Bind(bindDN, bindPassword)
+	} else {
+		// Anonymous bind attempt (depending on the server, this call is explicit or automatic)
+		err = l.UnauthenticatedBind("")
```
kolaente, line 105: Please log this case as `log.Info` so that it's clear to users why this is happening.

Quiwy:
Hi @kolaente, are the changes ok for you ?

PR #2233, opened 2026-02-13
feat(task): allow changing bucket from task detail view

### LINKED ISSUE #2167: Move between buckets when looking at a task
### Pre-submission checklist

### Description

It would be nice if you could change the bucket from the task detail screen. Now you have to close the detail screen and then drag-and-drop. Which works, but sometimes you forget which task you were looking at exactly.

_No response_

larsderidder:
Shows the current kanban bucket after the project name in the task detail subtitle. Clicking it opens a dropdown to move the task to a different bucket without leaving the detail view.

image

If the project has multiple kanban views, each bucket is prefixed with the view name.

Closes #2167

### DIFF
--- a/frontend/tests/e2e/task/bucket-select.spec.ts
+++ b/frontend/tests/e2e/task/bucket-select.spec.ts
@@ -80,6 +80,100 @@ test.describe('Task Bucket Select', () => {
 		await expect(page.locator('.task-view .subtitle .bucket-name')).not.toBeVisible()
 	})
 
+	test.describe('Multiple kanban views', () => {
+		async function createTaskWithMultipleKanbanViews() {
+			const projects = await ProjectFactory.create(1)
+			const listView = (await ProjectViewFactory.create(1, {
+				id: 1,
+				project_id: projects[0].id,
+				view_kind: 0,
+			}))[0]
+			const kanbanView1 = (await ProjectViewFactory.create(1, {
+				id: 2,
+				project_id: projects[0].id,
+				view_kind: 3,
+				bucket_configuration_mode: 1,
+			}, false))[0]
+			const kanbanView2 = (await ProjectViewFactory.create(1, {
+				id: 3,
+				project_id: projects[0].id,
+				view_kind: 3,
+				bucket_configuration_mode: 1,
+			}, false))[0]
+			const bucketsView1 = await BucketFactory.create(2, {
+				project_view_id: kanbanView1.id,
+			})
+			const bucketsView2 = await BucketFactory.create(2, {
+				id: (i: number) => i + 2,
+				project_view_id: kanbanView2.id,
+			}, false)
+			const tasks = await TaskFactory.create(1, {
+				project_id: projects[0].id,
+			})
+			await TaskBucketFactory.create(1, {
+				task_id: tasks[0].id,
+				bucket_id: bucketsView1[0].id,
+				project_view_id: kanbanView1.id,
+			})
+			await TaskBucketFactory.create(1, {
+				task_id: tasks[0].id,
+				bucket_id: bucketsView2[0].id,
+				project_view_id: kanbanView2.id,
+			}, false)
+			return {
+				project: projects[0],
+				listView,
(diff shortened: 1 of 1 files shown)


frontend/src/components/tasks/partials/BucketSelect.vue
```diff
@@ -0,0 +1,206 @@
...
+						>{{ kanbanView.title }}:</span>
+						{{ currentBucketTitle(kanbanView) }}
+						<Icon
+							icon="pencil-alt"
+							class="change-indicator"
+						/>
+					</BaseButton>
+				</template>
+				<DropdownItem
+					v-for="bucket in viewBuckets[kanbanView.id] || []"
```
kolaente: Unsure if this is a good way to solve this, the UI will get pretty crowded when there are multiple kanban boards for this task.

I get that the place at the top is logical, but we'll need a better way when there is more than one kanban board.

frontend/src/components/tasks/partials/BucketSelect.vue
```diff
@@ -0,0 +1,206 @@
...
+				continue
+			}
+			try {
+				const buckets = await bucketService.getAll({
+					projectId: props.task.projectId,
+					projectViewId: view.id,
+				} as IBucket)
+				viewBuckets.value[view.id] = buckets
+			} catch {
+				// silently ignore if we cannot load buckets
```
kolaente: we should at least log the error

frontend/src/components/tasks/partials/BucketSelect.vue
```diff
@@ -0,0 +1,206 @@
...
+
+		// If the task was not yet in this view, add the bucket
+		if (!updatedBuckets.find(b => b.projectViewId === view.id)) {
+			updatedBuckets.push({...bucket})
+		}
+
+		// Update the kanban store if the board is loaded
+		kanbanStore.moveTaskToBucket(props.task, bucket.id)
+
+		emit('update:task', {
```
kolaente: This does not update the done status when the task is moved to the done bucket, creating a confusing ui state

frontend/src/components/tasks/partials/BucketSelect.vue
```diff
@@ -0,0 +1,206 @@
...
+			...props.task,
+			buckets: updatedBuckets,
+			bucketId: bucket.id,
+		})
+
+		success({message: t('task.detail.bucketChangedSuccess')})
+	} catch {
+		// error is handled by the service layer
```
kolaente: it's only handled there if you don't catch it. This will just ignore the error, please don't do that

larsderidder:
Fair review. Updated the dropdown to only show the buckets for the current view. Improved the other things.

frontend/src/components/tasks/partials/BucketSelect.vue
```diff
@@ -81,106 +70,107 @@ const {t} = useI18n({useScope: 'global'})
 
 const projectStore = useProjectStore()
 const kanbanStore = useKanbanStore()
+const baseStore = useBaseStore()
 
 const project = computed(() => projectStore.projects[props.task.projectId])
 
-const kanbanViews = computed(() => {
+// Only show the bucket selector when the active view is a manual kanban view
```
kolaente: Won't this mean that I can't change a bucket when accessing the task from the list view?

larsderidder: Yes, intentionally. Showing a bucket from an arbitrary kanban view when you're on a list view seemed more confusing (per the earlier feedback about multiple views).

But I gave it some more thought, and I suppose we can easily detect that there is only a single view; then it's safe to show. I made that change, which should work for most users (assuming most only use a single kanban view, which is an assumption of course).

In case of multiple views, the only explicit alternative would be to somehow add a kanban view selector as well, but that would break the clean UI. I don't think users would appreciate that, but open to suggestions.

kolaente: I think that's a good way to go about this.

PR #2258, opened 2026-02-18
feat: clickable task notifications

mnadalin:
Make task-related notifications clickable, allowing users to directly open the corresponding task.

### DIFF
--- a/frontend/src/components/notifications/Notifications.vue
+++ b/frontend/src/components/notifications/Notifications.vue
@@ -170,7 +170,7 @@ function notificationHasRoute(n: INotification): boolean {
 function to(n: INotification, index: number) {
 	return async () => {
 		const route = getNotificationRoute(n)
-		if (route === null) return;
+		if (route === null) return
 		
 		const failure = await router.push(route)
 		if (isNavigationFailure(failure, NavigationFailureType.duplicated)) {


frontend/src/components/notifications/Notifications.vue
```diff
@@ -29,6 +29,8 @@
 					v-for="(n, index) in notifications"
 					:key="n.id"
 					class="single-notification"
+					:class="{'is-task-notification': isTaskNotification(n)}"
+					@click="() => { if (isTaskNotification(n)) to(n, index)() }"
```
kolaente: Please do that `if` in a function, not inline

frontend/src/components/notifications/Notifications.vue
```diff
@@ -48,7 +50,11 @@
 							>
 								{{ getDisplayName(n.notification.doer) }}
 							</span>
+							<template v-if="isTaskNotification(n)">
+								{{ n.toText(userInfo) }}
+							</template>
 							<BaseButton
+								v-else
 								class="has-text-start"
 								@click="() => to(n, index)()"
```
kolaente: why not just move this up?

kolaente (commented):
Thanks for the PR!

Please check out my comments.

frontend/src/components/notifications/Notifications.vue
```diff
@@ -179,12 +189,19 @@ function to(n, index) {
...
+				router.go(0)
+			}
 		}
 
 		n.read = true
-		const notificationService = new NotificationService()
-		allNotifications.value[index] = await notificationService.update(n)
+		if (allNotifications.value[index]) {
+			const notificationService = new NotificationService()
+			Object.assign(allNotifications.value[index], await notificationService.update(n))
```
kolaente, line 183: Why `Object.assign`?

frontend/src/components/notifications/Notifications.vue
```diff
@@ -150,8 +147,21 @@ function hidePopup(e) {
 	}
 }
 
+const CLICKABLE_NOTIFICATIONS: string[] = [
```
kolaente: This is essentially doubled with the `switch` block in `to`. Can you refactor that so that it calls a new function that runs the switch and generates the parameter and then to determine if a notification is clickable we'll just have to check if anything is returned from the switch.

frontend/src/components/notifications/Notifications.vue
mnadalin: A direct assignment caused the notification to "bump": the .created element disappear for a fraction of second.

frontend/src/components/notifications/Notifications.vue
```diff
@@ -147,52 +147,34 @@ function hidePopup(e) {
...
-	return CLICKABLE_NOTIFICATIONS.includes(n.name)
-}
-
-function to(n, index) {
-	const to: { name: string, params: Record<string, string | number> } = {
-		name: '',
-		params: {},
-	}
-
+function getNotificationRoute(n: INotification): object | null {
```
kolaente: Please use the correct type here instead of `object`.

kolaente (changes requested):

PR #2259, opened 2026-02-18
feat(gantt): show tasks with partial dates using gradient fade

kolaente:
## Summary

- Display tasks in the Gantt chart that have only a start date, only an end date, or only a due date, using a fade-out gradient on the open-ended side
- Partial-date tasks get a default 7-day span for visible width, with the open-ended side fading to transparent
- Resize handles are hidden on the open-ended side; drag/resize updates the correct date fields
- API filter updated to also fetch tasks by `end_date` or `due_date` within the visible range
- The "show tasks without dates" toggle now only affects fully dateless tasks

Here's what it looks like:

image


kolaente:
@codex review

kolaente:
auggie review

PR #2260, opened 2026-02-18
fix: merge AND-joined sub-table filters into single EXISTS subquery

### LINKED ISSUE #2245: Reminder filters with multiple conditions incorrectly match tasks with multiple reminders using rest API
### Pre-submission checklist

### Problem
When filtering tasks by reminders using multiple range conditions joined with &&, tasks with multiple reminders can be incorrectly included even if no single reminder satisfies all conditions.

### Example:
GET /api/v1/tasks?filter=reminders > now && reminders < now+5m

**Expected behavior:** Return only tasks that have at least one reminder in the (now, now+5m) window.

**Actual behavior:** Tasks with multiple reminders can be included even if:
One reminder is before now, another reminder is after now+5m for a task wich has no single reminder is actually in the (now, now+5m) window

Note: This bug only occurs when a task has multiple reminders. Tasks with a single reminder work correctly.

**How to reproduce:**
1. Create a task with multiple reminders, where none fall within the target window
2. Set a time window that excludes all reminders, and run the api GET request. For example, use a window from now to now+5m:
   GET /api/v1/tasks?filter=reminders%20%3E%20now%20%26%26%20reminders%20%3C%20now%2B5m&filter_include_nulls=false   
3. Observe the result: the task is returned

### Root cause
The bug is in convertFiltersToDBFilterCond in pkg/models/task_search.go. When processing filters on sub-table fields (like reminders), each filter condition is converted into a separate EXISTS subquery:
EXISTS (SELECT 1 FROM task_reminders WHERE tasks.id = task_id AND reminder > now)
AND
EXISTS (SELECT 1 FROM task_reminders WHERE tasks.id = task_id AND reminder < now+5m)

With a single reminder: Both EXISTS checks must pass on the same row, so it behaves correctly.
With multiple reminders: 
- Each EXISTS can be satisfied by different rows:
- First EXISTS can match a reminder at now+10m (satisfies > now)
- Second EXISTS can match a reminder at now-10m (satisfies < now+5m)
- Result: Task is incorrectly included even though no single reminder is in the window.

The problematic code was in the sub-table filter handling section: task_search.go in function convertFiltersToDBFilterCond

### Vikunja Version

v1.1.0

### Browser and version

_No response_

Yes

### Screenshots

_No response_

kolaente:
## Summary

- When multiple AND-joined filter conditions target the same sub-table (e.g., `reminders > X && reminders < Y`), they are now combined into a single `EXISTS` subquery so that all conditions must be satisfied by the same row
- Previously, each condition generated a separate `EXISTS` subquery that could match different rows, causing false positives (e.g., a task with reminders at `2018-08-01` and `2019-03-01` would incorrectly match a filter for `reminders > 2018-10-01 && reminders < 2018-12-10`)
- OR-joined conditions and conditions targeting different sub-tables continue to produce separate subqueries as before

Fixes #2245


kolaente:
auggie review

PR #2263, opened 2026-02-19
feat(comments): add sort order toggle for task comments

kolaente:
## Summary
- Adds a toggle button next to the "Comments" heading that lets users switch between oldest-first and newest-first comment sort order
- Persists the preference in frontend settings via the existing user settings API
- Button only appears when at least one comment exists; tooltip shows the action that will happen on click


kolaente:
auggie review

PR #2272, opened 2026-02-20
feat: add include subprojects task view support

### LINKED ISSUE #1529: Display tasks from subprojects.
### Description

I think [#559](https://github.com/go-vikunja/vikunja/issues/559) ended up being only about the breadcrumbs, I would like to second the idea of displaying sub-project tasks.

I have a project that has a bunch of subprojects that each have tasks. I would like to be able to see all tasks of subprojects.

Checking in "overview" shows all tasks, with no way to filter down, similarly with 'Upcoming'.

### LINKED ISSUE #2086: Project views with tasks from other (sub)project
### Description

When I create project A and sub project B, C, D and F. Then I'd like to make a view in project A that views all tasks from B, C, D and F.
When I apply the filter:
done = false && (
  project = ProjectA ||
  project = ProjectB ||
  project = ProjectC ||
  project = ProjectD ||
  project = ProjectE)

Then it will only display ProjectA tasks.

Luckily, the filter works in a filter trough projects, but that filter isn't shareable. Which could be another feature request.

Filter in projects. They can't be shared though.

JamesBatten:
## What
Adds an optional **“Include subprojects”** toggle in a project view.  
When enabled, the task list for the current project also includes tasks from all descendant subprojects, and tasks from other projects show which project they belong to.

Default behavior is unchanged (toggle off).

## API
Adds support for an optional query parameter on the project-view tasks endpoint:

- `GET /projects/{id}/views/{view}/tasks?include_subprojects=true`

When enabled, the backend expands the project set to include all readable descendants and returns tasks from all of them.

## UX / Scope
- Supported on flat task views (e.g. List/Table/Gantt).
- Not enabled by default.
- Tasks from subprojects are visually marked with their source project.

## How to test
1. Create a parent project `A` and a subproject `B` under it.
2. Create tasks in both projects.
3. Open project `A` in the frontend.
4. Toggle **Include subprojects**:
   - OFF: only tasks from `A` appear.
   - ON: tasks from `A` + `B` appear; tasks from `B` show a “from B” indicator.

## Screenshots (after)
- Toggle OFF:
 
Screenshot From 2026-02-20 14-33-48

- Toggle ON:

Screenshot From 2026-02-20 14-34-15

Resolves https://github.com/go-vikunja/vikunja/issues/2086
Resolves https://github.com/go-vikunja/vikunja/issues/1529

## Tests
- Backend: go test ./...
- Frontend: pnpm lint  ,  pnpm build

### DIFF
--- a/frontend/src/components/gantt/GanttChart.vue
+++ b/frontend/src/components/gantt/GanttChart.vue
@@ -246,7 +246,12 @@ function getRoundedDate(value: string | Date | undefined, fallback: Date | strin
 }
 
 function getTaskLabel(task: ITask): string {
-	if (!filters.value.includeSubprojects || task.projectId === filters.value.projectId) {
+	if (!filters.value.includeSubprojects) {
+		return task.title
+	}
+
+	const isProjectContext = filters.value.projectId > 0
+	if (isProjectContext && task.projectId === filters.value.projectId) {
 		return task.title
 	}
 
--- a/frontend/src/components/gantt/GanttChart.vue
+++ b/frontend/src/components/gantt/GanttChart.vue
@@ -102,6 +102,7 @@ import type {ITask, ITaskPartialWithId} from '@/modelTypes/ITask'
 import type {DateISO} from '@/types/DateISO'
 import type {GanttFilters} from '@/views/project/helpers/useGanttFilters'
 import type {GanttBarModel, GanttBarDateType} from '@/composables/useGanttBar'
+import {useProjectStore} from '@/stores/projects'
 
 import GanttChartBody from '@/components/gantt/GanttChartBody.vue'
 import GanttRow from '@/components/gantt/GanttRow.vue'
@@ -129,6 +130,7 @@ const emit = defineEmits<{
 const DAY_WIDTH_PIXELS = 30
 
 const {tasks, filters} = toRefs(props)
+const projectStore = useProjectStore()
 
 const dayjsLanguageLoading = useDayjsLanguageSync(dayjs)
 const ganttContainer = ref(null)
@@ -243,6 +245,19 @@ function getRoundedDate(value: string | Date | undefined, fallback: Date | strin
 	return roundToNaturalDayBoundary(value ? new Date(value) : new Date(fallback), isStart)
 }
 
+function getTaskLabel(task: ITask): string {
+	if (!filters.value.includeSubprojects || task.projectId === filters.value.projectId) {
+		return task.title
+	}
+
+	const projectTitle = projectStore.projects[task.projectId]?.title
+	if (!projectTitle) {
+		return task.title
+	}
+
+	return `${task.title} · ${projectTitle}`
+}
+
 function transformTaskToGanttBar(node: GanttTaskTreeNode): GanttBarModel {
 	const t = node.task
 	const DEFAULT_SPAN_DAYS = 7
@@ -284,7 +299,7 @@ function transformTaskToGanttBar(node: GanttTaskTreeNode): GanttBarModel {
 		start: startDate,
 		end: endDate,
 		meta: {
-			label: t.title,
+			label: getTaskLabel(t),
 			task: t,
 			color: taskColor,
 			hasActualDates: Boolean(t.startDate && (t.endDate || t.dueDate)),
--- a/frontend/src/components/gantt/GanttChart.vue
+++ b/frontend/src/components/gantt/GanttChart.vue
@@ -118,6 +118,7 @@ import {roundToNaturalDayBoundary} from '@/helpers/time/roundToNaturalDayBoundar
 const props = defineProps<{
 	isLoading: boolean,
 	filters: GanttFilters,
+	includeSubprojects: boolean,
 	tasks: Map<ITask['id'], ITask>,
 	defaultTaskStartDate: DateISO
 	defaultTaskEndDate: DateISO
@@ -246,7 +247,7 @@ function getRoundedDate(value: string | Date | undefined, fallback: Date | strin
 }
 
 function getTaskLabel(task: ITask): string {
-	if (!filters.value.includeSubprojects) {
+	if (!props.includeSubprojects) {
 		return task.title
 	}
 
--- a/frontend/src/components/project/views/ProjectList.vue
+++ b/frontend/src/components/project/views/ProjectList.vue
@@ -14,6 +14,14 @@
 					:project-id="projectId"
 					@update:modelValue="prepareFiltersAndLoadTasks()"
 				/>
+				<FancyCheckbox
+					v-if="projectId > 0"
+					v-model="includeSubprojects"
+					v-tooltip="$t('project.views.includeSubprojectsHint')"
+					class="include-subprojects-toggle"
+				>
+					{{ $t('project.views.includeSubprojects') }}
+				</FancyCheckbox>
 			</div>
 		</template>
 
@@ -109,6 +117,7 @@ import SingleTaskInProject from '@/components/tasks/partials/SingleTaskInProject
 import FilterPopup from '@/components/project/partials/FilterPopup.vue'
 import Nothing from '@/components/misc/Nothing.vue'
 import Pagination from '@/components/misc/Pagination.vue'
+import FancyCheckbox from '@/components/input/FancyCheckbox.vue'
 import {ALPHABETICAL_SORT} from '@/components/project/partials/Filters.vue'
 
 import {useTaskList} from '@/composables/useTaskList'
@@ -149,6 +158,7 @@ const {
 	loadTasks,
 	params,
 	sortByParam,
+	includeSubprojects,
 } = useTaskList(
 	() => projectId.value,
 	() => props.viewId,
@@ -199,7 +209,7 @@ onMounted(async () => {
 	ctaVisible.value = true
 })
 
-const canDragTasks = computed(() => canWrite.value || isSavedFilter(project.value))
+const canDragTasks = computed(() => (canWrite.value || isSavedFilter(project.value)) && !includeSubprojects.value)
 
 const isTouchDevice = ref(false)
 if (typeof window !== 'undefined') {
@@ -416,5 +426,9 @@ onBeforeUnmount(() => {
 	:deep(.card) {
 		margin-block-end: 0;
 	}
+}
+
+.include-subprojects-toggle {
+	margin-inline-start: .75rem;
 }
 </style>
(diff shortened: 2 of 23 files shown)


AlvinVrolijk:
Love this

frontend/src/components/gantt/GanttChart.vue
```diff
@@ -154,6 +156,19 @@ function getRoundedDate(value: string | Date | undefined, fallback: Date | strin
 	return roundToNaturalDayBoundary(value ? new Date(value) : new Date(fallback), isStart)
 }
 
+function getTaskLabel(task: ITask): string {
+	if (!filters.value.includeSubprojects || task.projectId === filters.value.projectId) {
```
kolaente: This will break for saved filters where the task's project id is not the filter id.

frontend/src/components/gantt/GanttChart.vue
```diff
@@ -154,6 +156,19 @@ function getRoundedDate(value: string | Date | undefined, fallback: Date | strin
...
+		return task.title
+	}
+
+	const projectTitle = projectStore.projects[task.projectId]?.title
+	if (!projectTitle) {
+		return task.title
+	}
+
+	return `${task.title} · ${projectTitle}`
```
kolaente, line 264: I'm unsure if this is really the best way to show the project here. Maybe we can find a way to do this similar to how it's solved in list and kanban views? Really not sure.

frontend/src/components/project/views/ProjectList.vue
```diff
@@ -14,6 +14,14 @@
 					:project-id="projectId"
 					@update:modelValue="prepareFiltersAndLoadTasks()"
 				/>
+				<FancyCheckbox
```
kolaente: Please make this a view setting instead - adding a new checkbox to the project view "inline" for a feature not everyone will use is not really the best use of space. Instead, people will look in the view settings for this.

It should still be possible to pass the query parameter to the api endpoint, but having it as a view setting would allow to persist it.

frontend/src/components/project/views/ProjectTable.vue
```diff
@@ -226,6 +234,12 @@
 												{{ t.title }}
 											</RouterLink>
 										</TaskGlanceTooltip>
+										<span
+											v-if="includeSubprojects && t.projectId !== projectId && projectStore.projects[t.projectId]"
```
kolaente: Since there's a project column as well, showing the project here as well will double that, which makes it redundant.

frontend/src/components/tasks/partials/SingleTaskInProject.vue
```diff
@@ -285,6 +285,44 @@ const currentProject = computed(() => {
 	} : baseStore.currentProject
 })
 
+const projectLabel = computed(() => {
```
kolaente, line 289: Why not just show the project title?

frontend/src/composables/useTaskList.ts
```diff
@@ -73,6 +73,14 @@ export function useTaskList(
 	const page = useRouteQuery('page', '1', { transform: Number })
 	const filter = useRouteQuery('filter')
 	const s = useRouteQuery('s')
+	const includeSubprojectsQuery = useRouteQuery('includeSubprojects')
+
+	const includeSubprojects = computed<boolean>({
```
kolaente: `useRouteQuery` can do transforms, please use that instead

https://vueuse.org/router/useRouteQuery/#useroutequery

frontend/src/views/project/helpers/useGanttFilters.ts
```diff
@@ -76,6 +79,10 @@ function ganttFiltersToRoute(filters: GanttFilters): RouteLocationRaw {
 		query.showTasksWithoutDates = String(filters.showTasksWithoutDates)
 	}
 
+	if (filters.includeSubprojects) {
+		query.includeSubprojects = '1'
```
kolaente: This should be `true`

pkg/models/task_collection.go
```diff
@@ -196,7 +202,58 @@ func getRelevantProjectsFromCollection(s *xorm.Session, a web.Auth, tf *TaskColl
 		}
 	}
 
-	return []*Project{{ID: tf.ProjectID}}, nil
+	if !tf.IncludeSubprojects || tf.ProjectID <= 0 {
+		return []*Project{{ID: tf.ProjectID}}, nil
+	}
+
+	projects, _, _, err = getRawProjectsForUser(
```
kolaente: Please use a recursive sql query (like it's done in `getRawProjectsForUser`). Check the `setArchiveStateForProjectDescendants` function, that uses a similar query.

kolaente (changes requested):
Hey thanks for the PR!

I have a few suggestions, the most important one is moving the checkbox to the view settings instead of having a new checkbox on every page.

frontend/src/components/project/views/ProjectList.vue
JamesBatten: This isn't intended to be a per-project setting that persists. The idea is to be able, for any given project, to rapidly flip between the toggle being on (seeing the subproject tasks) and the toggle being off (seeing only the immediate project-level tasks). Going into the view settings each time is too many clicks.

JamesBatten:
Thanks for the review — I pushed a follow-up commit addressing the requested changes.

### What I changed
- Reworked backend descendant expansion for `include_subprojects` to use a recursive SQL CTE (instead of in-memory traversal)
- Added/updated backend tests for descendant resolution and non-project-context regression
- Updated route query handling to use `useRouteQuery` transforms
- Changed Gantt query param to use boolean `true` (not `'1'`)
- Fixed Gantt label logic so it doesn’t assume project context IDs in saved-filter / non-project contexts
- Removed the redundant extra project label in table view
- Simplified `SingleTaskInProject` project label to show project title only

### About the toggle placement
I did not move the toggle into the view menu for this iteration.
Instead, I addressed the UI clutter concern by adding a user setting:
- **default OFF**
- when enabled, the inline “Include subprojects” toggle becomes visible in project views

image

This keeps the feature discoverable for users who want it while keeping the default UI unchanged.

I also ran local checks (`go test ./...`, frontend lint/build/unit tests, `vue-tsc --noEmit`).

frontend/src/components/project/views/ProjectList.vue
JamesBatten: I've addressed this in the latest commit. See explanation here: https://github.com/go-vikunja/vikunja/pull/2272#issuecomment-3937642916

AlvinVrolijk:
Looks so good!

Maybe a env or config setting (preferably both) to set default setting for new users? (or when migrating from non-existing seting)

Also possible to maybe persist the toggle in browser cache for that specific project, so it stays the same when browsing or changing views within the project?

[EDIT]
I added those in a PR to his fork for this PR.

JamesBatten:
I guess the maintainers have to choose which one they want, between the toggle being present in the user settings or always shown. The user settings option seems like a good compromise.

kolaente:
Thanks for the changes!

Moving this to user settings a good idea, but it should still be a per-view setting as well. Maybe users don't want to enable this everywhere.

For the button placement, I'm not sure if that's a good placement next to the filter button. Can you move it to the filter popup? It's not exactly belonging there as well, but we have similar setting there.

AlvinVrolijk:
Is this also with the persisting?
Or maybe in the view settings or project settings a toggle?

JamesBatten:
Pushed an update addressing the latest feedback.

Changes in this commit:

- Per-view setting: “Include subprojects” is now stored per project view and persists (so each view remembers its last state when you come back).

- Global visibility toggle kept: The user setting “Show ‘Include subprojects’ toggle in project views” still controls whether the option is shown at all (default off).

- Moved into Filters popup: The toggle is no longer next to the Filters button; it now lives inside the Filters popup.

- Also in View Settings: Added the same option to the Project View Settings, so it can be configured there too.

- Sync: Filters popup + View Settings stay in sync (they update the same per-view setting).


Updated toggle in settings:
image


New toggle in filter popup:
image


When set to active in project view:
image

kolaente:
Much better!

Does that mean the checkbox in the filter popup will only show up when the setting is enabled? Because in that case I think we don't need the setting.

Please also rebase this on main and resolve the conflicts.

JamesBatten:
### Pushed an update removing the global visibility setting and making the option always available.

<br>

**Changes in this update**:

- Removed the user setting that controlled whether the “Include subprojects” option is shown.

- The Include subprojects toggle is now always visible in project views:

    - inside the Filters popup

    - and in Project View Settings

- Kept the behavior as a per-view setting (persists per view and survives navigation / reload).

- Filters + View Settings stay in sync (both update the same per-view state).

**Checks run locally**:

- go test ./...

- pnpm lint, pnpm build, pnpm exec vue-tsc --noEmit, pnpm test:unit

<br>
Updated user settings with the global toggle no longer present:
image


<br>
<br>
Toggle still present in filter popup:
image

dodzod:
@kolaente @JamesBatten I don't have any development experience, but is there anything I can do to help finish this off? It's a feature I'd really like to use and it seems to be almost complete.

JamesBatten:
I should have some time this week, I can take another look at it

dodzod:
That would be brilliant, thank you very much.

dodzod:
Struggled with trying to use GitHub to merge this in my own fork, but after giving up on that I merged this with version 2.3.0 in a local copy and built it successfully - seems to work just fine. Thanks again @JamesBatten for developing the functionality. Would be good to get it in an official release eventually but what I have now works nicely for me.

lindonm:
@JamesBatten , like @dodzod  I don't have any dev experience either - but i'm also very keen on seeing this feature merged in.

Currently we're running my own vibe coded (my first attempt at vibe coding - probably terrible but it does actually work lol) version here:  https://github.com/go-vikunja/vikunja/pull/2801 (which was rejected and probably with good reason :) )

My version is working great for us, I too went with the toggle on every screen - My Kanban is a bit of a mess and in hindsight I wouldn't show the child projects on this one at all (just disable the toggle). I'd really rather not have to maintain my own fork (especially in light of recent security updates).

If there's anything I can do to help out with testing or whatever, please let me know.

Being able to see the hierarchy of all projects/tasks in the table & gantt views is a game changer for us.

Differences between my and your (ghcr.io/go-vikunja/vikunja:pr-2272) version (if it helps at all)
 - User profile setting to set default value (showchildtasks true or false)
 - View toggle applies to List, Gantt and Table views (instead of being in filters)
 - Kanban the only workable way I found was to have a "kanban row" per project - A bit clunky

Other than those minor differences, they are functionally the same.

JamesBatten:
@kolaente I see these code changes have been revived in https://github.com/go-vikunja/vikunja/pull/2939

Should we close this pull request in favour of that one?

kolaente:
yeah let's close this in favor of https://github.com/go-vikunja/vikunja/pull/2939

Issue #2270, opened 2026-02-20
UI profile name glitch


winkelement:
### Pre-submission checklist

- [x] I have searched for existing open or closed issue reports with the same problem.

### Description

When switching between pages in Vikunja (for example from `/` to `/projects`) the **Name** of the user glitches in the upper right corner. For a short time the `username` is visible, then the Name is shown. This only happens when a Name is set.

Screenrecording is from the demo instance after adding a Name to user demo:

https://github.com/user-attachments/assets/54c5d27b-a3b9-492b-824d-1ba54c872de1

The Problem is much more visible when dealing with oidc users, maybe has something to do with #2217.  

### Vikunja Version

1.1.0

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
I was able to verify this. Will look into a fix.

PR #2276, opened 2026-02-21
fix: eliminate display name flash on page navigation

kolaente:
## Summary

- **Fix broken throttle in `checkAuth()`**: The throttle condition checked `lastUserInfoRefresh > now + 1 minute`, which is never true. Changed to `now - 1 minute` so it actually prevents redundant API calls within a one-minute window.
- **Skip `setUser()` from JWT when user is already loaded**: The JWT token lacks the `name` field, so every route navigation was overwriting the full user object with incomplete data, causing the display name to briefly flash to the username. Now we skip `setUser()` if a complete user object with the same ID is already in the store.
- **Add unit tests for `getDisplayName`**: Covers both the display name and username fallback paths.

Fixes https://github.com/go-vikunja/vikunja/issues/2270

## Test plan

- [ ] Log in as a user with a display name different from username
- [ ] Navigate between pages and verify the display name in the top-right corner remains stable (no flash to username)
- [ ] Verify a user with no display name set still shows their username correctly
- [ ] Verify first login still works (JWT data used when no prior user loaded)
- [ ] Verify switching accounts loads the new user's info correctly
- [ ] Unit tests pass: `cd frontend && pnpm test:unit src/models/user.test.ts`

### DIFF
--- a/frontend/src/stores/auth.ts
+++ b/frontend/src/stores/auth.ts
@@ -452,6 +452,7 @@ export const useAuthStore = defineStore('auth', () => {
 		removeToken()
 		const loggedInVia = getLoggedInVia()
 		window.localStorage.clear() // Clear all settings and history we might have saved in local storage.
+		lastUserInfoRefresh.value = null
 		await router.push({name: 'user.login'})
 		await checkAuth()
 


kolaente:
auggie review

PR #2277, opened 2026-02-21
fix: detect mime type for attachments and improve download headers

kolaente:
## Summary
- Detect and store MIME type automatically when creating file attachments using `mimetype.DetectReader` in `files.Create`, so all callers (task attachments, unsplash backgrounds, project duplication) get MIME detection for free
- Set `Content-Disposition: attachment` instead of `inline` so browsers trigger a file download with the correct filename
- Move shared response headers (Content-Type, Content-Length, Last-Modified) out of the S3-specific branch so they apply to both local and S3 storage backends


kolaente:
auggie review

PR #2284, opened 2026-02-23
feat: Support hiding tasks with a future start date

AlD:
Allows hiding tasks that have a future start date.

e.g. discussed here: https://community.vikunja.io/t/hide-tasks-where-start-date-is-in-future/588


frontend/src/i18n/lang/de-DE.json
```diff
@@ -94,6 +94,7 @@
 				"playSoundWhenDone": "Einen Ton abspielen, wenn Aufgaben als erledigt markiert werden",
 				"allowIconChanges": "Zu bestimmten Zeiten spezielle Logos anzeigen",
 				"alwaysShowBucketTaskCount": "Immer die Anzahl der Aufgaben pro Kanban-Spalte anzeigen",
+				"hideFutureTasks": "Zukünftige Aufgaben in der Übersicht standardmäßig ausblenden",
```
kolaente: Please don't update translations here, they are synced through crowdin.

See https://vikunja.io/docs/translations/

frontend/src/views/Home.vue
```diff
@@ -39,6 +39,9 @@
 			v-if="projectStore.hasProjects"
 			:key="showTasksKey"
 			:label-ids="labelIds"
+			:hide-future="hideFuture"
```
kolaente: The tasks on the home page should not rely on query parameters, if you want to customize what you see on the dashboard, use a saved filter.

frontend/src/views/tasks/ShowTasks.vue
```diff
@@ -266,7 +286,19 @@ async function loadPendingTasks(from: Date|string, to: Date|string, filterId: nu
 		projectId = filterId
 	}
 
-	tasks.value = await taskStore.loadTasks(params, projectId)
+	const rawTasks = await taskStore.loadTasks(params, projectId)
+	if (props.hideFuture && showAll.value) {
+		const now = new Date()
+		tasks.value = rawTasks.filter(t => {
```
kolaente, line 292: Filtering should happen with a date query in the api, not on the client

Marck:
Would love to have this implemented, I too am longing for more cleaner overviews and not having all the future tasks in my overview

kolaente (changes requested):
Hey thanks for the PR!

There are a few things that need changing, see my comments below.

frontend/src/views/tasks/ShowTasks.vue
AlD: I agree, and that's where I initially implemented this, but I immediately ran into unintended side-effects caused by `filter_include_nulls` applying to all queried fields.

I'd be happy to implement some sort of per-field `NULL` expression handling, but would like to make sure that this is a direction you'd agree with.

kolaente: Yes, proper `null` filtering would be the way to solve this. This is a feature that I'd like to build eventually, happy to merge a PR if you want to have a go at it.

kolaente:
Can you check the failing test?

Marck:
@AlD kind reminder

PR #2286, opened 2026-02-24
fix(tasks): show drag handle icon on mobile devices

### LINKED ISSUE #2228: Can't drag to change the position of the task on android browser
### Pre-submission checklist

### Description

I found that we can't drag up or down to change the position of the task on android browser. But on the old version ,there was a drag icon on the right side of the task so we can drag a task on a phone easily.

### Vikunja Version

v1.0.0

### Browser and version

Android Kiwi Browser v124.0.6327.4

Yes

### Screenshots

_No response_

kolaente:
## Summary
- Adds a `grip-lines` icon to task rows in the list view, visible only on touch/mobile devices
- Uses `@media (hover: hover) and (pointer: fine)` to hide the handle on desktop
- The whole row remains draggable on all devices — the icon is purely a visual affordance

Resolves #2228

## Test plan
- [x] On a mobile device or using Chrome DevTools device emulation, verify the grip-lines icon appears to the left of each task
- [x] On desktop with a mouse, verify the icon is hidden
- [x] Verify drag-and-drop still works on both mobile and desktop
- [x] Verify the handle only appears when the user has write permissions (i.e. `canDragTasks` is true)

### DIFF
--- a/frontend/src/components/project/views/ProjectKanban.vue
+++ b/frontend/src/components/project/views/ProjectKanban.vue
@@ -149,8 +149,7 @@
 								<draggable
 									v-bind="DRAG_OPTIONS"
 									:handle="taskDragHandle"
-									:delay-on-touch-only="!isTouchDevice"
-									:delay="isTouchDevice ? 0 : 1000"
+									:delay="isTouchDevice ? 300 : 1000"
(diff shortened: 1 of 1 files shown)


kolaente:
auggie review

PR #2289, opened 2026-02-24
fix: make all database write operations atomic via proper transaction handling

kolaente:
## Summary

- `db.NewSession()` now auto-begins a transaction, making all write operations atomic by default
- Added missing `Commit()` calls to ~18 write callers that previously had no-op commits in auto-commit mode
- Fixed session leaks and removed redundant `Begin()` calls
- Eliminated nested database sessions that caused "database table is locked" errors by threading sessions through `notifications.Notify()`, `files.CreateWithSession()`, `DeleteBackgroundFileIfExists()`, and `ShouldNotify()`

## Test plan

- [x] `mage build` compiles successfully
- [x] `mage test:feature` passes (all failures are pre-existing xorm "hash of unhashable type" panics unrelated to this change)
- [x] `mage test:web` passes (same pre-existing failure)
- [x] Regression tests added for atomic parent project deletion (archived children + deeply nested hierarchies)

### DIFF
--- a/pkg/models/sessions.go
+++ b/pkg/models/sessions.go
@@ -230,6 +230,10 @@ func RegisterSessionCleanupCron() {
 		if deleted > 0 {
 			log.Debugf(logPrefix+"Deleted %d stale sessions", deleted)
 		}
+
+		if err := s.Commit(); err != nil {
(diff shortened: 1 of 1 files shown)


kolaente:
auggie review

kolaente:
auggie review

PR #2290, opened 2026-02-24
feat: session invalidation with refresh tokens

kolaente:
## Summary

- Add a `sessions` table to track active user sessions with refresh tokens
- Replace long-lived JWTs with short-lived JWTs (~10 min) + rotating HttpOnly refresh token cookies
- Users can list and revoke active sessions from a new Settings page
- All sessions are invalidated on password change and password reset
- Existing `POST /user/token` endpoint is locked down to link share renewal only

## New API endpoints

| Method | Path | Auth | Description |
|--------|------|------|-------------|
| `POST` | `/api/v1/user/token/refresh` | Cookie | Exchange refresh token for new JWT + rotated cookie |
| `POST` | `/api/v1/user/logout` | JWT | Delete current session (server-side logout) |
| `GET` | `/api/v1/user/sessions` | JWT | List active sessions |
| `DELETE` | `/api/v1/user/sessions/:session` | JWT | Revoke a specific session |

## New config

- `service.jwtttlshort` — TTL for short-lived JWTs (default: 600 seconds / 10 min)

## Test plan

- [ ] Login creates a session and sets the refresh token cookie
- [ ] Refresh endpoint returns new JWT and rotates the cookie
- [ ] Expired/invalid refresh tokens are rejected with 401
- [ ] Session list returns only the authenticated user's sessions
- [ ] Deleting another user's session is forbidden
- [ ] Password change/reset invalidates all sessions
- [ ] `POST /user/token` rejects user tokens (link shares only)
- [ ] Frontend login/logout/refresh flow works end-to-end
- [ ] Sessions settings page lists sessions and allows revocation

### DIFF
--- a/frontend/tests/e2e/user/session-refresh.spec.ts
+++ b/frontend/tests/e2e/user/session-refresh.spec.ts
@@ -1,0 +1,134 @@
+import {test, expect} from '../../support/fixtures'
+import {UserFactory} from '../../factories/user'
+import {TEST_PASSWORD} from '../../support/constants'
+
+async function loginViaBrowser(page, username: string) {
+	// Set the API URL so the frontend knows where to send requests.
+	const apiUrl = process.env.API_URL || 'http://127.0.0.1:3456/api/v1'
+	await page.addInitScript(({apiUrl}) => {
+		window.localStorage.setItem('API_URL', apiUrl)
+		window.API_URL = apiUrl
+	}, {apiUrl})
+
+	await page.goto('/login')
+	await page.locator('input[id=username]').fill(username)
+	await page.locator('input[id=password]').fill(TEST_PASSWORD)
+	await page.locator('.button').filter({hasText: 'Login'}).click()
+	await expect(page).toHaveURL('/')
+	await expect(page.locator('main h2')).toContainText(username)
+
+	// Wait for the proactive refresh (from useRenewTokenOnFocus) to complete
+	// so it doesn't race with our test assertions.
+	await page.waitForTimeout(1500)
+}
+
+test.describe('Session refresh and retry interceptor', () => {
+	let username: string
+
+	test.beforeEach(async ({apiContext}) => {
+		const [user] = await UserFactory.create(1)
+		username = user.username
+	})
+
+	test('Transparently retries a request and rotates the JWT when a 401 with code 11 is returned', async ({page}) => {
+		await loginViaBrowser(page, username)
+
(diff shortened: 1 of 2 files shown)


kolaente:
auggie review

pkg/models/sessions.go
```diff
@@ -0,0 +1,223 @@
...
+// RotateRefreshToken atomically replaces the session's refresh token hash.
+// The WHERE clause includes the old hash so that concurrent refreshes with the
+// same token cannot both succeed — only the first UPDATE matches a row; the
+// second sees 0 affected rows and returns ErrSessionNotFound.
+func RotateRefreshToken(s *xorm.Session, session *Session) (newRawToken string, err error) {
+	tokenBytes, err := utils.CryptoRandomBytes(128)
+	if err != nil {
+		return "", err
+	}
+	newRawToken = hex.EncodeToString(tokenBytes)
```
kolaente: this should be a new internal function `generateHashedToken` that's used everywhere so that we don't forget to implement the same thing multiple times

pkg/models/sessions_permissions.go
```diff
@@ -0,0 +1,38 @@
...
+	"code.vikunja.io/api/pkg/web"
+	"xorm.io/xorm"
+)
+
+func (sess *Session) CanDelete(s *xorm.Session, a web.Auth) (bool, error) {
+	session, err := GetSessionByID(s, sess.ID)
+	if err != nil {
+		return false, err
+	}
+	if session.UserID != a.GetID() {
```
kolaente, line 34: This should also check if the auth session is a user, otherwise this will allow deleting user sessions with the right link share token

kolaente:
auggie review

frontend/src/stores/auth.ts
```diff
@@ -297,15 +297,40 @@ export const useAuthStore = defineStore('auth', () => {
...
+					// user with the same ID. The JWT lacks fields like `name`, so
+					// overwriting a complete user object causes a visible flash
+					// where the display name briefly reverts to the username.
+					if (info.value === null || info.value.id !== jwtUser.id) {
+						setUser(jwtUser, false)
+					} else {
+						// Always keep exp in sync so token renewal checks stay accurate
+						info.value.exp = jwtUser.exp
+					}
+				} else if (jwtUser.type === AUTH_TYPES.USER) {
```
kolaente, line 312: maybe we can do an early reaturn here

kolaente (commented):
r

kolaente:
auggie review

PR #2301, opened 2026-02-25
feat: enforce password validation on reset and update flows

kolaente:
## Summary

Ensures password validation rules (min 8 chars, max 72 bytes for bcrypt) are enforced consistently across all password-related flows.

### Backend
- Added bcrypt_password validation tag to PasswordReset struct
- Added bcrypt_password validation tag to UserPassword struct  
- Added c.Validate() calls in password reset and password update handlers
- Added tests for password-too-short validation in both flows

### Frontend
- Updated PasswordUpdate.vue to use the Password component (with built-in validation) instead of a plain FormField

## Testing
- mage test:filter TestUserPasswordReset - passes
- mage test:filter TestUserChangePassword - passes
- mage lint - passes
- pnpm lint - passes

### DIFF
--- a/frontend/src/views/user/settings/PasswordUpdate.vue
+++ b/frontend/src/views/user/settings/PasswordUpdate.vue
@@ -5,24 +5,17 @@
 		:loading="passwordUpdateService.loading"
 	>
 		<form @submit.prevent="updatePassword">
-			<FormField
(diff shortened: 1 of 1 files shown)


kolaente:
auggie review

PR #2306, opened 2026-02-26
fix: collapse view buttons into dropdown when overflowing

maggch97:
**Problem**

When a project has many views, the view switcher buttons overflow the screen horizontally, making some views inaccessible.

**Solution**

  Automatically detect when the view buttons would overflow the container and collapse them into a dropdown menu.

   - When buttons fit within the container → show inline buttons as before (no change in UX)
   - When buttons overflow → show a dropdown displaying the current view name with a chevron, clicking it reveals all views in a dropdown menu

Recording 2026-02-26 at 15 47 39

### DIFF
--- a/frontend/src/components/project/ProjectWrapper.vue
+++ b/frontend/src/components/project/ProjectWrapper.vue
@@ -238,6 +238,7 @@ function getViewRoute(view: IProjectView) {
 .switch-view-button {
 	padding: .25rem .5rem;
 	display: block;
+	white-space: nowrap;
 	border-radius: $radius;
 	transition: all 100ms;
 


frontend/src/components/project/ProjectWrapper.vue
```diff
@@ -11,11 +11,56 @@
...
 
 		<div
+			ref="switchViewContainerRef"
 			class="switch-view-container d-print-none"
 			:class="{'is-justify-content-flex-end': views.length === 1}"
 		>
+			<!-- Hidden measurement element to detect total button width -->
 			<div
 				v-if="views.length > 1"
+				ref="measureRef"
```
kolaente: Why the extra element?

kolaente:
Should it do this?

Screen Shot 2026-03-04 at 23 51 41

maggch97:
Fixed — added `white-space: nowrap` to view buttons so they no longer wrap text. When there's not enough space, the dropdown menu is shown instead.

PR #2308, opened 2026-02-26
feat: add OAuth 2.0 PKCE authorization provider

kolaente:
## Summary

This PR implements an OAuth 2.0 authorization server to enable the Flutter app to authenticate using the authorization code flow with PKCE.

### Backend
- OAuthCode model (pkg/models/oauth_codes.go) - Stores authorization codes with PKCE challenge, user ID, and expiry
- Database migration (pkg/migration/20260226172819.go) - Creates oauth_codes table
- OAuth error types (pkg/models/error.go) - Added 7 error types (17001-17007) for OAuth-specific errors
- OAuth2 server package (pkg/modules/auth/oauth2server/):
  - client.go - Client validation (Flutter client ID: vikunja-flutter, redirect URI: vikunja://callback)
  - pkce.go - PKCE S256 verification
  - authorize.go - Authorization endpoint handler (GET /api/v1/oauth/authorize)
  - token.go - Token endpoint handler (POST /api/v1/oauth/token) with authorization_code and refresh_token grant types
- Route registration (pkg/routes/routes.go) - Registered OAuth endpoints

### Frontend
- Login.vue and OpenIdAuth.vue - Handle OAuth redirect parameter after successful login

### Tests
- Unit tests for client validation and PKCE verification
- Web tests (pkg/webtests/oauth2_test.go) - 11 integration test cases covering the full OAuth flow

## OAuth Flow

1. Flutter app opens /api/v1/oauth/authorize with client_id, redirect_uri, response_type, code_challenge, code_challenge_method
2. If not logged in, redirects to login page with return URL
3. After login, redirects back to authorize endpoint
4. Authorize endpoint creates code and redirects to vikunja://callback with code
5. Flutter app exchanges code for tokens via POST /api/v1/oauth/token
6. Refresh tokens can be exchanged for new access/refresh token pairs

## Security Features

- PKCE (S256) required for all authorization requests
- Single-use authorization codes (deleted on use)
- 10-minute code expiry
- Refresh token rotation
- Same-origin validation for OAuth redirects in frontend

### DIFF
--- a/frontend/tests/e2e/user/oauth-authorize.spec.ts
+++ b/frontend/tests/e2e/user/oauth-authorize.spec.ts
@@ -1,0 +1,80 @@
+import {createHash, randomBytes} from 'crypto'
+import {test, expect} from '../../support/fixtures'
+import {UserFactory} from '../../factories/user'
+import {setupApiUrl} from '../../support/authenticateUser'
+import {TEST_PASSWORD} from '../../support/constants'
+
+test.describe('OAuth 2.0 Authorization Flow', () => {
+	let username: string
+
+	test.beforeEach(async ({apiContext}) => {
+		const [user] = await UserFactory.create(1)
+		username = user.username
+	})
+
+	test('Full browser authorization code flow with PKCE', async ({page, apiContext}) => {
+		await setupApiUrl(page)
+
+		// Generate PKCE code_verifier and code_challenge (S256)
+		const codeVerifier = randomBytes(32).toString('base64url')
+		const codeChallenge = createHash('sha256').update(codeVerifier).digest('base64url')
+		const state = randomBytes(16).toString('base64url')
+
+		// Build the authorize URL as a frontend route with OAuth query params.
+		// The OAuthAuthorize.vue component reads these and POSTs to the API.
+		const authorizeParams = new URLSearchParams({
+			response_type: 'code',
+			client_id: 'vikunja',
+			redirect_uri: 'vikunja-flutter://callback',
+			code_challenge: codeChallenge,
+			code_challenge_method: 'S256',
+			state,
+		})
+
+		// Navigate to the OAuth authorize frontend route.
+		// The user is not logged in, so the router guard saves the route
+		// and redirects to /login.
+		await page.goto(`/oauth/authorize?${authorizeParams}`)
+		await expect(page).toHaveURL(/\/login/)
+
+		// Register the response listener BEFORE clicking Login, because after
+		// login redirectIfSaved() navigates back to /oauth/authorize and the
+		// component immediately POSTs to the API.
+		const authorizeResponsePromise = page.waitForResponse(
+			response => response.url().includes('/api/v1/oauth/authorize') && response.request().method() === 'POST',
+			{timeout: 15000},
+		)
+
+		// Log in via the browser UI
+		await page.locator('input[id=username]').fill(username)
+		await page.locator('input[id=password]').fill(TEST_PASSWORD)
+		await page.locator('.button').filter({hasText: 'Login'}).click()
+
+		// Wait for the authorize API call that fires after login redirect
+		const authorizeResponse = await authorizeResponsePromise
+		const authorizeBody = await authorizeResponse.json()
+		expect(authorizeBody.code).toBeTruthy()
+		expect(authorizeBody.redirect_uri).toBe('vikunja-flutter://callback')
+		expect(authorizeBody.state).toBe(state)
+
+		const code = authorizeBody.code
+
+		// Exchange the authorization code for tokens
+		const tokenResponse = await apiContext.post('oauth/token', {
+			data: {
+				grant_type: 'authorization_code',
+				code,
+				client_id: 'vikunja',
+				redirect_uri: 'vikunja-flutter://callback',
+				code_verifier: codeVerifier,
+			},
+		})
+
+		expect(tokenResponse.ok()).toBe(true)
+		const tokenBody = await tokenResponse.json()
+		expect(tokenBody.access_token).toBeTruthy()
+		expect(tokenBody.refresh_token).toBeTruthy()
+		expect(tokenBody.token_type).toBe('bearer')
+		expect(tokenBody.expires_in).toBeGreaterThan(0)
+	})
+})
--- a/pkg/modules/auth/oauth2server/client_test.go
+++ b/pkg/modules/auth/oauth2server/client_test.go
@@ -1,0 +1,50 @@
+// Vikunja is a to-do list application to facilitate your life.
+// Copyright 2018-present Vikunja and contributors. All rights reserved.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package oauth2server
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidateRedirectURI(t *testing.T) {
+	t.Run("accepts vikunja-flutter scheme", func(t *testing.T) {
+		assert.True(t, ValidateRedirectURI("vikunja-flutter://callback"))
+	})
+	t.Run("accepts vikunja-desktop scheme", func(t *testing.T) {
+		assert.True(t, ValidateRedirectURI("vikunja-desktop://auth"))
+	})
+	t.Run("rejects https scheme", func(t *testing.T) {
+		assert.False(t, ValidateRedirectURI("https://evil.com/callback"))
+	})
+	t.Run("rejects http scheme", func(t *testing.T) {
+		assert.False(t, ValidateRedirectURI("http://localhost/callback"))
+	})
+	t.Run("rejects javascript scheme", func(t *testing.T) {
+		assert.False(t, ValidateRedirectURI("javascript:alert(1)"))
+	})
+	t.Run("rejects data scheme", func(t *testing.T) {
+		assert.False(t, ValidateRedirectURI("data:text/html,<script>alert(1)</script>"))
+	})
+	t.Run("rejects non-vikunja custom scheme", func(t *testing.T) {
+		assert.False(t, ValidateRedirectURI("myapp://callback"))
+	})
+	t.Run("rejects empty URI", func(t *testing.T) {
+		assert.False(t, ValidateRedirectURI(""))
+	})
+}
--- a/pkg/webtests/oauth2_test.go
+++ b/pkg/webtests/oauth2_test.go
@@ -1,0 +1,308 @@
+// Vikunja is a to-do list application to facilitate your life.
+// Copyright 2018-present Vikunja and contributors. All rights reserved.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package webtests
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"code.vikunja.io/api/pkg/modules/auth"
+	"code.vikunja.io/api/pkg/modules/auth/oauth2server"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// authorizeRequestBody builds a JSON body for the authorize endpoint.
+func authorizeRequestBody(responseType, clientID, redirectURI, codeChallenge, codeChallengeMethod, state string) []byte {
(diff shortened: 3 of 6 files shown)


frontend/src/views/user/Login.vue
```diff
@@ -217,6 +218,23 @@ async function submit() {
 	try {
 		await authStore.login(credentials)
 		authStore.setNeedsTotpPasscode(false)
+
+		// Check for OAuth redirect parameter (used when returning from OAuth authorize)
+		const redirectParam = route.query.redirect as string
```
kolaente: This is the exact same code that's used in frontend/src/views/user/OpenIdAuth.vue - can we refactor this so that it uses a function or similar?

pkg/modules/auth/auth.go
```diff
@@ -47,12 +47,13 @@ type Token struct {
 	Token string `json:"token" example:"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"`
 }
 
-const RefreshTokenCookieName = "vikunja_refresh_token"      //nolint:gosec // not a credential
-const refreshTokenCookiePath = "/api/v1/user/token/refresh" //nolint:gosec // not a credential
```
kolaente, line 51: Why was this changed?

pkg/modules/auth/oauth2server/client.go
```diff
@@ -0,0 +1,44 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package oauth2server
+
+// FlutterClientID is the hard-coded client ID for the official Flutter app.
+const FlutterClientID = "vikunja-flutter"
```
kolaente: We should make this client-agnostic, as long as it redirects to a custom protocol no need to do anything here. Maybe rename to just `vikunja`?

pkg/modules/auth/oauth2server/pkce.go
```diff
@@ -0,0 +1,34 @@
...
+package oauth2server
+
+import (
+	"crypto/sha256"
+	"encoding/base64"
+)
+
+// VerifyPKCE verifies a PKCE code_verifier against a stored code_challenge using S256.
+// S256: BASE64URL(SHA256(code_verifier)) == code_challenge
+func VerifyPKCE(codeVerifier, codeChallenge, codeChallengeMethod string) bool {
```
kolaente, line 26: How long are these stored? Do we need to use a salt?

pkg/modules/auth/oauth2server/client.go
kolaente: What are the security implications of accepting any client id vs having a single hardcoded client id that is public? (the source code is public)

frontend/src/views/user/OAuthAuthorize.vue
```diff
@@ -0,0 +1,59 @@
...
+async function authorize() {
+	try {
+		const HTTP = AuthenticatedHTTPFactory()
+		const response = await HTTP.post('oauth/authorize', {
+			response_type: route.query.response_type,
+			client_id: route.query.client_id,
+			redirect_uri: route.query.redirect_uri,
+			state: route.query.state,
+			code_challenge: route.query.code_challenge,
+			code_challenge_method: route.query.code_challenge_method,
```
kolaente, line 57: We should validate in the frontend if all required query parameters are passed and show a proper error message if not

frontend/tests/e2e/user/oauth-authorize.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+		username = user.username
+	})
+
+	test('Full browser authorization code flow with PKCE', async ({page, apiContext}) => {
+		const apiUrl = (process.env.API_URL || 'http://127.0.0.1:3456/api/v1').replace(/\/+$/, '')
+
+		// Set the API URL so the frontend knows where to send requests
+		await page.addInitScript(({apiUrl}) => {
+			window.localStorage.setItem('API_URL', apiUrl)
+			window.API_URL = apiUrl
```
kolaente: Doesn't this already happen in the test setup?

frontend/tests/e2e/user/oauth-authorize.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+		// The user is not logged in, so the router guard saves the route
+		// and redirects to /login.
+		await page.goto(`/oauth/authorize?${authorizeParams}`)
+		await expect(page).toHaveURL(/\/login/)
+
+		// Register the response listener BEFORE clicking Login, because after
+		// login redirectIfSaved() navigates back to /oauth/authorize and the
+		// component immediately POSTs to the API.
+		const authorizeResponsePromise = page.waitForResponse(
+			response => response.url().includes('/oauth/authorize') && response.request().method() === 'POST',
```
kolaente: Should be 
```suggestion
			response => response.url().includes('/api/v1/oauth/authorize') && response.request().method() === 'POST',
```

pkg/modules/auth/oauth2server/token.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// tokenRequest holds the form-decoded body of a POST /oauth/token request.
+type tokenRequest struct {
+	GrantType    string `json:"grant_type" form:"grant_type"`
+	Code         string `json:"code" form:"code"`
+	ClientID     string `json:"client_id" form:"client_id"`
+	RedirectURI  string `json:"redirect_uri" form:"redirect_uri"`
+	CodeVerifier string `json:"code_verifier" form:"code_verifier"`
+	RefreshToken string `json:"refresh_token" form:"refresh_token"`
```
kolaente: This should only accept `json`, no form

pkg/models/oauth_codes.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"time"
+
+	"code.vikunja.io/api/pkg/utils"
+
+	"xorm.io/xorm"
+)
+
+// OAuthCode represents a short-lived OAuth 2.0 authorization code.
+type OAuthCode struct {
```
kolaente, line 28: Why is this in the `models` package?

pkg/modules/auth/oauth2server/client.go
```diff
@@ -0,0 +1,45 @@
...
+package oauth2server
+
+import "net/url"
+
+// ClientID is the client ID for Vikunja native apps.
+const ClientID = "vikunja"
+
+// ValidateClient checks that the client_id is recognized.
+func ValidateClient(clientID string) bool {
+	return clientID == ClientID
```
kolaente: Do we even have to do this? Can't we just accept any client id?

pkg/modules/auth/oauth2server/client_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidateClient(t *testing.T) {
+	assert.True(t, ValidateClient("vikunja"))
+	assert.False(t, ValidateClient("unknown-client"))
+	assert.False(t, ValidateClient(""))
+}
+
+func TestValidateRedirectURI(t *testing.T) {
```
kolaente, line 25: Each of these cases should be a sub test with t.Run

pkg/modules/auth/oauth2server/authorize.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"code.vikunja.io/api/pkg/db"
+	"code.vikunja.io/api/pkg/models"
+	"code.vikunja.io/api/pkg/user"
+
+	"github.com/labstack/echo/v5"
+)
+
+// AuthorizeRequest represents the JSON body for the authorize endpoint.
+type AuthorizeRequest struct {
```
kolaente: Why is this public but the `tokenRequest` is not?

pkg/modules/auth/oauth2server/authorize.go
```diff
@@ -0,0 +1,110 @@
...
+		return &models.ErrOAuthClientNotFound{}
+	}
+
+	// Validate redirect_uri
+	if !ValidateRedirectURI(req.RedirectURI) {
+		return &models.ErrOAuthInvalidRedirectURI{}
+	}
+
+	// Validate PKCE (required)
+	if req.CodeChallenge == "" || req.CodeChallengeMethod != "S256" {
```
kolaente, line 66: If we're only supporting S256 we should validate that and return a proper error

pkg/modules/auth/oauth2server/authorize.go
```diff
@@ -0,0 +1,110 @@
...
+	u, err := user.GetCurrentUser(c)
+	if err != nil {
+		return err
+	}
+
+	s := db.NewSession()
+	defer s.Close()
+
+	// Verify the user account is still active
+	fullUser, err := user.GetUserByID(s, u.ID)
```
kolaente, line 79: Doesn't the `getUser` function that's called there underneath validate?

pkg/modules/auth/oauth2server/token.go
```diff
@@ -0,0 +1,222 @@
...
+func HandleToken(c *echo.Context) error {
+	var req tokenRequest
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "Invalid request body")
+	}
+
+	switch req.GrantType {
+	case "authorization_code":
+		return handleAuthorizationCodeGrant(c, &req)
+	case "refresh_token":
```
kolaente, line 60: Do we need to handle token refrech here? How does that relate to the existing refresh mechanism we already have?

pkg/modules/auth/oauth2server/token.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Get the user for JWT claims
+	u, err := user.GetUserByID(s, oauthCode.UserID)
+	if err != nil {
+		_ = s.Rollback()
+		return err
+	}
+
+	// Check account status
+	if u.Status == user.StatusDisabled {
```
kolaente: Again, do we need to check that here or does the user `GetUserByID` already check it?

And this does only check Disabled, there's Locked as well

Also checked in other places

pkg/webtests/oauth2_test.go
```diff
@@ -0,0 +1,351 @@
...
+	})
+	return body
+}
+
+// doAuthorize performs a POST to /api/v1/oauth/authorize with the given JWT token and returns the recorder.
+func doAuthorize(e http.Handler, token string, body []byte) *httptest.ResponseRecorder {
+	rec := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodPost, "/api/v1/oauth/authorize", bytes.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "Bearer "+token)
```
kolaente, line 53: Why does this use a Bearer token?

frontend/tests/e2e/user/oauth-authorize.spec.ts
kolaente: Okay, but that sounds like a workaround. Please refactor the setup so that we have a reusable part that both use

pkg/models/oauth_codes.go
kolaente: all right

pkg/modules/auth/oauth2server/authorize.go
kolaente: fine as is, seems like I misread the erorr

pkg/modules/auth/oauth2server/authorize.go
```diff
@@ -87,7 +82,7 @@ func HandleAuthorize(c *echo.Context) error {
 		_ = s.Rollback()
 		return err
 	}
-	if fullUser.Status == user.StatusDisabled {
+	if fullUser.Status == user.StatusDisabled || fullUser.Status == user.StatusAccountLocked {
```
kolaente: This is checked in multiple places, extract to a shared helper.

or better: create a helper in the user package that fetches the user and checks its status and returns an error for non-active users, so that the same checks are used everywhere - basically exposing what getUser already does

kolaente: Great, but there are many redundant comments now, fix those

kolaente:
auggie review

kolaente:
@codex review

pkg/modules/auth/oauth2server/client.go
```diff
@@ -0,0 +1,37 @@
...
+// handled by native apps, not by browsers, preventing open redirect attacks.
+// Standard web schemes (http/https) are rejected.
+func ValidateRedirectURI(redirectURI string) bool {
+	u, err := url.Parse(redirectURI)
+	if err != nil || u.Scheme == "" {
+		return false
+	}
+
+	// Reject standard web schemes — only allow custom protocols
+	if u.Scheme == "http" || u.Scheme == "https" {
```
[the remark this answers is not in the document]
kolaente: Let's make this an allowlist of everything starting with a `vikunja-` prefix for now

pkg/modules/auth/oauth2server/authorize.go
```diff
@@ -0,0 +1,100 @@
...
+	s := db.NewSession()
+	defer s.Close()
+
+	fullUser, err := user.GetUserByID(s, u.ID)
+	if err != nil {
+		_ = s.Rollback()
+		return err
+	}
+
+	code, err := models.CreateOAuthCode(s, fullUser.ID, req.ClientID, req.RedirectURI, req.CodeChallenge, req.CodeChallengeMethod)
```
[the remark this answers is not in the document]
kolaente: This is fine since there is no way to register clients.

pkg/modules/auth/oauth2server/token.go
kolaente: Is this duplicated with the code path from the frontend refresh?

pkg/modules/auth/auth.go
```diff
@@ -216,3 +216,100 @@ func CreateUserWithRandomUsername(s *xorm.Session, uu *user.User) (u *user.User,
...
+	if err != nil {
+		_ = s.Rollback()
+		if models.IsErrSessionNotFound(err) {
+			return nil, echo.NewHTTPError(http.StatusUnauthorized, "Refresh token already used.")
+		}
+		return nil, err
+	}
+
+	var u *user.User
+	if withEmail {
```
kolaente: Why do we need this gate?

PR #2313, opened 2026-02-27
fix(deps): use forked afero-s3 to fix S3 read performance regression

maggch97:
The upstream afero-s3 library has a severe performance regression where every Read() call opens a new HTTP connection to S3, resulting in hundreds of requests for sequential reads (e.g. ~320 requests for a 10MB file with 32KB buffers instead of 1).

This temporarily replaces afero-s3 with a fork that fixes the issue by requesting from the current offset to end-of-file when opening a read stream, and properly closing the stream on Seek().

Upstream PR: https://github.com/fclairamb/afero-s3/pull/938 This replace directive should be removed once the upstream PR is merged and a new version is released.

all releases after Jan 15(https://github.com/go-vikunja/vikunja/commit/843ef7218a141ee23d25f99f46243bd4598f59fa ) are affected


maggch97:
I ran into this issue after upgrading to 1.1, and I’ve now rolled back to the November version. However, I’m seeing notifications does not work properly after rolling back, and I’m not sure whether the downgrade introduced any other issues.

Given that this package is currently making the S3 filesystem completely unusable — a single avatar request takes more than 10 seconds — could we merge this PR first and wait for the upstream fix afterward?


@kolaente

kolaente:
Okay, but do we actually need the afero-s3 package? As far as I see it, we could think about just using the sdk directly since we don't use that many S3 apis.

Then again, I've only lightly used S3 so I'm not sure if I know enough about this.

maggch97:
> Okay, but do we actually need the afero-s3 package? As far as I see it, we could think about just using the sdk directly since we don't use that many S3 apis.
> 
> Then again, I've only lightly used S3 so I'm not sure if I know enough about this.

Agree with you, consider to S3 is an easy use protocol, we can maintain the wrapper

I can spend more time on this later, one thing that can benefit from this is our warpper can respond with s3 presigned link, then data does not need to go through vikunja.

maggch97:
@kolaente Before the deprecation of afero-s3, I suggest to merge this change first, current S3 filesystem in latest Vikunja is completely unusable.

PR #2320, opened 2026-02-28
fix: use MinPositionSpacing threshold in calculateNewPositionForTask

maggch97:
calculateNewPositionForTask only checked for lowestPosition == 0 before triggering a full position recalculation. Extremely small position values (e.g. 3.16e-285) passed this check, causing new tasks to get meaningless positions via the index * 2^16 fallback, breaking sort order.

Use the same < MinPositionSpacing threshold that createPositionsForTasksInView and the position update handler already use.


maggch97:
image

All tasks in my project are generated by API call, after 2000 tasks generated, all new tasks are moved to the end of list view.  I remember there was a position 3.16e-285 in the first page, but I'm not sure if it's the first one.

kolaente:
Thanks!

PR #2322, opened 2026-02-28
Adding TLS options to Redis & mailer connections

### LINKED ISSUE #2321: feat: Add TLS options to the Redis connection
### Pre-submission checklist

### Description

The underlying go-redis library is able to do TLS, including client certificates for mutual TLS. It would be good to have this option. A pull request is incoming for review.

N/A

### LINKED ISSUE #2323: feat: Add TLS client certificate options to the mailer connection
### Pre-submission checklist

### Description

The underlying go-mail library is able to do TLS client certificates for mutual TLS. It would be good to have this option. A pull request is incoming for review.

_No response_

rhclayto:
Resolves https://github.com/go-vikunja/vikunja/issues/2321
Resolves https://github.com/go-vikunja/vikunja/issues/2323

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -117,17 +117,20 @@ const (
 	DatabaseTLS                   Key = `database.tls`
 	DatabaseSchema                Key = `database.schema`
 
-	MailerEnabled       Key = `mailer.enabled`
-	MailerHost          Key = `mailer.host`
-	MailerPort          Key = `mailer.port`
-	MailerUsername      Key = `mailer.username`
-	MailerPassword      Key = `mailer.password`
-	MailerAuthType      Key = `mailer.authtype`
-	MailerSkipTLSVerify Key = `mailer.skiptlsverify`
-	MailerFromEmail     Key = `mailer.fromemail`
-	MailerQueuelength   Key = `mailer.queuelength`
-	MailerQueueTimeout  Key = `mailer.queuetimeout`
-	MailerForceSSL      Key = `mailer.forcessl`
+	MailerEnabled         Key = `mailer.enabled`
+	MailerHost            Key = `mailer.host`
+	MailerPort            Key = `mailer.port`
+	MailerUsername        Key = `mailer.username`
+	MailerPassword        Key = `mailer.password`
+	MailerAuthType        Key = `mailer.authtype`
+	MailerSkipTLSVerify   Key = `mailer.skiptlsverify`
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -117,22 +117,30 @@ const (
...
+	MailerPort            Key = `mailer.port`
+	MailerUsername        Key = `mailer.username`
+	MailerPassword        Key = `mailer.password`
+	MailerAuthType        Key = `mailer.authtype`
+	MailerSkipTLSVerify   Key = `mailer.skiptlsverify`
+	MailerFromEmail       Key = `mailer.fromemail`
+	MailerQueuelength     Key = `mailer.queuelength`
+	MailerQueueTimeout    Key = `mailer.queuetimeout`
+	MailerForceSSL        Key = `mailer.forcessl`
+	MailerTLSClientCert   Key = `mailer.tlsclientcert`
```
kolaente: This (and the other new variables) should end on `path` to make it clear that these expect file paths to certs, not actual certificate strings.

kolaente (changes requested):
Thanks for the PR! Please see my suggestions below.

pkg/mail/mail.go
```diff
@@ -47,13 +49,33 @@ func getClient() (*mail.Client, error) {
 		tlsPolicy = mail.TLSMandatory
 	}
 
+	tlsConfig := &tls.Config{
+		//#nosec G402
+		InsecureSkipVerify: config.MailerSkipTLSVerify.GetBool(),
+		ServerName:         config.MailerHost.GetString(),
+		MinVersion:         tls.VersionTLS12,
```
kolaente, line 56: Why this change?

This is a breaking change, we can't just add it like this.

pkg/red/redis.go
```diff
@@ -40,11 +43,39 @@ func InitRedis() {
...
-	r = redis.NewClient(&redis.Options{
+	redisOpts := &redis.Options{
 		Addr:     config.RedisHost.GetString(),
 		Password: config.RedisPassword.GetString(),
 		DB:       config.RedisDB.GetInt(),
-	})
+	}
+
+	if config.RedisTLS.GetBool() {
+		tlsConfig := &tls.Config{
```
kolaente, line 53: It seems like the two ways the TLS config is set up is pretty much the same for redis and mail, just the inputs are different. Can you make that a helper function?

kolaente (changes requested):

PR #2325, opened 2026-03-01
feat: add task creation to table view

maggch97:
## Summary

Users who primarily use the table view for task management currently have no way to create new tasks without switching to another view.

This adds the `AddTask` component (with Quick Add Magic support) to the table view, matching the list view's behavior. After creating a task, the user is navigated to the task detail page.

## Changes

- Add `AddTask` component to `ProjectTable.vue`
- Navigate to task detail page after creation so the user can immediately edit the new task
- Only shown when user has write permission

### DIFF
--- a/frontend/src/components/project/views/ProjectTable.vue
+++ b/frontend/src/components/project/views/ProjectTable.vue
@@ -305,6 +305,7 @@
 
 <script setup lang="ts">
 import {computed, type Ref, watch} from 'vue'
+import {useRouter} from 'vue-router'
 
 import {useStorage} from '@vueuse/core'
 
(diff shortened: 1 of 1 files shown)


kolaente:
What does this look like?

maggch97:
> What does this look like?

image

kolaente:
Unsure about this. The task input field matches for a list, but not so much for the table view.

To create tasks quickly, I think we should have some kind of wizard ultimately that allows users to create the task with more questions.

I think as a first step, we could add a button at the top (next to columns) that just triggers the quick actions with the create task action and the current project pre-selected?

maggch97:
> Unsure about this. The task input field matches for a list, but not so much for the table view.
> 
> To create tasks quickly, I think we should have some kind of wizard ultimately that allows users to create the task with more questions.
> 
> I think as a first step, we could add a button at the top (next to columns) that just triggers the quick actions with the create task action and the current project pre-selected?

Ok, I will add a 'new task' button that be shown in each view to provide a consistent experience

kolaente:
Closing this in favor of https://github.com/go-vikunja/vikunja/pull/2350

PR #2326, opened 2026-03-01
feat: add 'only my tasks' filter on homepage

maggch97:
## Summary

In multi-user teams, the homepage's Current Tasks section shows all undone tasks across all projects, making it hard for users to find their own assignments.
image

This adds a checkbox next to the Current Tasks title: **"Only show tasks assigned to me"**. When enabled, it filters tasks by `assignees = currentUser`.

The setting is persisted in `frontendSettings` so it syncs across devices.

## Changes

- `IUserSettings.ts`: Add `showOnlyMyTasks` to `IFrontendSettings`
- `userSettings.ts`: Default value `false`
- `ShowTasks.vue`: Add checkbox, save to frontendSettings on toggle, add assignee filter to task query
- `en.json`: Add i18n text

### DIFF
--- a/frontend/src/i18n/lang/en.json
+++ b/frontend/src/i18n/lang/en.json
@@ -833,7 +833,8 @@
       "select": "Select a date range",
       "noTasks": "Nothing to do — Have a nice day!",
       "filterByLabel": "Filtering by label {label}",
-      "clearLabelFilter": "Clear label filter"
+      "clearLabelFilter": "Clear label filter",
+      "onlyMyTasks": "Only show tasks assigned to me"
     },
     "detail": {
       "chooseDueDate": "Click here to set a due date",
--- a/frontend/src/modelTypes/IUserSettings.ts
+++ b/frontend/src/modelTypes/IUserSettings.ts
@@ -25,6 +25,7 @@ export interface IFrontendSettings {
 	alwaysShowBucketTaskCount: boolean
 	sidebarWidth: number | null
 	commentSortOrder: 'asc' | 'desc'
+	showOnlyMyTasks: boolean
 }
 
 export interface IExtraSettingsLink {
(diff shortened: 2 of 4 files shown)


kolaente:
Why is this better than creating a saved filter and using that for what is shown?

(not trying to be mean here, just to see if we can keep the scope low)

maggch97:
> Why is this better than creating a saved filter and using that for what is shown?
> 
> (not trying to be mean here, just to see if we can keep the scope low)

From my perspective, current behavior of Overview and Upcoming is designed for person or 2~3 people team. When there're 10 people in a same project, most tasks shown in Overview and Upcoming are not related to the current user. This makes Overview and Upcoming meaningless.

Saved filter can help on this and it is also the way I'm currently using, but
- Each people in the project needs at least 1 view, the view is only meaningful for one user but is displayed for all users
- I need to open each project to see all the tasks assigned to me; there's no aggregated view.

'only my tasks' filter is a simple change to make Overview and Upcoming useful again for team users.

If we want to go further, next step could be making the checkbox to a dropdown menu:
- Assigned to me
- Mentioned
- My activity

kolaente:
> I need to open each project to see all the tasks assigned to me; there's no aggregated view.

But this is exactly what the saved filter is solving? I mean the general saved filter (projects -> create saved filter), not per view.

maggch97:
> > I need to open each project to see all the tasks assigned to me; there's no aggregated view.
> 
> 
> 
> But this is exactly what the saved filter is solving? I mean the general saved filter (projects -> create saved filter), not per view.

Oh, yes that meets my requirements. I didn't know about it. But I think a checkbox or a dropdown menu is still a convenient way for most users and makes the Overview entrypoint more useful for team users.

Since the saved filter solves my issue, I’m happy to leave it up to you whether this PR should be merged.

kolaente:
Closing this as solved then, I'd be open to add a user setting that enables this though.

PR #2333, opened 2026-03-02
fix: defer event dispatch until after transaction commit

### LINKED ISSUE #2315: Webhook payload contains stale task data due to reloadEventData reading uncommitted transaction
### Pre-submission checklist

### Description

I use the webhook to track each task's done status, after upgrading to 2.0.0, I noticed that the done status in webhook payload maybe wrong.

Below is anylsis from copilot
```
  reloadEventData() (introduced in #2135) re-reads entities from the database before sending
  webhook payloads. However, events.Dispatch() is called inside updateSingleTask() before the
  enclosing transaction is committed in the web handler (pkg/web/handler/update.go:86).

  Since WebhookListener.Handle() opens a new database session (db.NewSession()), it cannot see
  the uncommitted changes from the original transaction. This causes reloadTaskInEvent() to
  overwrite the correct event payload with stale data from the database.

  Steps to reproduce:

   1. Register a webhook for task.updated events
   2. Toggle a task's done status
   3. Observe that the webhook payload sometimes contains the old done value

  Root cause:

   updateSingleTask() — inside transaction s:
     1. s.Update(&ot)          // write done=true, NOT committed
     2. events.Dispatch(...)   // publish event (payload has done=true)
        └─ WebhookListener goroutine:
           3. s2 := db.NewSession()
           4. reloadTaskInEvent(s2) → ReadOne()  // reads done=false (old value)
     
   web handler:
     5. s.Commit()             // done=true finally visible
     ```

### Vikunja Version

latest

### Browser and version

_No response_

Please select

### Screenshots

_No response_

kolaente:
## Summary

Fixes #2315

Events (especially webhooks) were being dispatched before database transactions committed, causing webhook listeners that open new database sessions to read stale data.

### New Deferred Event API (pkg/events/events.go)
- `DispatchOnCommit(key, event)` - Stores an event to be dispatched later
- `DispatchPending(key)` - Dispatches all accumulated events for a key (call after s.Commit())
- `CleanupPending(key)` - Discards pending events (call on rollback paths)

### Updated CRUD Handlers
All generic CRUD handlers now call `events.DispatchPending(s)` after commit and `events.CleanupPending(s)` on rollback paths.

### Converted Model Methods
All `events.Dispatch()` calls in model methods were converted to `events.DispatchOnCommit(s, ...)`:
- Tasks and task sub-entities (assignees, comments, attachments, relations, kanban buckets)
- Projects and project sharing (users, teams)
- Teams and team members
- User creation
- Task positions

### Updated Non-Handler Call Sites
- CalDAV handlers
- Migration (create_from_structure.go)
- User export

### Test Updates
Tests that call model methods directly and assert event dispatch now call `events.DispatchPending(s)` before assertions.

## Testing
- All existing tests pass
- Lint passes with 0 issues

### DIFF
--- a/pkg/models/task_comments_test.go
+++ b/pkg/models/task_comments_test.go
@@ -45,6 +45,7 @@ func TestTaskComment_Create(t *testing.T) {
 		assert.Equal(t, int64(1), tc.Author.ID)
 		err = s.Commit()
 		require.NoError(t, err)
+		events.DispatchPending(s)
 		events.AssertDispatched(t, &TaskCommentCreatedEvent{})
 
 		db.AssertExists(t, "task_comments", map[string]interface{}{
--- a/pkg/models/tasks_test.go
+++ b/pkg/models/tasks_test.go
@@ -70,6 +70,7 @@ func TestTask_Create(t *testing.T) {
 			"bucket_id": 1,
 		}, false)
 
+		events.DispatchPending(s)
 		events.AssertDispatched(t, &TaskCreatedEvent{})
 	})
 	t.Run("with reminders", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


kolaente:
auggie review

PR #2339, opened 2026-03-03
fix(cli): make user deletion confirmation check Windows compatible

chau-intl:
The current implementation of the user confirmation when deleting a user using the CLI seems to favour Linux. On Windows the <ENTER> command adds "\r\n" to the user input while Linux only adds "\n".

I have added an extra check to the confirmation, but GO is a new language for me, so there is probably a much better way to do this.

### DIFF
--- a/pkg/cmd/user.go
+++ b/pkg/cmd/user.go
@@ -364,7 +364,7 @@ var userDeleteCmd = &cobra.Command{
 
 			// On Windows <ENTER> a newline is \r\n, while on Linux it is only \n.
 			text = strings.TrimRight(text, "\r\n")
-			
+
 			if text != "YES, I CONFIRM" {
 				log.Fatalf("invalid confirmation message")
 			}


pkg/cmd/user.go
```diff
@@ -361,7 +361,7 @@ var userDeleteCmd = &cobra.Command{
 			if err != nil {
 				log.Fatalf("could not read confirmation message: %s", err)
 			}
-			if text != "YES, I CONFIRM\n" {
+			if text != "YES, I CONFIRM\n" && text != "YES, I CONFIRM\r\n" {
```
kolaente: The better way would be to trim the string before checking: https://pkg.go.dev/strings#Trim

Then the check becomes `text != "YES, I CONFIRM"`

chau-intl: I have changed the check to use `TrimRight` to remove either or both `\r` and `\n`. Is that a better way to do it?

kolaente:
Please fix the lint issues. https://github.com/go-vikunja/vikunja/actions/runs/22658808490/job/65745885775?pr=2339

kolaente:
seems like the bot didn' want to fix at first, hence the comment. It should be fine now

PR #2348, opened 2026-03-03
feat: replace afero-s3 with minimal in-tree S3 implementation

### LINKED ISSUE #2347: Evaluate replacing afero-s3 with direct S3 SDK usage
## Context

In #2313 we merged a temporary fork of afero-s3 to fix a severe S3 read performance regression (every `Read()` call was opening a new HTTP connection, resulting in ~320 requests for a 10MB file).

As discussed in that PR, we should evaluate whether to keep using afero-s3 at all, or replace it with direct AWS S3 SDK usage. We don't use many S3 APIs, so the abstraction layer may not be worth the maintenance burden and exposure to upstream regressions like this one.

## Tasks

- [ ] Audit current afero-s3 usage in the codebase — which `afero.Fs` methods do we actually use?
- [ ] Evaluate effort to replace with direct `aws-sdk-go` (or v2) calls
- [ ] Remove the `replace` directive in `go.mod` once upstream merges the fix (https://github.com/fclairamb/afero-s3/pull/938), or proceed with full removal

cc @kolaente @maggch97

maggch97:
## Summary

Replaces the third-party `fclairamb/afero-s3` library (and its temporary fork) with a minimal in-tree `afero.Fs` implementation (~200 lines) that only supports the three S3 operations Vikunja actually uses:

- **Open** → `HeadObject` (existence check) + lazy `GetObject` on first `Read`
- **Remove** → `HeadObject` + `DeleteObject`
- **Stat** → `HeadObject`

All other `afero.Fs`/`afero.File` methods return `ErrS3NotSupported` since they are never called in the S3 code path (writes already use direct `PutObject`).

## Changes

1. **`pkg/files/s3fs.go`** (new) — Minimal S3 `afero.Fs` + `afero.File` + `os.FileInfo` implementation
2. **`pkg/files/filehandling.go`** — Replace `aferos3.NewFsFromClient()` with `newS3Fs()`
3. **`pkg/routes/api/v1/user_export.go`** — Add S3 branch for export download (`io.Copy` instead of `http.ServeContent` which requires `io.ReadSeeker`)
4. **`go.mod` / `go.sum`** — Remove `fclairamb/afero-s3` dependency and `replace` directive

## Why

- Removes dependency on a library that caused a severe S3 read performance regression (#2313)
- Removes the temporary fork `replace` directive
- The full afero-s3 library implements ~400 lines of S3 operations we never use (Create, Mkdir, Rename, Chmod, Write streams, Readdir, etc.)
- Our implementation is small enough to maintain in-tree with minimal effort

Closes #2347

### DIFF
--- a/pkg/files/s3fs.go
+++ b/pkg/files/s3fs.go
@@ -26,7 +26,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
-	"github.com/aws/smithy-go"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
 	"github.com/spf13/afero"
 )
 
@@ -74,10 +74,19 @@ func (f *s3Fs) Stat(name string) (os.FileInfo, error) {
 		return nil, s3ToPathError("stat", name, err)
 	}
 
+	var size int64
+	if head.ContentLength != nil {
+		size = *head.ContentLength
+	}
+	var modTime time.Time
+	if head.LastModified != nil {
+		modTime = *head.LastModified
+	}
+
 	return &s3FileInfo{
 		name:    path.Base(name),
-		size:    *head.ContentLength,
-		modTime: *head.LastModified,
+		size:    size,
+		modTime: modTime,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/routes/api/v1/user_export.go
```diff
@@ -147,6 +150,15 @@ func DownloadUserDataExport(c *echo.Context) error {
 		return err
 	}
 
+	if config.FilesType.GetString() == "s3" {
+		c.Response().Header().Set("Content-Disposition", "attachment; filename=\""+exportFile.Name+"\"")
+		c.Response().Header().Set("Content-Type", "application/zip")
+		c.Response().Header().Set("Content-Length", strconv.FormatUint(exportFile.Size, 10))
+		c.Response().Header().Set("Last-Modified", exportFile.Created.UTC().Format(http.TimeFormat))
+		_, err = io.Copy(c.Response(), exportFile.File)
+		return err
```
kolaente, line 159: This seems like a difffernt change, I think we're already doing a similar thing for attachments?

pkg/files/s3fs.go
```diff
@@ -0,0 +1,207 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+	"github.com/spf13/afero"
+)
+
+// s3Fs is a minimal afero.Fs implementation backed by S3.
+// It only supports Open (read), Remove, and Stat — the operations
```
kolaente, line 34: No create or write? how exactly does that work then to create upload attachments?

pkg/routes/api/v1/user_export.go
maggch97: Yes, for s3 filesystem, data export does not work without this change.

maggch97: It was introduced in the first S3 PR, which I only noticed this time.

pkg/files/s3fs.go
maggch97: That's handled by https://github.com/go-vikunja/vikunja/blob/c6f0d8babe6f36e6d25d22a932c9f0a075a5a359/pkg/files/files.go#L176

kolaente: Okay, but if we're bypassing the s3 fs there already, what advantage do we have to reimplement it? (albeit partially)

maggch97: Yes, I also thought about this before.
1. We still use afero fs for local file operation. If we want to use S3 directly, we need to deal with 2 filesystem in caller or create a new abstraction. If so, why don't we continue using the afero abstraction
2. The benfit to maintain the s3 fs wrapper is we can avoid some mistakes. For example, the previous lib supports seek operation, but if the code uses it in a wrong way, performance issues would arise. The new implementation is very simple, less than 100 lines of actual logic, only supports stream read.

kolaente:
Will be replaced by https://github.com/go-vikunja/vikunja/pull/2418

PR #2349, opened 2026-03-03
fix: allow browser caching for file downloads

maggch97:
## Problem

The global API middleware sets `Cache-Control: no-store` on all `/api/v1/` responses to prevent browsers from heuristically caching JSON data. However, this also applies to binary file downloads (project backgrounds, task attachments), causing browsers to re-download these files on every request — even though they have `Last-Modified` headers set.

## Fix

Override `Cache-Control` to `no-cache` in the file download endpoints for:
- **Project backgrounds** (`/api/v1/projects/{id}/background`)
- **Task attachments** (`/api/v1/tasks/{id}/attachments/{id}`)

`no-cache` allows the browser to store the file locally but requires revalidation via `If-Modified-Since` before using the cached copy. Since these files are immutable (updates create a new file with a new ID), the `Last-Modified` header already set by these endpoints is sufficient for correct revalidation.

Export downloads are intentionally left with `no-store` since they are one-time downloads.

### DIFF
--- a/pkg/modules/background/handler/background.go
+++ b/pkg/modules/background/handler/background.go
@@ -380,9 +380,21 @@ func GetProjectBackground(c *echo.Context) error {
 		return err
 	}
 
+	// Override the global no-store directive so browsers can cache background images.
+	// no-cache allows caching but requires revalidation via If-Modified-Since.
+	c.Response().Header().Set("Cache-Control", "no-cache")
+
 	// Set Last-Modified header if we have the file stat, so clients can decide whether to use cached files
 	if stat != nil {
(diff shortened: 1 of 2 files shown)


kolaente:
The cache prevention was added to fix this security issue: https://github.com/go-vikunja/vikunja/security/advisories/GHSA-7jp5-298q-jg98

Does this still not happen with the changes from this PR?

maggch97:
This PR doesn't affect the GHSA-7jp5-298q-jg98 fix. The XSS protection relies on `Content-Disposition: attachment` (which forces download instead of inline rendering) — that header is untouched. We only change `Cache-Control` from `no-store` to `no-cache`, which controls browser caching behavior, not how files are rendered.

PR #2350, opened 2026-03-04
feat: add unified New Task button with full detail view creation

maggch97:
## Summary

Following the discussion in #2325, this adds a unified "New Task" button across all project views (List, Table, Kanban, Gantt) with a full task detail creation experience.

Instead of a quick-add text input, clicking "New Task" opens the full task detail view in create mode — similar to Azure DevOps where you fill in all fields before saving.

### New Task Button
- Added in `ProjectWrapper.vue` (right side of header, after filter/columns buttons)
- Visible when user has write permission and project is not archived
- Route: `/projects/:projectId/tasks/new`

### Create Mode (`TaskDetailView`)
- Shows **"New"** as the task identifier
- Editable title (inline contenteditable, same as existing tasks)
- Full **TipTap rich text editor** for description
- **Priority, dates, color, percent done, reminders, repeat interval** — all editable locally, no API calls until Save
- **Assignees** — selected locally, assigned via API after creation
- **Save button** creates the task and redirects to the real task detail URL
- Hidden until after creation: labels, attachments, comments, reactions, related tasks, move project, subscriptions, favorites

### Technical Approach
- `provide/inject` for `isNewTask` flag (avoids prop drilling into Description, EditAssignees)
- `saveTask()` is a no-op in create mode — Vue `v-model` keeps local state in sync
- `createNewTask()` reads directly from `task.value` (avoids `klona` Date serialization issues)

### Bug Fixes (independent, first commit)
- **`objectToCamelCase`** was recursing into `Date` objects, destroying them into `{}`. Added the same Date skip guard that `objectToSnakeCase` already has.
- **`TaskService.processModel`** used `new Date(null).toISOString()` for `created`/`updated` which throws. Now uses the null-safe `parseDate` helper like all other date fields.

### E2E Tests
13 Playwright tests covering:
- Button visibility across views and read-only projects
- Navigation to new task page
- UI elements (identifier, project name, Save/Done buttons, comments)
- Task creation with title, priority, and due date

## Known Limitations
- **Attachments** are not available in create mode (API requires a real task ID). This also means **images cannot be inserted in the description** during creation.
- **Labels** require a real task ID for the label-task association API, so they are hidden in create mode.
- These features become available immediately after clicking Save and being redirected to the real task detail view.

Closes #2325

### DIFF
--- a/frontend/src/components/project/ProjectWrapper.vue
+++ b/frontend/src/components/project/ProjectWrapper.vue
@@ -32,6 +32,7 @@
 				class="is-flex is-align-items-center"
 				style="gap: .5rem"
 			>
+				<slot name="header" />
 				<XButton
 					v-if="canWrite && !currentProject?.isArchived"
 					icon="plus"
@@ -39,7 +40,6 @@
 				>
 					{{ $t('task.newTask') }}
 				</XButton>
-				<slot name="header" />
 			</div>
 		</div>
 		<CustomTransition name="fade">
--- a/frontend/src/components/project/ProjectWrapper.vue
+++ b/frontend/src/components/project/ProjectWrapper.vue
@@ -28,7 +28,19 @@
 					{{ getViewTitle(view) }}
 				</BaseButton>
 			</div>
-			<slot name="header" />
+			<div
+				class="is-flex is-align-items-center"
+				style="gap: .5rem"
+			>
+				<XButton
+					v-if="canWrite && !currentProject?.isArchived"
+					icon="plus"
+					:to="{name: 'task.new', params: {projectId: projectId}}"
+				>
+					{{ $t('task.newTask') }}
+				</XButton>
+				<slot name="header" />
+			</div>
 		</div>
 		<CustomTransition name="fade">
 			<Message
@@ -49,6 +61,7 @@ import {computed} from 'vue'
 import {useI18n} from 'vue-i18n'
 
 import BaseButton from '@/components/base/BaseButton.vue'
+import XButton from '@/components/input/Button.vue'
 import Message from '@/components/misc/Message.vue'
 import CustomTransition from '@/components/misc/CustomTransition.vue'
 
@@ -58,6 +71,8 @@ import {useTitle} from '@/composables/useTitle'
 import {useBaseStore} from '@/stores/base'
 import {useProjectStore} from '@/stores/projects'
 import {useViewFiltersStore} from '@/stores/viewFilters'
+
+import {PERMISSIONS} from '@/constants/permissions'
 
 import type {IProject} from '@/modelTypes/IProject'
 import type {IProjectView} from '@/modelTypes/IProjectView'
@@ -83,6 +98,10 @@ const currentProject = computed<IProject>(() => {
 	}
 })
 useTitle(() => currentProject.value?.id ? getProjectTitle(currentProject.value) : '')
+
+const canWrite = computed(() => {
+	return (currentProject.value?.maxPermission ?? 0) > PERMISSIONS.READ && currentProject.value?.id > 0
+})
 
 const views = computed(() => projectStore.projects[props.projectId]?.views)
 
--- a/frontend/src/components/tasks/partials/Heading.vue
+++ b/frontend/src/components/tasks/partials/Heading.vue
@@ -67,6 +67,7 @@
 <script setup lang="ts">
 import {ref, computed, onMounted, onBeforeUnmount, watch} from 'vue'
 import {useRouter} from 'vue-router'
+import {useI18n} from 'vue-i18n'
 
 import BaseButton from '@/components/base/BaseButton.vue'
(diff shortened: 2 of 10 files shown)


maggch97:
## Future improvements

The following features are currently hidden in create mode because their APIs require a real task ID. They can be added in follow-up PRs using the same pattern as assignees (store locally, call API after creation):

- **Labels** — `EditLabels` already has partial `taskId === 0` handling. Need to fix `createAndAddLabel` to not early-return when `taskId === 0`, enable the labels section in the template, and call `taskStore.addLabel()` in `createNewTask()` after creation.
- **Attachments** — Would need to buffer selected files locally, upload via `PUT /tasks/{taskId}/attachments` after creation. This would also enable **image insertion in the description editor** during creation.
- **Related tasks** — Similar pattern: store relation definitions locally, call the relation API after creation.

frontend/src/components/project/ProjectWrapper.vue
```diff
@@ -28,7 +28,19 @@
 					{{ getViewTitle(view) }}
 				</BaseButton>
 			</div>
-			<slot name="header" />
+			<div
+				class="is-flex is-align-items-center"
+				style="gap: .5rem"
+			>
+				<slot name="header" />
+				<XButton
```
kolaente, line 36: I think we should add this only to the Table view and maybe to the Gantt view - the Kanban and List views already have ways to create tasks. Having this button as well would make it confusing.

kolaente (commented):
Thanks!

I'm unsure if this is the best way to do this, especially since the task detail view is getting cluttered with these many new checks. What do you think about creating just a simple form that allows to set a title, description, maybe color and dates? Then redirect to the task's detail view, allowing to edit the task with all related properties.

In either case, it should auto focus the title input and be shown in a modal instead of this full view.

maggch97:
> Thanks!
> 
> I'm unsure if this is the best way to do this, especially since the task detail view is getting cluttered with these many new checks. What do you think about creating just a simple form that allows to set a title, description, maybe color and dates? Then redirect to the task's detail view, allowing to edit the task with all related properties.
> 
> In either case, it should auto focus the title input and be shown in a modal instead of this full view.

I use Azure DevOps a lot and it has a same page layout for create and edit, so I prefer this instead of a new simple dialog. I believe different users may have different requirements for task creation, some users may want to set labels, some users may want to set assignees and some others may want to upload some files. So finally, we may want the new task page to be consistent with the details page.

maggch97:
Updated per review feedback:
- New Task button now only appears in Table view (List/Kanban/Gantt have their own creation UIs)
- Opens as a modal instead of a full page view
- Auto-focuses the title on open
- E2E tests updated accordingly, all 14 passing

frontend/src/components/tasks/partials/Heading.vue
```diff
@@ -103,7 +107,12 @@ async function copyUrl() {
 const taskStore = useTaskStore()
 const loading = computed(() => taskStore.isLoading)
 
-const textIdentifier = computed(() => getTaskIdentifier(props.task))
+const textIdentifier = computed(() => {
+	if (isNewTask.value) {
+		return t('task.detail.newTaskIdentifier')
```
kolaente, line 112: We should just hide the identifier if it's new

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -308,7 +311,7 @@
 
 					<!-- Labels -->
 					<div
-						v-if="activeFields.labels"
+						v-if="activeFields.labels && !isNewTask"
```
kolaente, line 314: For all of these if checks: If we have the button that triggers `activeFields.labels` we should not need this check as the value can never be true.

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -694,6 +722,9 @@ const kanbanStore = useKanbanStore()
 const authStore = useAuthStore()
 const baseStore = useBaseStore()
 
+const isNewTask = computed(() => props.taskId === 0 && (props.newTaskProjectId ?? 0) > 0)
+provide('isNewTask', isNewTask)
```
kolaente, line 726: Any special reason to use provide/inject instead of just passing a prop?

frontend/src/views/tasks/TaskDetailView.vue
```diff
@@ -1047,6 +1103,36 @@ async function saveTask(
 	success({message: t('task.detail.updateSuccess')}, actions)
 }
 
+async function createNewTask() {
+	const t_ = task.value
```
kolaente, line 1107: Why this extra variable? Why the `newTask` const instead of passing `task.value` to the service directly?

kolaente (changes requested):
I've tested this on the preview, but it did not open as a modal for me. And the E2E test seem missing. Did you push your changes?

PR #2367, opened 2026-03-05
fix(menu): prevent dropdown from closing when cursor crosses offset gap

### LINKED ISSUE #2311: Dropdown menu running away from you (not literally)
### Pre-submission checklist

### Description

When clicking the dots behind a project, in the sidebar, the dropdown disappears the moment I stop hovering the dots.
This is only the case (it seems) when the sidebar has a lot of projects, and - as you can see in the video - only happens when the dropdown tries to render below it.
Couldn't reproduce on demo instance, because I think it did not have enough projects, and was too lazy (sorry, love you guys).

### Vikunja Version

v1.1.0-137-d637c52b88 <- currently cloned one with a feature that's not yet on mainstream, tho nothing changing to this, and had it before.

### Browser and version

_No response_

Yes

### Screenshots

https://cdn.funzynoob.tv/chrome_Mb6WsQwP9b.mp4

n-nkm:
Fix #2311 by adding an invisible ::before inset element that is as big as the offset from floating UI so the cursor does not accidentally hover a different, underlying element and closes the dropdown. 

Added the offset as a style variable so any future changes don't bring up a "magic number" issue.

Testing efforts:
- Ran unit tests in frontend  
- Tested (manually) several resolutions ranging from 1920X1080 to 800 x 400 (so the hover goes from the bottom up)
- Cross checked the secondary dropdown on the top right to keep working

I did NOT get playwright to run on this setup, unfortunately. I would appreciate if somebody could run it through

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -64,7 +64,9 @@
       "electron"
     ],
     "overrides": {
-      "minimatch": "^10.2.3"
+      "minimatch": "^10.2.3",
+      "tar": "^7.5.10",
+      "@tootallnate/once": "^3.0.1"
     }
   }
 }
--- a/desktop/pnpm-lock.yaml
+++ b/desktop/pnpm-lock.yaml
@@ -6,6 +6,8 @@ settings:
 
 overrides:
   minimatch: ^10.2.3
+  tar: ^7.5.10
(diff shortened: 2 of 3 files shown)


kolaente (approved):
Thanks!

Issue #2393, opened 2026-03-14
Add api_min_compatible field to /api/v1/info for client compatibility


ilvez:
## Problem

The mobile app (and potentially other third-party clients) has no reliable way to determine if it's compatible with a given server version. Currently the app hardcodes an exact server version (`supportedServerVersion = Version(2, 1, 0)`) and shows a "Version mismatch — use at your own risk!" warning whenever the server reports anything different.

This creates a few problems:

- Every server upgrade triggers the warning until a new app version is published with an updated constant
- Users see the warning on **every app launch** (and on share-to-app), even when nothing is actually broken
- The app can't distinguish between "server is newer but still compatible" and "server made breaking API changes"
- Related issues: go-vikunja/app#235, go-vikunja/app#204, go-vikunja/app#241

## Proposal

Add a single new field to the `/api/v1/info` response:

```json
{
  "version": "2.2.0",
  "api_min_compatible": "2.1.0",
  ...
}
```

`api_min_compatible` declares the oldest client target version that still works with this server. It only gets bumped when a breaking API change ships (endpoint removal/rename, changed request/response shape, etc.).

### Client-side logic

A client built for version X is compatible if:

```
api_min_compatible <= X <= server version
```

- If `X < api_min_compatible` → client is too old, warn/block
- If `X > server version` → server is too old, warn
- Otherwise → compatible, no warning needed
- If field is missing (older server) → fall back to current behavior

### Why this approach

- One new field, no new endpoints or headers
- Server-side it's just a constant that gets updated on breaking changes
- Backwards compatible — clients that don't know about the field ignore it, clients that do get smarter behavior
- The existing boolean feature flags in `/info` (`caldav_enabled`, `task_comments_enabled`, etc.) already cover per-feature discovery — this only needs to cover API-level breaking changes

### Alternatives considered

- **Per-feature capability strings** (e.g. `capabilities: ["tasks.v2", "reactions"]`) — more granular but adds complexity that isn't needed given the project's release cadence. Existing boolean flags already serve this purpose.
- **API version headers per-request** — requires changes across all endpoints, high implementation cost.
- **Monotonically increasing schema version number** — loses semver semantics, harder to reason about.

ilvez:
Got essentially resolved by: https://github.com/go-vikunja/app/pull/252

PR #2394, opened 2026-03-14
Feat: API minimum compatible version

ilvez:
Potential solution for introducing minimum compatible app version checks.
  - Add `api_min_compatible` field to `/api/v1/info` endpoint for range-based client compatibility checks
  - Add snapshot test for the info endpoint

Proposal: https://github.com/go-vikunja/vikunja/issues/2393


ilvez:
Got essentially resolved by: https://github.com/go-vikunja/app/pull/252

PR #2408, opened 2026-03-17
Fix tasks api expand parameters

jeanyvesl:
## ✨ Key Changes

- __Normalized Parsing:__ Added `ParseExpandParameters` in `pkg/models/task_collection.go` to seamlessly process, validate, and deduplicate values from both the `expand[]` array query format and the `expand` CSV query format.
- __Struct Enhancements:__ Updated `TaskCollection` and `Task` structs to include a new `ExpandCSV` field mapped to `query:"expand"`.
- __Query Handlers Updated:__ Modified `TaskCollection.ReadAll` and `Task.ReadOne` to safely pass both parameter sets through the unified parsing logic before querying relationships.
- __Comprehensive Testing:__ Added a dedicated unit test suite (`TestParseExpandParameters`) in `pkg/models/task_expand_test.go` covering empty cases, mixed parameters, malformed CSV spaces, duplicate values, and invalid parameters.

## 🧪 Testing Instructions

- __CSV Format (New):__ `GET /api/v1/tasks?expand=subtasks,comments`
- __Array Format (Legacy):__ `GET /api/v1/tasks?expand[]=subtasks&expand[]=comments`
- __Mixed Formatting:__ `GET /api/v1/tasks?expand=subtasks&expand[]=comments`

All of the above will successfully expand both `subtasks` and `comments`. Invalid values correctly return a validation error.


kolaente:
Why this change? What's the CSV format?

jeanyvesl:
> Why this change? What's the CSV format?

I made the change since the API documentation is not reflecting the actual behaviour of the API. The CSV format that's what it's referred to in the swagger pkg/swagger/swagger.json
```
{
    "type": "array",
    "items": {
        "type": "string"
    },
    "collectionFormat": "csv",
    "description": "If set to `subtasks`, Vikunja will fetch only tasks which do not have subtasks and then in a second step, will fetch all of these subtasks. This may result in more tasks than the pagination limit being returned, but all subtasks will be present in the response. If set to `buckets`, the buckets of each task will be present in the response. If set to `reactions`, the reactions of each task will be present in the response. If set to `comments`, the first 50 comments of each task will be present in the response. You can set this multiple times with different values.",
    "name": "expand",
    "in": "query"
}
```

jeanyvesl:
More concretely, based on the API documentation, I was trying to do 
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand=buckets`
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand=comments`
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand=subtasks`
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand=subtasks,comments,buckets`

And it did not work, but then I looked into the code and I figured out that I should try the following:
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand[]=buckets`
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand[]=comments`
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand[]=subtasks`
`GET` https://vikunja.cloudron.may4th.be/api/v1/tasks/4?expand[]=subtasks&expand[]=comments&expand[]=buckets`

And it worked smoothly. So I figured, better to revamp the API underlying logic to match the API description since it is more elegant than the array approach. If that does not fit what you were expecting it to do, then let's change my approach and modify the API documentation then.

kolaente:
This is definitely an inconsistency, it should work (and be documented) in the same way as the `sort_by`/`order_by` query params. I think the solution in this PR is a very complicated on, I've opened a follow-up to implement a simpler solution here: https://github.com/go-vikunja/vikunja/pull/2415

PR #2417, opened 2026-03-19
Fix iOS/MacOS Reminders CalDAV Support

hcuk94:
Resolves issue #475 by modifying CalDAV discovery so Apple Reminders can use /dav/projects/ as the home set without exposing that synthetic path as a real task list, preserving the existing principal-based flow. This is because Apple Reminders defaults back to the /dav/projects/ URL, rather than accepting the /dav/principals/username/ URL specified in Vikunja.

### DIFF
--- a/config-raw.json
+++ b/config-raw.json
@@ -974,6 +974,11 @@
                     "key": "proxypassword",
                     "default_value": "",
                     "comment": "The proxy password to use when authenticating against the proxy."
+                },
+                {
+                    "key": "allownonroutableips",
+                    "default_value": "false",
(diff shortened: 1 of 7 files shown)


pkg/routes/caldav/listStorageProvider.go
```diff
@@ -144,6 +155,24 @@ func (vcls *VikunjaCaldavProjectStorage) GetResources(rpath string, withChildren
...
+	}
+	return path + "/"
+}
+
+func principalPathForUser(username string) string {
+	return withTrailingSlash(PrincipalBasePath + `/` + username)
+}
+
+func projectHomeSetPath() string {
+	return ProjectBasePath + `/`
```
kolaente: If this is just a const + const, please make the result a const instead of defining a new function.

hcuk94: Good call, updated now with that change.

PR #2430, opened 2026-03-20
feat: Add option to select landing page

surfingbytes:
Currently, after login, vikunja opens last loaded page after login.

This change gives user choice whether he wants to load last loaded page (default = no behaviour change), overview, upcoming, or default project. It also does this redirect not only after login, but when the app is loaded - for example when user opens this app but is already logged in (open vikunja, log in, close tab, do some other work, open new tab with vikunja)


frontend/src/router/index.ts
```diff
@@ -42,6 +43,39 @@ const router = createRouter({
...
+
+				const authStore = useAuthStore()
+				const defaultPage = authStore.settings?.frontendSettings?.defaultPage
+
+				switch (defaultPage) {
+					case DEFAULT_PAGE.UPCOMING:
+						return {name: 'tasks.range'}
+					case DEFAULT_PAGE.DEFAULT_PROJECT: {
+						const projectId = authStore.settings?.defaultProjectId
+						if (projectId) {
```
kolaente: This should check if the project exists

kolaente (changes requested):
Hey, thanks for the PR! Looks good overall, but some changes required.

Please also add E2E tests via Playwright for this.

frontend/src/router/index.ts
surfingbytes: Check added, tests too

frontend/src/composables/useRedirectToLastVisited.ts
```diff
@@ -30,7 +31,7 @@ export function useRedirectToLastVisited() {
 				return {name: 'tasks.range'}
 			case DEFAULT_PAGE.DEFAULT_PROJECT: {
 				const projectId = authStore.settings?.defaultProjectId
-				if (projectId) {
+				if (projectId && useProjectStore().projects[projectId]) {
```
kolaente: Please instantiate the store first

also in the router

kolaente (changes requested):

surfingbytes:
Project store instantiated

surfingbytes:
What blocks this PR please? Would be really helpful to have this app start on default project instead of showing all tasks across projects...

surfingbytes:
Please run the tests again - it failed on postgres image pull..

kolaente:
The last e2e test failure seems to be an actual test failure.

PR #2436, opened 2026-03-22
feat: Better support for deploying the frontend at a subpath

mdrkrg:
Make the frontend work correctly when served at a subpath (e.g. `/vikunja/` instead of `/`). Previously, API URLs, OIDC redirects, CalDAV URLs, and the service worker all had paths hardcoded relative to the domain root, which broke any deployment behind a reverse proxy with a path prefix.

This PR is based on the fix in #2440 .

## What changed

- **`index.html` / `static.go`**: Replaced the hardcoded `window.API_URL` value with a `__VIKUNJA_API_URL__` marker. Vite replaces it at build time (for standalone nginx deploys) if `VIKUNJA_BUILD_STANDALONE=true` is set in `.env.local`. Otherwise the Go backend replaces it at serve time using `service.publicurl`.

  - I'm unsure about this change. The original `static.go` targets `'http://localhost:3456/api/v1'` to replace, but in actual `index.html` it is `'/api/v1'`, so this is already broken?

  - Current implementation would work for this case: Serving frontend and backend on the same server under a subpath (e.g., `example.com/vikunja`), in which it does not matter whether Vite does the `API_BASE` replacement or the go backend.

  - ~~However, it is not feasible to serve frontend at subpath but backend at subdomain (e.g., `example.com/vikunja` served by nginx, and backend at `api.example.com`). Might need some refactoring.~~ Fixed in 2e134cec1. See [this comment](#issuecomment-4105788703).

- **`checkAndSetApiUrl.ts`**: Extracted `API_DEFAULT_PATH` constant and `joinPath`/`hasApiPath` helpers.

- **`config.ts` (`apiBase` computed)**: See !2435. Uses `pathname` and strips `/api/v1` to derive the deployment base. This fixes CalDAV URL construction in `Caldav.vue`.

- **`redirectToProvider.ts`**: OIDC redirect URL now includes the base path via `getFullBaseUrl()`, so the callback lands on a route the Vue router handles.

- **`sw.ts`**: Service worker API route regex is anchored to the configured base path instead of matching `api/v1/` anywhere in the URL.

- **`auth/auth.go`**: Compute refresh token cookie base path from `service.publicurl` rather than hard coding the base as `/api/v1`.

- **`api/v1/docs.go`**: Build `RedocUITemplate` as template, and replace with `apiBase + api/v1/docs.json`.

- **Test files**: Centralized `TEST_API_URL` constant so the API base URL for tests is defined in one place.

- **`vite.config.ts`**: Added a `transformIndexHtml` plugin for the marker replacement. Made the dev proxy path and rewrite dynamic based on `VIKUNJA_FRONTEND_BASE` so local development works with a subpath.

## How to deploy at a subpath

1. Set `VIKUNJA_FRONTEND_BASE` to the desired path (e.g. `/vikunja/`) before building the frontend. This sets Vite's `base` and bakes the correct API URL into the built HTML.

2. Configure the reverse proxy to forward requests from the subpath to Vikunja. The Go backend still expects `/api/v1` at its root — the proxy should strip the prefix before forwarding.

   Example Caddyfile:

   ```Caddyfile
   example.com:80 {
    handle_path /vikunja* {
     reverse_proxy 127.0.0.1:3456
    }
   }
   ```

3. Set `service.publicurl` in `config.yml` to the full public URL including the subpath (e.g. `https://example.com/vikunja`). When the Go backend serves the frontend, it will replace the `window.API_BASE` placeholder.

4. For OIDC: no extra configuration needed — the redirect URL is now built dynamically from the frontend's base path.

5. For migration services (Trello, Todoist, Microsoft To Do): the redirect URLs in `config.yml` are absolute URLs set by the admin, so include the subpath there if applicable.

## Docs changes needed

The following should be mentioned in the deployment documentation:

- **`VIKUNJA_FRONTEND_BASE`**: (Already there).

- **`VIKUNJA_API_URL`**: The full URL to the Vikunja API. Only required when building the standalone frontend (see below `VIKUNJA_BUILD_STANDALONE`). If unset when building a standalone frontend, it would be automatically derived as `VIKUNJA_FRONTEND_BASE + /api/v1`. Only set this when the backend is at a different origin or path than the frontend.

- **`VIKUNJA_BUILD_STANDALONE`**: If building a frontend that is intended to be served at a subpath by a reverse proxy, set this to `true` so that Vite replaces `__VIKUNJA_API_URL__` with `VIKUNJA_API_URL` at build time.
  Otherwise go backend will handle the replacement using `service.publicurl`.

- **Reverse proxy subpath configuration**: Document that when deploying at a subpath, the reverse proxy should strip the prefix before forwarding to the backend.
  Include example configs for common web servers.

- **Include the subpath in `service.publicurl`**: If the Go backend serves the frontend and the deployment uses a subpath, `service.publicurl` must reflect that (e.g. `https://example.com/vikunja`).

- **Migration redirect URLs**: Admins must include the subpath in `migration.*.redirecturl` values in `config.yml` if deploying at a subpath. These are not auto-detected.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -349,7 +349,9 @@ jobs:
           echo "{\"VERSION\": \"${{ steps.ghd.outputs.describe }}\"}" > src/version.json
       - name: Build frontend
         working-directory: frontend
-        run: VIKUNJA_BUILD_STANDALONE=true pnpm build
+        env:
+          VIKUNJA_BUILD_STANDALONE: true
+        run: pnpm build
       - name: Store Frontend
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -349,7 +349,7 @@ jobs:
           echo "{\"VERSION\": \"${{ steps.ghd.outputs.describe }}\"}" > src/version.json
       - name: Build frontend
         working-directory: frontend
-        run: pnpm build
+        run: VIKUNJA_BUILD_STANDALONE=true pnpm build
       - name: Store Frontend
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
         with:
--- a/frontend/src/helpers/checkAndSetApiUrl.ts
+++ b/frontend/src/helpers/checkAndSetApiUrl.ts
@@ -1,7 +1,7 @@
 import {useConfigStore} from '@/stores/config'
 
 const API_DEFAULT_PORT = '3456'
-const API_DEFAULT_PATH = '/api/v1'
+const API_PATH_SUFFIX = '/api/v1'
 
 export const ERROR_NO_API_URL = 'noApiUrlProvided'
 
@@ -35,7 +35,7 @@ function joinPath(base: string, suffix: string): string {
  */
 function hasApiPath(pathname: string): boolean {
 	const clean = pathname.endsWith('/') ? pathname.slice(0, -1) : pathname
-	return clean.endsWith(API_DEFAULT_PATH)
+	return clean.endsWith(API_PATH_SUFFIX)
 }
 
 export const checkAndSetApiUrl = (pUrl: string | undefined | null): Promise<string> => {
@@ -77,7 +77,7 @@ export const checkAndSetApiUrl = (pUrl: string | undefined | null): Promise<stri
 			console.warn(`Could not fetch 'info' from the provided endpoint ${pUrl} on ${window.API_URL}/info. Some automatic fallback will be tried.`)
 			// Check if it is reachable at the base path + /api/v1 via http
 			if (!hasApiPath(urlToCheck.pathname)) {
-				urlToCheck.pathname = joinPath(urlToCheck.pathname, API_DEFAULT_PATH)
+				urlToCheck.pathname = joinPath(urlToCheck.pathname, API_PATH_SUFFIX)
 				window.API_URL = urlToCheck.toString()
 				return configStore.update()
 			}
@@ -87,7 +87,7 @@ export const checkAndSetApiUrl = (pUrl: string | undefined | null): Promise<stri
 			// Check if it is reachable at the base path + /api/v1 via https
 			urlToCheck.pathname = origPathname
 			if (!hasApiPath(urlToCheck.pathname)) {
-				urlToCheck.pathname = joinPath(urlToCheck.pathname, API_DEFAULT_PATH)
+				urlToCheck.pathname = joinPath(urlToCheck.pathname, API_PATH_SUFFIX)
 				window.API_URL = urlToCheck.toString()
 				return configStore.update()
 			}
@@ -106,7 +106,7 @@ export const checkAndSetApiUrl = (pUrl: string | undefined | null): Promise<stri
 			// Check if it is reachable at :API_DEFAULT_PORT with base path + /api/v1
 			urlToCheck.pathname = origPathname
 			if (!hasApiPath(urlToCheck.pathname)) {
-				urlToCheck.pathname = joinPath(urlToCheck.pathname, API_DEFAULT_PATH)
+				urlToCheck.pathname = joinPath(urlToCheck.pathname, API_PATH_SUFFIX)
 				window.API_URL = urlToCheck.toString()
 				return configStore.update()
 			}
--- a/frontend/src/helpers/checkAndSetApiUrl.ts
+++ b/frontend/src/helpers/checkAndSetApiUrl.ts
@@ -1,6 +1,7 @@
 import {useConfigStore} from '@/stores/config'
 
 const API_DEFAULT_PORT = '3456'
+const API_DEFAULT_PATH = '/api/v1'
 
 export const ERROR_NO_API_URL = 'noApiUrlProvided'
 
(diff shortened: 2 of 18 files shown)


frontend/src/helpers/checkAndSetApiUrl.ts
```diff
@@ -46,7 +64,7 @@ export const checkAndSetApiUrl = (pUrl: string | undefined | null): Promise<stri
 		throw new InvalidApiUrlProvidedError()
 	}
 
-	const origUrlToCheck = urlToCheck
+	const origPathname = urlToCheck.pathname
```
mdrkrg: I think this might better be a separate bugfix PR?

mdrkrg:
Compare different deployment cases

|  | Frontend | Backend | Config |
|---|----------|---------|---------------|
| 1 | `example.com/` | `example.com/api/v1` | Nothing (default) |
| 2 | `example.com/vikunja/` | `example.com/vikunja/api/v1` | `VIKUNJA_FRONTEND_BASE=/vikunja/` (Go backend `service.publicurl` or standalone) |
| 3 | `example.com/vikunja/` | `api.example.com/api/v1` | `VIKUNJA_FRONTEND_BASE=/vikunja/` + (Go backend `service.publicurl` or standalone `VIKUNJA_API_URL=https://api.example.com/api/v1`) |
| 4 | `app.example.com/` | `api.example.com/api/v1` |Go backend  `service.publicurl` or standalone `VIKUNJA_API_URL=https://api.example.com/api/v1` |
| 5 | `example.com/app/` | `example.com/backend/api/v1` | `VIKUNJA_FRONTEND_BASE=/app/` + (Go backend `service.publicurl` or standalone `VIKUNJA_API_URL=/backend/api/v1`) |

Case 1 and 4 work without a frontend rebuild. Case 2, 3, and 5 require rebuild w/ or w/o `VIKUNJA_BUILD_STANDALONE=true`, depending on who serves the static files.

Am I making this overly complicated?

pkg/routes/api/v1/docs.go
mdrkrg, line 1: Better new PR to move the docs as seperate HTML.

mdrkrg:
I'm passing this test locally. The testing CI does not set `VIKUNJA_BUILD_STANDALONE=true` when building the frontend so it fails...

There might a better solution :thinking:

pkg/routes/api/v1/docs.go
mdrkrg: #2465

mdrkrg:
Hey @kolaente, I think things are mostly done. Could you drop a review on this? Thanks in advance!

.github/workflows/test.yml
```diff
@@ -349,7 +349,7 @@ jobs:
           echo "{\"VERSION\": \"${{ steps.ghd.outputs.describe }}\"}" > src/version.json
       - name: Build frontend
         working-directory: frontend
-        run: pnpm build
+        run: VIKUNJA_BUILD_STANDALONE=true pnpm build
```
kolaente: Please set this as an env variable on the job

frontend/src/helpers/checkAndSetApiUrl.ts
```diff
@@ -1,6 +1,7 @@
 import {useConfigStore} from '@/stores/config'
 
 const API_DEFAULT_PORT = '3456'
+const API_DEFAULT_PATH = '/api/v1'
```
kolaente: Maybe we can call this `API_PATH_SUFFIX` or something to make it clear how its used

pkg/modules/auth/auth.go
```diff
@@ -47,8 +48,21 @@ type Token struct {
...
-const RefreshTokenCookieName = "vikunja_refresh_token"      //nolint:gosec // not a credential
-const refreshTokenCookiePath = "/api/v1/user/token/refresh" //nolint:gosec // not a credential
+const RefreshTokenCookieName = "vikunja_refresh_token" //nolint:gosec // not a credential
+
+// getRefreshTokenCookiePath returns the cookie path for the refresh token,
+// derived from service.publicurl.
+func getRefreshTokenCookiePath() string {
+	publicURL := config.ServicePublicURL.GetString()
+	u, err := url.Parse(publicURL)
+	if err != nil || publicURL == "" || publicURL == "/" {
```
kolaente: This works, but do we really need the if?

kolaente (changes requested):
Let me challenge your approach more:

What is not working for you with the current documented approach for deploying Vikunja at a subpath?

Why do you need to deploy Vikunja at a subpath at all?

Not trying to be mean or anything, I just want to understand what brought this change about and if this is really the best solution.

kolaente:
Please resolve the conflicts.

mdrkrg:
@kolaente Thanks for the review, and here are answers to your questions:

1. The current documented approach does not work quite well, because in the frontend there are still several hard coded `/api/v1`. Also in `static.go`, the go backend will attempt to replace `window.API_URL = '/api/v1'` with `service.publicurl`, but as for the time of implementation, this seems to be broken because of a string mismatch. As a consequence the user have to input the correct backend url on initial load, which is not quite desirable.
2. I know this is rare and weird, but imagine someone who have no control of an `*.example.com` domain but want to deploy several different apps, they can only stuff everything under subpaths, which is exactly my case. Since the codebase already have a solid foundation for this feature, I think it would be helpful to support this fully. However if this adds unnecessary maintenance burden for such a small demand, I guess it probably won't be too hard for people with this demand maintain their own fork based on the changes of this PR.

They only thing I'm concerned is that, if we only need to support root path and sub-path deployment, some of the complexity of this PR might be unnecessary, e.g. the `__VIKUNJA_API_URL__` replacement and `VIKUNJA_BUILD_STANDALONE` env var. However I'm out of ideas.

mdrkrg:
I think we can split the PR into two separate ones. One addresses the hard-coded '/api/v1' api base in the frontend / backend, to make path based serving actually work, but ignore `window.API_URL`. The other addresses the broken `window.API_URL` replacement in `static.go`.

mdrkrg:
This is largely resolved

PR #2465, opened 2026-03-24
fix(api): Use embed fs to embed the redoc template html

mdrkrg:
Cleanup the code a little bit? There are to much escapes to "\`".

Have no idea where to put the templates.


kolaente:
Hey thanks for this!

I've taken the approach and modified it a little in this PR: https://github.com/go-vikunja/vikunja/pull/2503

PR #2472, opened 2026-03-24
feat: detect and prevent duplicate OpenID Connect issuers

kolaente:
## Summary
This PR adds validation to detect and prevent multiple OpenID Connect providers from being configured with the same issuer URL, which would cause team sync conflicts.

## Key Changes
- **New error type** (`ErrDuplicateOIDCIssuer`): Added a custom error type in `pkg/modules/auth/openid/providers.go` to represent duplicate issuer configuration errors, with a helper function `IsErrDuplicateOIDCIssuer()` for type checking.

- **Duplicate issuer detection in provider initialization**: Modified `GetAllProviders()` to check for duplicate issuers across all configured providers after they are loaded. If duplicates are found, initialization fails with the new error type.

- **Startup validation**: Updated `pkg/initialize/init.go` to treat duplicate issuer errors as fatal configuration errors, preventing the application from starting with conflicting provider configurations.

- **Doctor command enhancement**: Enhanced the OpenID diagnostic check in `pkg/doctor/services.go` to:
  - Extract issuer URLs from provider discovery responses
  - Detect and report duplicate issuers among successfully configured providers
  - Provide clear error messages identifying which providers share the same issuer

- **Comprehensive test coverage**: Added three new test cases in `pkg/modules/auth/openid/providers_test.go`:
  - `TestDuplicateIssuersDetected`: Verifies that duplicate issuers are properly detected and rejected
  - `TestUniqueIssuersAllowed`: Confirms that providers with different issuers load successfully
  - `TestFailedDiscoverySkippedInIssuerCheck`: Ensures that providers with failed discovery don't block other providers

## Implementation Details
- The duplicate detection works by collecting issuer URLs from successful provider discovery and checking for collisions
- Providers that fail OIDC discovery are skipped during issuer validation to avoid false positives
- A mock OIDC server helper (`newMockOIDCServer()`) was added to support testing with realistic discovery documents

https://claude.ai/code/session_01H2W7YWuhtbq8GhUD7QM7T3

### DIFF
--- a/pkg/doctor/services.go
+++ b/pkg/doctor/services.go
@@ -18,13 +18,16 @@ package doctor
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"time"
 
 	"code.vikunja.io/api/pkg/config"
 	"code.vikunja.io/api/pkg/modules/auth/ldap"
+	"code.vikunja.io/api/pkg/modules/auth/openid"
 	"code.vikunja.io/api/pkg/red"
 )
 
@@ -243,9 +246,22 @@ func checkOpenID() CheckGroup {
 	}
 
 	var results []CheckResult
+	providerIssuers := make(map[string]string)
 	for key, p := range providerMap {
-		result := checkOpenIDProvider(key, p)
+		result, issuer := checkOpenIDProvider(key, p)
 		results = append(results, result)
+		if issuer != "" {
+			providerIssuers[key] = issuer
+		}
+	}
+
+	// Check for duplicate issuers among successful providers
+	for _, dup := range openid.FindDuplicateIssuers(providerIssuers) {
+		results = append(results, CheckResult{
+			Name:   "Duplicate Issuer",
+			Passed: false,
+			Error:  dup.Error(),
+		})
 	}
 
 	return CheckGroup{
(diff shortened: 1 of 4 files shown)


pkg/modules/auth/openid/providers.go
```diff
@@ -97,6 +117,25 @@ func GetAllProviders() (providers []*Provider, err error) {
...
 			}
 		}
+
+		// Check for duplicate issuers across providers
+		issuerToKey := make(map[string]string)
+		for _, p := range providers {
+			issuer, issuerErr := p.Issuer()
+			if issuerErr != nil {
+				// Provider failed OIDC discovery, skip issuer check
+				continue
```
kolaente, line 146: We should at least log the error - what happens in other cases where that fails?

pkg/doctor/services.go
```diff
@@ -243,9 +245,30 @@ func checkOpenID() CheckGroup {
...
 		results = append(results, result)
+		if issuer != "" {
+			providerIssuers = append(providerIssuers, providerIssuer{key: key, issuer: issuer})
+		}
+	}
+
+	// Check for duplicate issuers among successful providers
+	issuerToKey := make(map[string]string)
+	for _, pi := range providerIssuers {
+		if existingKey, exists := issuerToKey[pi.issuer]; exists {
```
kolaente: This is duplicate logic now?

kolaente: It's still duplicate. Refactor so that there's a function that does the check and returns all duplicates, then use that in both places.

PR #2476, opened 2026-03-25
feat: improve wording and UX around CalDAV tokens

milzer:
Better phrasing for CalDAV texts. Previously it was implied that using token is optional.

### DIFF
--- a/frontend/src/i18n/lang/en.json
+++ b/frontend/src/i18n/lang/en.json
@@ -151,7 +151,7 @@
         "more": "More information about CalDAV in Vikunja",
         "tokens": "CalDAV Tokens",
         "tokensHowTo": "For CalDAV authentication you can use either your normal account password or a dedicated CalDAV token.",
-        "createToken": "Create a new CalDAV token",
+        "createToken": "Create a CalDAV token",
         "tokenCreated": "Here is your new token: {token}",
         "wontSeeItAgain": "Write it down or save it securely — you will not be able to see it again.",
         "mustUseToken": "You need to create a CalDAV token to use CalDAV with any third-party client. Enter the token in the password field of your client.",


frontend/src/i18n/lang/en.json
```diff
@@ -147,15 +147,15 @@
...
+        "howTo": "You can connect Vikunja to CalDAV clients to view and manage all your tasks from different clients. Enter this URL into your client:",
         "more": "More information about CalDAV in Vikunja",
         "tokens": "CalDAV Tokens",
-        "tokensHowTo": "You can use a CalDAV token to use instead of a password to log in the above endpoint.",
-        "createToken": "Create a token",
-        "tokenCreated": "Here is your token: {token}",
-        "wontSeeItAgain": "Write it down, you won't be able to see it again.",
-        "mustUseToken": "You need to create a CalDAV token if you want to use CalDAV with a third party client. Use the token as the password.",
-        "usernameIs": "Your username is: {0}"
+        "tokensHowTo": "CalDAV requires a dedicated token for authentication. You must use a CalDAV token as the password (your normal account password will not work).",
```
kolaente: The token is only required when you're using oauth. Can you adjust it so that your message shows up only when the user using oauth?

milzer: Ok, so my actual problem was that I tried my email as username. How about changing the text to `For CalDAV authentication you can use either your normal account password or a dedicated CalDAV token.` and show the username for all users? I could include removing the `<template v-if="!isLocalUser">` in the PR, but I can't run the tests locally as setting up the dev environment in my immutable distro is a tad challenging.

kolaente: Text change sounds great, please go ahead!

> I can't run the tests locally as setting up the dev environment in my immutable distro is a tad challenging.

NixOS? Try the devenv and the mage commands, that works for me.

milzer: Ok I managed to install devenv and everything seems to work after the changes.

frontend/src/i18n/lang/en.json
```diff
@@ -147,15 +147,15 @@
...
         "more": "More information about CalDAV in Vikunja",
         "tokens": "CalDAV Tokens",
-        "tokensHowTo": "You can use a CalDAV token to use instead of a password to log in the above endpoint.",
-        "createToken": "Create a token",
-        "tokenCreated": "Here is your token: {token}",
-        "wontSeeItAgain": "Write it down, you won't be able to see it again.",
-        "mustUseToken": "You need to create a CalDAV token if you want to use CalDAV with a third party client. Use the token as the password.",
-        "usernameIs": "Your username is: {0}"
+        "tokensHowTo": "For CalDAV authentication you can use either your normal account password or a dedicated CalDAV token.",
+        "createToken": "Create a new CalDAV token",
```
kolaente:
```suggestion
        "createToken": "Create a CalDAV token",
```

It's not possible to create an "old" token, hence this is the correct phrasing :)

PR #2486, opened 2026-03-26
config: Expand environment variables in `some.config.value.path.file`

tfc:
This little change adds the possibility to have config values like:

```yaml
mailer:
  password:
    file: $CREDENTIALS_DIRECTORY/smtp.secret
```

Which allows to provide such secrets via systemd's `LoadCredential` directives.
See official docs: https://systemd.io/CREDENTIALS/


tfc:
cc @kolaente

tfc:
Amazing, thank you @kolaente for the quick reaction!

PR #2491, opened 2026-03-26
chore: add .pnpm-store to .gitignore

surfingbytes:
Added .pnpm-store to .gitignore so new devs have easier start - this is a cache folder (similar to node_modules) and should not be tracked.


kolaente:
I wonder why that is used in the git repo? I've only ever had that stored in my home directory.

PR #2492, opened 2026-03-26
Automatic OIDC redirect

surfingbytes:
When both local and LDAP auth are disabled and there's exactly one OIDC provider, it shows login page where user can't do anything but click the single button. This change redirects user to the OIDC provider automatically in that case - no point of displaying one single button which redirects there after click.


kolaente:
We had this in the past but removed it because it created redirect chains where the user clicked logout -> redirect to the openid provider -> redirect back to Vikunja -> login again.

The pragmatic solution here would be to use the `?redirectToProvider=foobar` query param from wherever you're opening Vikunja.

Another option would be to only do this if the provider has a logout url configured because then we can be more sure that the provider handles a logout properly and will not just redirect back to Vikunja.

surfingbytes:
But that redirect chain is exactly what should be there. If you have just one login provider, you logout, and you go back to that login provider. If user has SSO, of course he's brought to it. He'd need to logout in his login provider.

surfingbytes:
Requested changes are done

surfingbytes:
OIDC logout was added, is there anything else blocking this?

kolaente:
The new approach seems to set the key in the wrong place in the stack, which led to sometimes not being redirected properly. I've reimplemented that in https://github.com/go-vikunja/vikunja/pull/2968, will close this PR in favor of that one.

PR #2498, opened 2026-03-28
docs: correct task comment endpoint description and title

j-hugo:
Fixes the description and summary of the endpoint to get a task comment.


kolaente (approved):
Good catch! Thanks

PR #2499, opened 2026-03-28
feat: add MCP server

HarryEMartland:
I have been playing round in how to get Vikunja into my agents e.g zeroclaw (like openclaw). Rather than building  an integration into the agent or using a 3rd party MCP tool which I don't really trust I thought it would be good if Vikunja supported this natively. 

The server is disabled by default and runs on a separate port and uses the existing auth tokens.  Happy to take feedback on this.

No worries if this is not wanted I have learnt a fair bit along the way.

Further improvements could be made in the future to support other actions. I thought basic task management would  be a good start.

**Future Enhancements**
- [ ] Add more tools for models
- [ ] Define response schema
- [ ] Filter by index-able fields


kolaente:
Hey thanks for the PR!

I have two objections to this:

1. What makes this different than one of the existing Vikunja MCP servers? Just the fact that it would be a first-party approach?
2. Why use an mcp server and not teach your agent how to use a cli? (like `vja`)

I'm not totally against having this, but I'm not sure if this is a good way to integrate this.

HarryEMartland:
In my opinion having a first party MCP is really important from a security perspective. I hear lots about skills and packages having malware embedded in them. I would much rather a vikunja solution. If one is not provided someone will create one, as has happened. That inadvertently opens vikunja up for bad press if something does go wrong with the packages/skills.

My argument for a remote MCP VS a package is more about simplicity. The agent and the way it is running may limit what you can install. I run things in publicly built containers so installing things becomes a challenge unless I build my own image. That also applies for a skill using vja, however; a skill using a basic http fetch could be an option. 

As an appeal to authority Jira and GitHub both have remote MCP but I am sure there are just as many that use other methods.

I am by no means an expert in this area 😇.

kolaente:
Sounds good! I think we can move forward here, will do a review.

pkg/cmd/web.go
```diff
@@ -183,6 +184,18 @@ var webCmd = &cobra.Command{
 			log.Info("shutting down...")
 		}()
 
+		// Start MCP server if enabled
+		var mcpServer *mcp.ServerWrapper
+		if config.MCPEnabled.GetBool() {
+			mcpCfg := mcp.GetMCPConfig()
+			mcpServer = mcp.NewMCPServerWrapper()
+			go func() {
+				if err := mcpServer.RunHTTP(mcpCfg.Host, mcpCfg.Port); err != nil {
```
kolaente: Why run it on a separate port? This makes it harder to use for some deployments.

pkg/config/config.go
```diff
@@ -484,6 +488,11 @@ func InitDefaultConfig() {
 	PluginsEnabled.setDefault(false)
 	PluginsDir.setDefault(ResolvePath("plugins"))
 
+	// MCP
+	MCPEnabled.setDefault(false)
+	MCPHost.setDefault("127.0.0.1")
```
kolaente: This means it will bind to the server interface, making it inaccessible from anywhere but the server. If running in a container, that means it will only be accessible from inside that container, making it pretty much noch usable.

pkg/mcp/handlers.go
```diff
@@ -0,0 +1,462 @@
...
+
+	u, err := user.GetUserByID(session, apiToken.OwnerID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get user: %w", err)
+	}
+
+	return u, nil
+}
+
+func (s *Server) HandleGetTasks(params json.RawMessage, authToken string) (*mcp.CallToolResult, error) {
```
kolaente, line 66: To make this maintainable, we should generate the Handlers. We already have extensive swagger docs for all of these, that should make it pretty straight forward to generate.

Another way would be to create a wrapper like we're doing with the `web` package. Then registering new handlers for the MCP would be as easy as just adding new routes.

Check the routes registration to see what I mean.

pkg/mcp/handlers.go
```diff
@@ -0,0 +1,462 @@
...
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+type Server struct{}
+
+func NewMCPServer() *Server {
+	return &Server{}
+}
+
+func (s *Server) authenticate(token string) (*user.User, error) {
```
kolaente, line 38: We should also (only?) support the recently added oauth server: https://vikunja.io/docs/oauth-server/

MCPs are an end-user feature, adding an MCP should be as easy as "paste a URL, then login via browser", having to create an api token adds friction.

pkg/mcp/server.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/mark3labs/mcp-go/server"
+)
+
+type ServerWrapper struct {
+	httpServer *server.StreamableHTTPServer
+	handler    *Server
+}
+
+func NewMCPServerWrapper() *ServerWrapper {
+	srv := server.NewMCPServer("Vikunja", "1.0.0")
```
kolaente: Why not use the Vikunja version?

kolaente (changes requested):

pkg/config/config.go
HarryEMartland: resolved this  by moving to the api server

pkg/mcp/handlers.go
HarryEMartland: Totally agree with this. I have had a bit of a play tonight but not got too far. Ill keep digging

pkg/mcp/handlers.go
HarryEMartland: Yeah I like this ill have a play and see what I can reuse

HarryEMartland: Ok managed to refactor this out what do you think?

pkg/mcp/handlers.go
HarryEMartland: Looks like we will need dynamic client registration for this. My latest push has hacked this together but I think a bit more needs to be done to make it secure.

I noticed that there was an explicit rule to only allow vikuna prefixed clients. I think I can put something together which will keep this rule, but also allow registered clients. We can maybe have this as another config param to enable. Does this sound OK to you?

HarryEMartland: I have validated the client id and redirects are the same across the register and authorize endpoints

pkg/modules/auth/oauth2server/client.go
```diff
@@ -19,14 +19,21 @@ package oauth2server
...
 
 // ValidateRedirectURI checks that the redirect_uri uses a scheme starting with
 // "vikunja-". This allowlists only Vikunja native app schemes (e.g.
 // vikunja-flutter://callback) and rejects dangerous schemes like javascript:,
 // data:, http:, https:, etc.
-func ValidateRedirectURI(redirectURI string) bool {
-	u, err := url.Parse(redirectURI)
+func ValidateRedirectURI(req authorizeRequest, client *models.OAuthClient) bool {
+
+	if client != nil && strings.Contains(client.RedirectURIs, req.RedirectURI) {
```
HarryEMartland: I think this needs a bit more work. Ill try base 64 encode to stop , being abused

pkg/models/oauth_client.go
```diff
@@ -0,0 +1,79 @@
...
+// OAuthClient represents a dynamically registered OAuth 2.0 client.
+type OAuthClient struct {
+	ClientID     string    `xorm:"varchar(50) not null unique pk" json:"client_id"`
+	ClientName   string    `xorm:"varchar(255) not null" json:"client_name"`
+	RedirectURIs string    `xorm:"text not null" json:"redirect_uris"`
+	Created      time.Time `xorm:"created not null" json:"created"`
+}
+
+func (*OAuthClient) TableName() string {
+	return "oauth_clients"
```
kolaente, line 34: Do we really need to have this in a table? Why not use PKCE exclusively?

joebnb:
waiting for this feature to be merged

joebnb:
i want vikunja can hold a ai kanban to manage project develop

HarryEMartland:
Ill try get back on this in the next couple of days :)

pkg/models/oauth_client.go
HarryEMartland: Sorry for only just getting back on this. I have had a bit more of a read of https://www.rfc-editor.org/rfc/rfc7592#section-5 which states:

> If a client is deprovisioned from a server, any outstanding registration access token for that client MUST be invalidated at the same time.

My thinking was storing the client/id in the DB would allow us to delete the entry essentially invalidating the client stopping it from creating any more tokens.

If we did use PKCE exclusively then I think we would still need some persistence to invalidate it? But we could do the inverse where if it is in the DB then it is blocked? 

When I was initially working on this I had the register endpoint pretty much a no op other than generating a random client id. We could go back to that until we actually have the ability to invalidate the clients?

BenediktMiller:
Hi! 👋 Just wanted to check in on this - any updates on the status of this MR? It looks like a really useful feature and I'd love to see it land. Is there anything blocking the merge, or anything I can help with? Thanks for your work on this!

lerenn:
Interested by it too !

benpbo:
+1 Also interested in this!

sebt3:
I think we're a bunch interested in this :)

kolaente:
If you want something to use right now, use https://github.com/go-vikunja/vikunja/tree/main/veans

Commenting +1 on this PR will not get it landed faster.

kolaente:
In other news, I started exploring a different implementation in https://github.com/go-vikunja/vikunja/pull/2800 that exposes the endpoints differently. Since this PR here seems to be abandoned, I'll close this one in favor of the other one.

Please ping if you want to continue working on this!

aichholzer:
Perhaps a pinch too late for this thread, but here's my take on how an agent should be able to talk to Vikunja.
https://github.com/aichholzer/altiplano

Tiny & fast. Works like a charm — I use it both at home and work.

PR #2502, opened 2026-03-30
feat: rename ServiceJWTSecret to ServiceSecret with deprecation

kolaente:
## Summary
This PR renames the `ServiceJWTSecret` configuration key to `ServiceSecret` to better reflect its purpose as a general service secret used for JWT token signing. The old key is deprecated but still supported for backward compatibility.

## Key Changes
- Introduced new `ServiceSecret` config key (`service.secret`) as the primary configuration option
- Marked `ServiceJWTSecret` (`service.JWTSecret`) as deprecated with a migration path
- Added deprecation logic in `InitConfig()` that automatically migrates values from the old key to the new key if the old key is explicitly set
- Updated all code references to use `ServiceSecret.GetString()` instead of `ServiceJWTSecret.GetString()`:
  - JWT token signing in auth module
  - JWT token parsing in test utilities and middleware
  - Configuration validation in doctor module
- Updated configuration schema documentation to reflect the new primary key and deprecation status

## Implementation Details
- The migration logic checks if `ServiceJWTSecret` differs from `ServiceSecret` and copies the value while logging a deprecation warning
- Both keys are initialized with the same default random value to ensure backward compatibility
- All internal code now uses the new `ServiceSecret` key, while the old key remains functional for existing configurations

https://claude.ai/code/session_01WxMqCGturowwjgGfRMUW7J

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -638,9 +638,13 @@ func InitConfig() {
 
 	// Deprecation: migrate service.JWTSecret → service.secret only when the
 	// user has not explicitly set service.secret (so the new key takes precedence).
-	if ServiceJWTSecret.GetString() != "" && !viper.IsSet(string(ServiceSecret)) {
-		log.Warning("config: service.jwtsecret is deprecated and will be removed in a future release. Please use service.secret instead.")
-		ServiceSecret.Set(ServiceJWTSecret.GetString())
+	if ServiceJWTSecret.GetString() != "" {
+		if viper.IsSet(string(ServiceSecret)) {
+			log.Warning("config: both service.secret and service.jwtsecret are set. Using service.secret. Please remove service.jwtsecret, it is deprecated and will be removed in a future release.")
+		} else {
+			log.Warning("config: service.jwtsecret is deprecated and will be removed in a future release. Please use service.secret instead.")
+			ServiceSecret.Set(ServiceJWTSecret.GetString())
+		}
 	}
 
 	if _, err := url.ParseRequestURI(AvatarGravatarBaseURL.GetString()); err != nil {


config-raw.json
```diff
@@ -4,10 +4,15 @@
             "key": "service",
             "children": [
                 {
-                    "key": "JWTSecret",
+                    "key": "secret",
                     "default_value": "\u003cjwt-secret\u003e",
                     "comment": "This token is used to verify issued JWT tokens.\nDefault is a random token which will be generated at each startup of Vikunja.\n(This means all already issued tokens will be invalid once you restart Vikunja)"
```
kolaente: This needs an update. We will use the secret for more things in the future, not just JWT

pkg/config/config.go
```diff
@@ -333,6 +334,7 @@ func InitDefaultConfig() {
 	}
 
 	// Service
+	ServiceSecret.setDefault(random)
 	ServiceJWTSecret.setDefault(random)
```
kolaente: This should not be set, will be handled by the `ServiceSecret.setDefault`. The jwt secret should be emtpy by default

pkg/config/config.go
```diff
@@ -637,11 +636,11 @@ func InitConfig() {
 
 	readConfigValuesFromFiles()
 
-	// Deprecation: if service.JWTSecret is explicitly set but service.secret is not,
-	// migrate the value and warn the user.
-	if ServiceJWTSecret.GetString() != ServiceSecret.GetString() {
+	// Deprecation: if service.JWTSecret is explicitly set, migrate its value
+	// to service.secret and warn the user.
+	if jwtSecret := ServiceJWTSecret.GetString(); jwtSecret != "" {
```
kolaente: This seems overly complicated

PR #2508, opened 2026-03-31
fix: set Content-Length on JSON API responses to fix macOS curl piping

AmirL:
## Summary
- Added `ContentLengthJSONSerializer` middleware that marshals JSON to bytes first, sets `Content-Length` header, then writes the response
- This prevents Go's default chunked transfer encoding which macOS curl mishandles when piping to another program (e.g. `curl | jq` produces empty stdin)
- Added web test verifying Content-Length header is present on list endpoints

## Red-Green Evidence

**RED (without fix, middleware not registered in routes.go):**
```
=== RUN   TestContentLengthOnListEndpoints
=== RUN   TestContentLengthOnListEndpoints/ReadAll_response_has_Content-Length
    content_length_test.go:52:
        Error Trace: content_length_test.go:52
        Error:       Should NOT be empty, but was
        Messages:    Content-Length header must be set on list responses to prevent macOS curl piping issues
    content_length_test.go:55:
        Error Trace: content_length_test.go:55
        Error:       Received unexpected error:
                     strconv.Atoi: parsing "": invalid syntax
--- FAIL: TestContentLengthOnListEndpoints (0.04s)
    --- FAIL: TestContentLengthOnListEndpoints/ReadAll_response_has_Content-Length (0.04s)
FAIL
```

**GREEN (with fix applied):**
```
=== RUN   TestContentLengthOnListEndpoints
=== RUN   TestContentLengthOnListEndpoints/ReadAll_response_has_Content-Length
--- PASS: TestContentLengthOnListEndpoints (0.04s)
    --- PASS: TestContentLengthOnListEndpoints/ReadAll_response_has_Content-Length (0.04s)
PASS
ok  	code.vikunja.io/api/pkg/webtests	0.756s
```

## Full test and lint output

**Web tests (mage test:web):**
```
ok  	code.vikunja.io/api/pkg/webtests	15.490s
```

**Lint (mage lint):**
```
0 issues.
```

### DIFF
--- a/pkg/routes/content_length_middleware.go
+++ b/pkg/routes/content_length_middleware.go
@@ -1,0 +1,64 @@
+// Vikunja is a to-do list application to facilitate your life.
+// Copyright 2018-present Vikunja and contributors. All rights reserved.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package routes
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/labstack/echo/v5"
+)
+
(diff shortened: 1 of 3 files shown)


kolaente:
This is the wrong place to fix this. This is a macOS issue, not a Vikunja issue. I'm not fixing the bugs of multi-billion dollar companies in an open-source project.

There are more workarounds for you: use `curl --no-buffer` or `curl -N`, or installing a proper curl via Homebrew. wget et al. handle this just fine as well.

AmirL:
Hi @kolaente 
This is content length, not actually macOS bug. It just easier to see in MacOs, but could be in other places as well. 
Everything else works with curl well, and it was frustrating that vikinja doesn’t work.

PR #2509, opened 2026-03-31
feat: add bucket selector to task detail view

AmirL:
## Summary
- Adds a bucket selector component to the task detail sidebar, allowing users to change a task's Kanban bucket without navigating back to the board
- Supports projects with multiple Kanban views (grouped display)
- Hides the selector for projects without Kanban views
- Fixes a bug where changing the bucket would cause task detail sidebar UI elements to disappear (the API response task lacks expanded relations, so only changed fields are now propagated)


kolaente:
How is this different from? https://github.com/go-vikunja/vikunja/pull/2233

AmirL:
Not much actually, just  design. didn’t notice the PR, when started this. Just was annoyed to move task between buckets by mouse :)

kolaente:
I've just merged https://github.com/go-vikunja/vikunja/pull/2233 - closing this one because it's obsolete now.

If you want a different UX, happy to talk about that in an issue :)

Issue #1483, opened 2025-09-11
Default my open Tasks filter


NeptunedGmbH:
### Description

I'm managing a team of very much non technical people and I know for some of them it's an unachievable hurdle to create this filter themself. So a default saved filter 
`assignees = {USERNAME} && done = 0`
would be very helpful.
Otherwise sending a webhook for each newly registered user would work for me as well since I could just create the filter thru the API then.

PR #2533, opened 2026-04-03
feat(filters): generate open task saved filter on user creation

gabeklavans:
During user creation, in addition to creating the "Inbox" project, also create a saved filter for the user that shows the open tasks assigned to them. This provides utility for new users to get a quick view of their tasks with no extra setup, as well as gives them an example of a predefined saved filter so they can quickly get up to speed with the syntax and UI layout.

Quick, band-aid solution for #1483 

Provides a "good enough" solution for https://community.vikunja.io/t/add-default-filter-for-tasks-assigned-to-current-logged-in-user/4417

## Considerations

- could be gated with a config value, something like `defaultsettings.new_user_create_filter`
- this won't create the filter for existing users

### DIFF
--- a/pkg/models/project.go
+++ b/pkg/models/project.go
@@ -760,6 +760,15 @@ func getRawProjectsForUser(s *xorm.Session, opts *projectOptions) (projects []*P
 	return allProjects, len(allProjects), totalItems, err
 }
 
+func CreateDefaultSavedFiltersForUser(s *xorm.Session, u *user.User) error {
+	sf := &SavedFilter{
+		Title:   "My Open Tasks",
+		Filters: &TaskCollection{Filter: fmt.Sprintf("done = false && assignees = %s", u.Username)},
+	}
+
+	return sf.Create(s, u)
+}
+
 func getSavedFilterProjects(s *xorm.Session, doer *user.User, search string) (savedFiltersProjects []*Project, err error) {
 	savedFilters, err := getSavedFiltersForUser(s, doer, search)
 	if err != nil {
@@ -1105,6 +1114,10 @@ func RegisterUser(s *xorm.Session, u *user.User) (*user.User, error) {
 	}
 
 	if err := CreateNewProjectForUser(s, newUser); err != nil {
+		return nil, err
+	}
+
+	if err := CreateDefaultSavedFiltersForUser(s, newUser); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


kolaente:
I'm really torn on this because it affects all users, even those who will never need the filter. On the other hand, having a way for users to quickly see their assigned tasks has been requested a few times now.

The main way to solve this woul obviously be to make filtering easier, but that's a whole different can of worms.

pkg/models/project.go
```diff
@@ -651,6 +651,15 @@ func getRawProjectsForUser(s *xorm.Session, opts *projectOptions) (projects []*P
 	return allProjects, len(allProjects), totalItems, err
 }
 
+func CreateDefaultSavedFiltersForUser(s *xorm.Session, u *user.User) error {
+	sf := &SavedFilter{
+		Title:   "My Open Tasks",
```
kolaente, line 765: This should be translated in the frontend, in the same way we're translating the "Inbox" project title

kolaente (changes requested):

gabeklavans:
> I'm really torn on this because it affects all users, even those who will never need the filter. On the other hand, having a way for users to quickly see their assigned tasks has been requested a few times now.

I can certainly gate this behind a config that is off by default.

> The main way to solve this woul obviously be to make filtering easier, but that's a whole different can of worms.

Even if filtering were easier, I still think there's value in an automatic or "one click" way to get very common/requested filters for new users. Maybe fundamentally this could be a mechanism for globally saving/sharing filters and then having some default globally saved ones that can be loaded into a user's saved filter collection. But as you say, that's a whole can of worms.

kolaente:
After thinking about it again, let's add this. Please address the review comment and then we're good to merge this.

I'd not add a config because that just adds one more config flag, and we already have many. If users don't want the Filter, they can delete it. And since this will only affect new users, it will probably be fine.

gabeklavans:
I think I did that right. Let me know if you want me to rebase to the main branch

kolaente (approved):
Looks good! Thanks.

gabeklavans:
I'm trying to figure out why [this test failed](https://github.com/go-vikunja/vikunja/actions/runs/24127034241/job/70395052223) to take this PR out of the merge queue. It looks to me that it's just reading the users in the wrong order... I'm not sure if it has anything to do with this PR's changes. I'll try rebasing to main.

gabeklavans:
on rebasing to latest main, I noticed the function for creating the default inbox project was removed/moved

should I move the function for creating this default view, as well?

kolaente:
> should I move the function for creating this default view, as well?

Yes, it should move as well. Took the liberty to do that and rebased on main while at it. Ready to merge now!

PR #2588, opened 2026-04-10
fix: Backend hard coded api base

### LINKED ISSUE #2657: Token refresh fails when hosting in a subdirectory
### Pre-submission checklist

### Description

The refresh token cookie is hard-coded to the `/api/v1/user/token/refresh` path:
https://github.com/go-vikunja/vikunja/blob/56aef0369766268a6bf9a5c72f6e2b2ff0155c10/pkg/modules/auth/auth.go#L51

When running Vikunja in a subdirectory (https://vikunja.io/docs/running-vikunja-in-a-subdirectory/), this makes the token refresh call always fail with "No refresh token provided.". The cookie is never sent to `<sub-path>/api/v1/user/token/refresh` because of the path mismatch.

I guess one possible fix is to build the cookie path dynamically based on the requests' path? Or, alternatively, use the same env var `VIKUNJA_FRONTEND_BASE` to build the path?

### Vikunja Version

v2.3.0

### Browser and version

_No response_

No

### Screenshots

_No response_

mdrkrg:
Porting stale changes from #2436, this mainly focus on the backend.

Resolves https://github.com/go-vikunja/vikunja/issues/2657

## What changed

- `auth/auth.go`: Compute refresh token cookie base path from `service.publicurl` rather than hard coding the base as `/api/v1`.
- `api/v1/docs.go`: Build `redocHTML` as template, and replace with `apiBase + api/v1/docs.json` and `apiBase + api/v1/docs/redoc.standalone.js`.

### DIFF
--- a/pkg/modules/auth/auth.go
+++ b/pkg/modules/auth/auth.go
@@ -19,6 +19,7 @@ package auth
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


kolaente (approved):
Thanks!

PR #2592, opened 2026-04-10
feat: show sub-project tasks in parent project views 🤖🤖🤖

mokshasoft:
When viewing a parent project, tasks from all descendant sub-projects are now displayed alongside the parent's own tasks. This enables a unified view of work across an entire project hierarchy.

## Summary

- Parent project views now display tasks from all descendant sub-projects alongside their own tasks
- Task positions are stored at root project view level for consistent ordering across the hierarchy
- Centralized position management via `StoreTaskPosition` and `ReplaceAllPositionsForView`

## Technical details

- `getDescendantProjectIDs` traverses project tree using BFS
- Replaced recursive CTEs with iterative Go loops to avoid xorm session state corruption
- Root view ID resolution happens before query building in task search
- `setTaskInBucketInViews` returns positions for deferred storage

## Test plan

- [x] Unit tests added for hierarchy traversal (`pkg/models/hierarchy_test.go`)
- [x] Position storage tests verify root view resolution
- [x] `mage test:feature` passes
- [x] `mage test:web` passes
- [x] `mage lint` passes (no new issues)


jonasseglare:
I tried it out a bit in the preview deployment with a top project and two sub projects. This is what it looks like in the list view:

Screenshot 2026-04-10 at 10 21 38

and here is a Gantt view of the same tasks.

Screenshot 2026-04-10 at 10 21 51

It looks quite neat.

mokshasoft:
I noticed that there is a problem with the kanban view, and each sub-project can have their own columns. So I don't know yet how to solve this.

kingkongola:
Oh this Is exactly what I need! will follow the progress closely! ❤️

mokshasoft:
**Resolution:** Kanban views now only show tasks from their own project level, not from child projects.

Each sub-project can define its own kanban columns (buckets), and these don't translate across the hierarchy. If we showed child tasks in a parent's kanban, each task would need independent bucket associations at every level, creating confusing multi-state situations (e.g., task shows "Done" in parent but "Dev" in child).

List, table, and gantt views continue to show tasks from all descendant projects since they don't have per-view state like buckets.

**Migration:** The migration deletes orphaned positions from child project views, then adds positions for child project tasks in root project views (excluding kanban). Existing user-ordered positions are preserved - new tasks from child projects are appended after them, ordered by priority (highest first), due date (earliest first, nulls last), then creation date.

mds08011:
Thanks, I’m excited to see this!  This seems like a nice clean solution.  Just flagging a similar effort for reference: https://github.com/go-vikunja/vikunja/pull/2272

mokshasoft:
> Thanks, I’m excited to see this! This seems like a nice clean solution. Just flagging a similar effort for reference: #2272

Thank you! Don't know why I didn't see this.. but maybe there are things to learn from both patches.

mokshasoft:
> > Thanks, I’m excited to see this! This seems like a nice clean solution. Just flagging a similar effort for reference: #2272
> 
> Thank you! Don't know why I didn't see this.. but maybe there are things to learn from both patches.

One difference between these patch-sets is that I specifically did not want to add a setting for this feature.

kolaente:
Sorry for taking so long to reply here!

Since we now have a bunch of PRs implementing some version of this, I'll close this one in favor of https://github.com/go-vikunja/vikunja/pull/2939. This here is not doable as-is since we should not make this a default and rather put it behind a setting as [outlined previously](https://github.com/go-vikunja/vikunja/pull/2272#discussion_r2835304401).

PR #2647, opened 2026-04-16
fix(desktop): support Ctrl and mouse-wheel zoom shortcuts

### LINKED ISSUE #2623: Desktop app: no way to zoom UI (Ctrl+scroll / Ctrl+= not supported)
### Pre-submission checklist

### Description
The Vikunja desktop app currently has no way to zoom the UI. Standard Electron-style shortcuts (Ctrl + scroll wheel, Ctrl + Plus/Minus) do nothing, and there is no menu option for it either.

### Expected behaviour
Some way to scale the UI. Any of the following would work:
- Ctrl + mouse wheel to zoom in/out
- Ctrl + Plus / Ctrl + Minus / Ctrl + 0 to reset
- A zoom control accessible from the tray icon menu

### Related niceties
- **Persistent zoom level:** the chosen zoom should survive app restart, not reset every launch.
- **Menu entries for discoverability:** surface the zoom controls in the View menu (and/or tray menu) so the shortcuts aren't hidden.
- **Respect system DPI scaling:** on high-DPI displays the app should honor the OS scale factor as the baseline before any user zoom is applied.

Useful on high-DPI displays and for accessibility.

Tokra110:
## Summary

Adds keyboard and mouse-wheel zoom shortcuts to the Electron desktop app, with the chosen level persisted across restarts. Fixes #2623.

- **Ctrl + Plus / Ctrl + Minus / Ctrl + 0** via `webContents.before-input-event`, `event.preventDefault()` so the page doesn't double-handle them.
- **Ctrl + mouse wheel** via the `zoom-changed` webContents event.
- Zoom clamped to `[-7, +7]` (Chromium's own range, roughly 28% to 358%).
- Persisted to `zoom.json` in `app.getPath('userData')` and re-applied on every `did-finish-load`, since Electron resets zoom on page reloads.

No new menu/tray entries, following maintainer feedback on the issue that shortcuts should "just work" rather than surfacing UI for them.

## Test plan

- [x] Ctrl+= / Ctrl+- / Ctrl+0 change zoom as expected
- [x] Ctrl + mouse wheel zooms smoothly in both directions
- [x] Zoom persists across a full quit + relaunch
- [x] Zoom clamps at bounds under rapid repeated input
- [x] Tested on Fedora 43, KDE Plasma 6, Wayland

### DIFF
--- a/desktop/main.js
+++ b/desktop/main.js
@@ -10,6 +10,7 @@ const {
 	screen,
 } = require('electron')
 const path = require('path')
+const fs = require('fs')
 const express = require('express')
 const portInUse = require('./portInUse.js')
 const oauth = require('./oauth.js')
@@ -23,6 +24,11 @@ const SAFE_PROTOCOLS = new Set([
 
 const QUICK_ENTRY_WIDTH = 680
 const QUICK_ENTRY_COLLAPSED_HEIGHT = 56
+
+const ZOOM_MIN = -7
+const ZOOM_MAX = 7
+const ZOOM_STEP = 0.5
+const ZOOM_CONFIG_FILE = 'zoom.json'
 
 const BASE_WEB_PREFERENCES = {
 	nodeIntegration: false,
@@ -52,6 +58,7 @@ let isQuitting = false
 let pendingDeepLinkUrl = null
 let pendingApiUrl = null
 let currentShortcut = null
+let zoomLevel = 0
 const DEFAULT_QUICK_ENTRY_SHORTCUT = 'CmdOrCtrl+Shift+A'
 const launchedWithQuickEntry = process.argv.includes('--quick-entry')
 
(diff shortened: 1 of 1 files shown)


kolaente:
Does Electron really not have this built in?

Tokra110:
From what I could find, no. Electron exposes `webContents.setZoomLevel`, `setZoomFactor`, and a `zoom-changed` event for Ctrl+scroll, but doesn't bind the keyboard shortcuts for you (Chromium's zoom keybindings are in Chrome's UI layer, not the content shell Electron embeds). A few other Electron apps I looked at (VS Code, Slack) wire their own `before-input-event` handlers the same way.

Happy to go a different route if you know of a simpler approach I missed.

kolaente:
@Tokra110 I've had my bot analyze this and its response sounds solid, please take a look.

Tokra110:
Thanks for the analysis! I looked into using menu roles (`{role: 'zoomIn'}` etc.) but they don't quite fit here:

1. **Persistence**: Roles call `setZoomLevel` internally without emitting an event we can hook into, so there's no way to persist the level to `zoom.json` after a role-triggered zoom. We'd need custom `accelerator` + `click` handlers anyway to call `saveZoomLevel`, which is structurally the same as `before-input-event`.

2. **Numpad support**: `before-input-event` checks `input.key`, which gives `'+'`/`'-'` regardless of whether the user presses the main keyboard or numpad. Menu accelerators would need separate hidden entries (`CommandOrControl+numadd`, `CommandOrControl+numsub`) to match.

I did adopt the suggestion to drop the `[-7, +7]` clamp since Chromium caps zoom levels internally anyway. That simplifies the code without changing behavior.

kolaente:
> I did adopt the suggestion to drop the [-7, +7] clamp since Chromium caps zoom levels internally anyway. That simplifies the code without changing behavior.

Sounds good, please go ahead!

Tokra110:
Already pushed — commit 283646d dropped the clamp before your last message. Nothing else to do on our side, the PR should be ready for another look whenever you have time.

kolaente (approved):
Thanks!

PR #2656, opened 2026-04-18
feat: support all days of the week as first day of the week

xela-zone:
This PR adds support for all days of the week as the "first day of the week" in user settings.

**Changes:**
- **Backend:** Fixed `week_start` validation in `user_settings.go` to use `range(0|6)` (previously allowed an invalid value).
- **Frontend:** Added missing Tuesday-Saturday translation keys and updated the settings dropdown options.
- **Testing:** Added a permanent E2E test case in `settings.spec.ts` using the modern `getByLabel` pattern to verify persistence across reloads.
- **Docs:** Added `GEMINI.md` symlink.

   Verified with `mage test:e2e` and full linting suite.
   
   idea and initial editing done by me, e2e testing and range validation found and done by Gemini


frontend/tests/e2e/user/settings.spec.ts
```diff
@@ -109,4 +109,33 @@ test.describe('User Settings', () => {
...
+
+		// The save button only appears when isDirty becomes true (settings changed)
+		// We use the data-cy here as it's the standard in this project, 
+		// though it might require window.TESTING=true in some environments.
+		const saveButton = page.locator('[data-cy="saveGeneralSettings"]')
+		await expect(saveButton).toBeVisible({timeout: 10000})
+		await saveButton.click()
+
+		await expect(page.locator('.global-notification')).toContainText('Success')
+
```
kolaente, line 129: This test should assert the API request that actually stores the updated setting

kolaente (changes requested):

GEMINI.md
kolaente, line 1: As far as I understand it, Gemini cli supports AGENTS.md which we already have: https://github.com/google-gemini/gemini-cli/issues/25205

So this would be rather redundant.

Did you have issues with Gemini cli and the AGENTS.md?

kolaente (commented):
Thanks for the PR! Some nits, but looks good overall

frontend/tests/e2e/user/settings.spec.ts
xela-zone: So not anything to do with the frontend, right?

GEMINI.md
xela-zone: Huh. Through a set of insane curcumstance, I used like three different versions of Gemini CLI when working on this. Going forward, it would definitely be cleaner to not have a dedicated file per agent, I'll take this commit out of my branch.

frontend/tests/e2e/user/settings.spec.ts
kolaente: I mean it should assert that the request that was send to the api to store the setting was sent and contains the correct value.

Additionally, we should also have a test that sets the setting via a factory and then asserts the frontend uses that setting to change what is shown in the frontend (not the settings, but the place where the setting is actually used).

xela-zone:
changes made - gemini once again hleped me write the tests. 

it mentioned (as part of a convo afterward, and during) that it fixed a 'shared-object mutation in useFlatpickrLanguage', and enabled window.TESTING to allow data-cy attributes to work in e2e tests. 

seeing the mention of the flatpickrLanguage potential shared state bug, it might be a good idea to add a test for that, but i am not one to write tests (the only reason there's one on this commit is gemini wrote it when i was trying to test the project, and i thought 'hay that's a free test, why not upstream that too'

frontend/tests/support/authenticateUser.ts
```diff
@@ -12,6 +12,7 @@ export async function setupApiUrl(page: Page) {
 	await page.addInitScript(({apiUrl}) => {
 		window.localStorage.setItem('API_URL', apiUrl)
 		window.API_URL = apiUrl
+		window.TESTING = true
```
kolaente: This should not be required, the tests in CI use dev builds which set the flag

frontend/tests/e2e/user/settings.spec.ts
```diff
@@ -94,19 +94,56 @@ test.describe('User Settings', () => {
 
 		// Wait for the settings page to be fully loaded and the input to be enabled
 		const nameInput = page.locator('.general-settings input.input').first()
-		await expect(nameInput).toBeVisible({timeout: 10000})
+		await expect(nameInput).toBeVisible({timeout: 15000})
```
kolaente: why the increased timeout?

frontend/tests/support/updateUserSettings.ts
```diff
@@ -2,14 +2,19 @@ import type {APIRequestContext} from '@playwright/test'
 import {objectToSnakeCase} from '../../src/helpers/case'
 
 export async function updateUserSettings(apiContext: APIRequestContext, token: string, settings: any) {
-	const apiUrl = process.env.API_URL || 'http://localhost:3456/api/v1'
+	let apiUrl = process.env.API_URL || 'http://localhost:3456/api/v1'
+	apiUrl = apiUrl.replace(/\/+$/, '')
```
kolaente: This seems like an unrelated change as well

kolaente (changes requested):
The implementation is good, the tests need some work.

frontend/tests/e2e/user/settings.spec.ts
xela-zone: Increased to match other test in file

xela-zone: Oh I see, Gemini was stupid and I didn't review it's output well enough :/ sorry for the wasted time with that review

frontend/tests/support/authenticateUser.ts
xela-zone: turns out somthing like it is needed - atleast for running the tests locally. 
See 70582b94f4610dd7011ef74d02f6ca3fffc093ea for details

kolaente (approved):
Thanks!

Issue #1383, opened 2025-09-02
More Quick Actions


xeruf:
### Description

Currently I can favorite a task from the overview - if I could also assign a user, add labels, set a due date - such common operations - that would be very handy:

<img width="1230" height="140" alt="Image" src="https://github.com/user-attachments/assets/df883ced-16cf-4156-9bc9-1518ebb49fc7" />

These could be conditional, i.e. only show user assignment action when no user is assigned.

### Which alternatives did you consider using instead?

This would of course also be very doable in the table view, which in my eyes will only really become useful once it is editable just like an actual spreadsheet, making vikunja very intuitive for people coming from spreadsheet task management.

Currently a lot of clicks and loads if I added some tasks and want to modify them quickly

kolaente:
Do you think this would be solved with a right click context menu?

Adding icons for everything right in the task bar would make it overcrowded very quickly.

xeruf:
maybe a mix? most common operations in the right, more with right click? because right click does not work on all interfaces well.

the tool could even learn common operations, cause I think an extra config option would be a bit silly

for example 
- in a project without collaborators the assignee option is not needed
- a task with a label may not need an add label quick access button
- setting prio option if there is no prio
- setting color option if there is no color AND any other task has a color

cause this is for initial quick classification, detailed edits can still be made later

acwilsondev:
I have built a prototype for a quick actions menu. In my design, this is not a dynamic menu, but offers a default set of edits. I am happy to polish my prototype and submit a PR.

Happy to bring this as a new issue as well, since it is slightly different from the OP.

<img width="1568" height="573" alt="Image" src="https://github.com/user-attachments/assets/3060ee7e-5fa9-4da0-b4b3-5eb33a2056a1" />

PR #2663, opened 2026-04-20
feat(frontend): inline quick-add fields for Kanban task creation

Tokra110:
## Summary

Adds configurable inline fields to the Kanban add-task form so users can
set assignees, dates, priority, labels, reminders, color, and progress
without opening the full task editor.

- New user setting (Settings > Appearance > "Inline fields when adding a
  task") with drag-to-reorder and an "Add field" picker using the
  existing Dropdown component
- Self-contained `InlineQuickAddFields` component with chip bar and
  popup editors for each field type
- Submit button (checkmark) on the task input so tasks can be saved
  without pressing Enter
- Hover clear button (×) on each chip to reset a field value
- Defaults to assignee + due date for new users

Related to #1383 (quick actions from views), though that issue focuses
on editing existing tasks in list view rather than inline creation in
Kanban.

## Test plan

- [x] Enable/disable fields in Settings > Appearance, verify they appear/disappear in Kanban add-task
- [x] Drag to reorder fields in settings, verify chip order matches in Kanban
- [x] Set each field type (assignee, due/start/end date, priority, labels, reminder, color, progress) and create a task
- [x] Verify the created task has the correct field values
- [x] Hover a set chip and click × to clear, verify it resets
- [x] Verify dark mode works correctly for settings dropdown and Kanban chips
- [x] Test with no setting saved (new user) — should default to assignee + due date


Tokra110:
Closing in favor of a broader unified approach that I'm reworking.

PR #2666, opened 2026-04-20
feat(editor): add tiptap/markdown incl. markdown editor 🤖🤖🤖

### LINKED ISSUE #2196: Tiptap bidirectional markdown support is now available
### Pre-submission checklist

### Description

This request addresses https://community.vikunja.io/t/markdown-as-first-class-citizen/2975

Tiptap now offers bidirectional Markdown support https://tiptap.dev/docs/editor/markdown (`@tiptap/markdown`) and this can help solve all the issue for the discussion above without much pain and different editor.

I understand the reasoning for it being removed before and my suggestion now is if implemented to be optional feature in under `General Settings`. If you enable Markdown support you get bidirectional markdown preserving the syntax regardless if you edit it with a CalDAV client that expect markdown or the web editor. Both will be happy and able to render.

### Senario 1 - Creating Description with the web UI editor (`tiptap`)

**Workflow**: Create a task in the WebUI -> Sync it with CalDAV / fetch from api (they expect markdown)

Example content in the WebUI:

Image

Raw result:

```html
<h1>Welcome to the Markdown Demo</h1><p>This demo showcases <strong>bidirectional</strong> markdown support in Tiptap with extended features.</p><h2>Features</h2><ul><li><p><strong>Bold text</strong> and <em>italic text</em></p></li><li><p><code>inline code</code> and code blocks</p></li><li><p><a target="_blank" rel="noopener noreferrer nofollow" href="https://tiptap.dev">Links</a></p></li><li><p>Lists and more!</p></li></ul><p></p>
```
What the CalDAV client (JTXboard) sees after sync:

Image

### Senario 2 - Creating task from CalDAV client

**Workflow**: Create a task on a CalDAV client (JTXboard) -> Sync it with Vikunja

Example rendered markdown content for the new task in JTXboard:

Image

Raw markdown content:

```md
# Welcome to the Markdown Demo

This demo showcases **bidirectional** markdown support in Tiptap with extended features.

## Features

- **Bold text** and *italic text*
- `inline code` and code blocks
- [Links](https://tiptap.dev)
- Lists and more!
```

What you see in Vikunja WebUI after sync:

Image

None exist

cheriimoya:
- Adds @tiptap/markdown extension (requires upgrading all @tiptap/* from 3.17.0 → 3.22.4)
- New user setting useMarkdownEditor, off by default, fully backward-compatible
- When enabled: editor reads/writes markdown instead of HTML, fixing CalDAV interop
- No separate editor component, the existing TipTap editor switches mode via getMarkdown()/contentType: 'markdown'
- existing tasks with HTML descriptions won't auto-convert (i wrote a migration script https://github.com/cheriimoya/vikunja/tree/feat/html-to-markdown-migrator)

resolves #2196


kolaente:
Which problem does this solve that's not solved by the markdown shortcuts or the markdown conversion on paste?

nomorsug:
The issue is explained in #2196, markdown shortcuts or the markdown conversion on paste are not relevant

cheriimoya:
its solving the issue of me not being able to read my vikunja tasks on my phone which i have connected to vikunja via caldav and the tasks.org app. so this:



renders to this:



instead of that:

kolaente:
That should be fixed in the caldav layer, not in the frontend.

nomorsug:
@kolaente I feel like you don't understand the issue. It does affect **not only caldav but also REST** and is caused by the frontend - the `tiptap` editor does not preserve by default the original formatting and will always output HTML. Please take a look again at #2196. This is fine for the people that don't use vikunja outside the web client but breaks the usage outside the web ui.

Given that the issue is caused by the frontend breaking the formatting, it makes no sense to try to fix it somewhere else when you can prevent it  to begin with. This is why tiptap has added `@tiptap/markdown` to deal with non-native DSL.

Of course you can go this route, but you need intermediate step to check for the format and convert to for example again markdown which will be error prone and will not always preserve the original...

kolaente:
Consider this scenario: Your teammate does not use caldav at all, only accesses Vikunja via the web UI. They don't have the setting this PR introduces enabled. Then they edit a task that you have access to - it shows up as html for you in caldav. The only way to fix this is for you to open the task in Vikunja and edit and save the description.

I know that converting markdown back and forth between html has it's own set of issues, but only doing this in the web editor does not fix the underlying problem, unless we'd make CalDAV the only way of editing text in Vikunja.

nomorsug:
Don't forget the REST API. It's the same as with CalDAV. Have not considered this, but I agree, it will be a mess if used by multiple people with different settings.

In that case it will need to be done by intermediate layer indeed. Maybe on the CalDAV setup page there will be an option to enable markdown compatibility, and same with a parameter for the REST API. There are a lot of pitfalls to be avoided but it's doable. (the multiple markdown flavors etc.)

cheriimoya:
Fair point which I didn't have on my mind for multi-user setups, as for single-user instances this concern doesn't apply at all. I run Vikunja solo, and I personally know three other people who do the same. I'd guess there are quite a few of us out there.

Also, fixing this at the CalDAV layer would probably come with its own implications. I can see the CalDAV rewrite is a significant undertaking and the markdown conversion would naturally depend on it landing first, which means it could be a while. Keeping it in the frontend as an opt-in setting to me feels like the lower-risk approach for now, and a way to help affected users in the meantime.

This issue has clearly been a pain point for a while. Beyond #2196, there's a bunch of community threads [Markdown as first class citizen](https://community.vikunja.io/t/markdown-as-first-class-citizen/2975), [Documents interoperability](https://community.vikunja.io/t/documents-interoprability-task-description-and-other-content/3324) and [`<p></p>` added into description field](https://community.vikunja.io/t/p-p-added-into-description-field/1975) where the same HTML-in-CalDAV-clients problem has been open since February 2024 with people still asking for the implementation earlier this year.

Since this PR doesn't change any default behavior and directly addresses a long-standing problem for a real segment of users, would it really be such a bad thing to get this into Vikunja?

kolaente:
Hey sorry for not replying, seems like this slipped through in the pile of PRs.

Unfortunately, we can't ship this if the two points from earlier are not solved. Solving this in the frontend in the way that you're proposing is simply the wrong place. I'll spend some time thinking about how to solve this properly.

kolaente:
There is now https://github.com/go-vikunja/vikunja/pull/2986 which adds markdown support to the rest api (opt in) and caldav (always on). I'll close this PR in favor of the other one. Thanks for raising this.

PR #2676, opened 2026-04-21
Fixing issues with subtasks in filtered views. 1. Subtasks that match…

### LINKED ISSUE #1786: Project "New Saved Filter" View | Duplicates Subtasks
### Description

If you create a view that uses a query like:

`project = FirstProject || project = SecondProject || project = ThirdProject`

All subtasks of tasks will be duplicated in the view.

That is:

__ parent1
____ child1
____ child2
__ child1
__ child2

Will be what is displayed. This makes a confusing view when you just want to see a single instance of each subtask properly formatted below the parent task.

### Vikunja Version

Version: v1.0.0-rc2

### Browser and version

Edge

Yes

### Screenshots

Image

### LINKED ISSUE #2646: Subtasks not shown in custom filters if parent does not match
### Pre-submission checklist

### Description

There is one closed report, which does not resolve this problem:  #2494 

Custom filters do not show subtasks, if their criteria matches, but the parents criteria does not match.

## Expected behavior
All tasks which match the filter criteria of a smart filter are displayed. All tasks which do not match, are not displayed. Regardless of any relations.

## Replication
1. Add a filter "DO NOW!!!" with the criteria: `done = false && priority = 5`
2. Create a "Parent task" with priority medium
3. Create a task "Subtask" with priority "Do now"
4. Add the relation to make it a subtask
5. Check the "DO NOW!!!" filter --> No tasks will be displayed and therefore this urgent task gets overlooked

Thank you very much!

### Vikunja Version

Version: v2.3.0-94-50465818

### Browser and version

Version 146.0.7680.177 & MS Edge

Yes

### Screenshots

_No response_

v-marinkov:
resolves https://github.com/go-vikunja/vikunja/issues/2646

Fixing the following bugs for subtasks in saved filters:

1. Subtasks that matched a filter were missing if their parent task did not match the filter. https://github.com/go-vikunja/vikunja/issues/2646
2. Subtasks that did not match a filter were incorrectly shown if their parent task matched the filter. https://github.com/go-vikunja/vikunja/issues/1786#issuecomment-4287980648
3. Subtasks were visually duplicated in the list, appearing both nested under their parent and independently at the root level. https://github.com/go-vikunja/vikunja/issues/1786

The core issue stems from how the `expand=subtasks` logic interacts with filtered views. `expand=subtasks` was originally designed for unfiltered lists to isolate root tasks and recursively fetch their children. When applied to filtered views, it aggressively stripped out matching subtasks (bug #1) and indiscriminately pulled in non-matching subtasks via its recursive SQL query (bug #2).

Bug #3 was reintroduced after a regression that fixed https://github.com/go-vikunja/vikunja/issues/1786 so the PR reverts back to that fix.

I have tested the app locally and the filtering with subtasks now works perfectly in saved filters. Namely:

- subtasks show one time as standalone tasks in the filtered view if they fit the filter but their parent doesn't
- if the parent fits the filter but some subtasks don't, those subtasks are NOT shown
- if the parent and some of its subtasks match the filter, then the parent is shown and the matching subtasks are shown as subtasks, not standalone tasks. Thus, also the duplicate displaying issue is solved.


fritzbauer:
Just tested the PR preview and can confirm that this fixes https://github.com/go-vikunja/vikunja/issues/2646 as well. Thanks!

kolaente:
while your fix here works, it breaks when I want to show tasks that match a filter and their subtasks. Now if the parent matches but the subtask doesn't, the subtask is not shown even when I pass `expand=subtasks`. This breaks the contract that's expected for the parameter.

Took another stab at it that fixes this at the root: https://github.com/go-vikunja/vikunja/pull/2952, superseeding this PR.

PR #2688, opened 2026-04-24
fix(webhook): use same casing for basic auth fields

### LINKED ISSUE #2687: Webhook Basic Authentication data does not persist
### Pre-submission checklist

### Description

When setting up a webhook that uses basic authentication, the username and password are not persisted in the backend. As a result, whenever the webhook fires it always gets a 403 from the protected endpoint.

Actual steps:
1. Setup a ntfy instance with protected publishing. Allow read/write access to a user on the vikunja topic
2. Create a webhook that publishes messages on the vikunja topic. 
3. Trigger the webhook.

Expected: notification in NTFY
Actual: No notification, 403 failure logged in vikunja.

This actually happens because when creating the webhook, the two fields are set in the front-end model as lowercase, but the backend expects camel case (basicuserauth vs basicUserAuth).

### Vikunja Version

2.3.0

### Browser and version

_No response_

Yes

### Screenshots

_No response_

civascu:
Fixes #2687

Webhook basic authentication fields are not persisted because of a mismatch between frontend and backend model names.

### DIFF
--- a/frontend/src/components/misc/WebhookManager.vue
+++ b/frontend/src/components/misc/WebhookManager.vue
@@ -145,7 +145,7 @@ function doDelete() {
 					<template #default="{id}">
 						<FormInput
 							:id="id"
-							v-model="newWebhook.basicauthuser"
+							v-model="newWebhook.basicAuthUser"
 						/>
 					</template>
 				</FormField>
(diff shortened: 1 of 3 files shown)


kolaente (approved):
Good catch! Thanks

PR #2689, opened 2026-04-24
feat: Add an assigned to me filter in ShowTasks

mdrkrg:
Add a checkbox to enable filtering for tasks that are assigned to current user, displayed in "Upcoming" tasks view.


kolaente:
Sorry, but this should be a saved filter, not another checkbox.

Issue #42, opened 2022-08-12
Sync Vikunja teams to OpenID groups


AlvinVrolijk:
Add the support to sync your OpenID user groups to the (existing) Vikunja teams, maybe with the option to create automatically new teams when they're not yet in Vikunja.

kolaente:
Related to https://kolaente.dev/vikunja/api/issues/1196

PR #2700, opened 2026-04-27
feat: add configurable task identifier display format (service.taskidentifierdisplay)

### LINKED ISSUE #2698: feat: user-configurable task identifier format string to expose global task ID in the UI
## Background: task index instability

Vikunja assigns each task a per-project `index` using `MAX(current_project_index) + 1` (see `calculateNextTaskIndex` in `pkg/models/tasks.go`). This is simple, generally race-safe (the `UNIQUE(project_id, index)` constraint added in the migration shipped with PR #2596 closes the concurrent-write gap), and produces familiar `PROJ-N` identifiers in the UI. What it does not guarantee is that a given `PROJ-N` identifier refers to the same task over time.

### Scenario A - an index is permanently orphaned

1. Task A created in project TAL -> assigned index 1 (`TAL-1`)
2. Task B created in TAL -> assigned index 2 (`TAL-2`)
3. Task A moved to project ABC -> receives a new index in ABC; TAL now contains only B (max index = 2)
4. Task A moved back to TAL -> `calculateNextTaskIndex` returns 3; Task A is now `TAL-3`

`TAL-1` is a permanent gap. `GET /projects/TAL/tasks/by-index/1` returns 404. Task A is only findable via its stable global `id`.

### Scenario B - an index is silently reused

1. Tasks 1-5 created in TAL; Task E is `TAL-5` (the current maximum)
2. Task E moved to project ABC -> TAL's `MAX(index)` drops to 4
3. New Task F created in TAL -> `calculateNextTaskIndex` returns 5; Task F is now `TAL-5`

`TAL-5` now silently refers to Task F. Anyone holding a reference to the old `TAL-5` resolves the wrong task with no error, no redirect, and no warning.

### The new by-index endpoint acknowledges this

PR #2596 (merged April 2026) added `GET /projects/{project}/tasks/by-index/{index}` along with the `UNIQUE(project_id, index)` constraint. The endpoint's own Swagger annotation states: *"Note that task indexes are reassigned when a task is moved between projects, so **long-lived references should use the returned task `id` instead**."* For the orphaned index in Scenario A, the endpoint returns 404. The task is only reachable via its global `id`.

---

## We are not proposing to change the index assignment algorithm

`MAX + 1` is simple, it works for the common case, and changing it would be a significant migration with unclear benefit. The scenarios above are real but not catastrophic for typical human workflows where identifiers are short-lived references (a PR description, a standup comment).

The problem becomes acute for **API consumers, integrations, and automation** (including AI tooling and scripts):

- All Vikunja API calls that reference a specific task use the **global `id`** (the database primary key), not the project index
- The global `id` is always stable -- it never changes regardless of project moves, renames, or anything else
- But there is currently no way to see the global `id` from the UI without reading it out of the browser URL bar on the task detail page
- Meanwhile the per-project index (`TAL-5`) is prominently rendered in the task heading and in every view (Kanban card, table row, etc.)

A user who wants to reference a task in an API call or a script must navigate to the task detail page, copy the URL, and extract the numeric ID from it -- while the number prominently displayed (`TAL-5`) is the one they cannot use in the API.

---

## Proposed: a user-configurable identifier format string

Rather than a fixed enum of display modes, we propose a format string with a small, pre-defined set of named placeholders:

| Placeholder | Value |
|---|---|
| `{identifier}` | the current computed value: `TAL-5` if the project has an identifier prefix set, `#5` if not |
| `{id}` | the global database primary key -- stable, API-safe, never changes |
| `{index}` | the raw per-project sequential number |
| `{prefix}` | the project identifier prefix (e.g. `TAL`), empty string if not set |

`{identifier}` is the key placeholder. It encapsulates the existing `getTaskIdentifier()` logic -- the `identifier === '' ? #${index} : identifier` branch -- and is the correct building block for any format string that needs to express "current behaviour plus something else." Using `{prefix}-{index}` directly would render `-5` when no project prefix is configured. `{identifier}` never does.

### Example format strings

| Format string | Project with prefix | Project without prefix |
|---|---|---|
| `{identifier}` | `TAL-5` | `#5` |
| `#{id}` | `#42` | `#42` |
| `{identifier} (#{id})` | `TAL-5 (#42)` | `#5 (#42)` |
| `#{id} {identifier}` | `#42 TAL-5` | `#42 #5` |

The default would be `{identifier}`, preserving current behaviour exactly -- no change for existing users who do not touch the setting.

### Implementation sketch

The rendering logic is almost centralised already. `getTaskIdentifier()` in `frontend/src/models/task.ts` is used by `Heading.vue` (the task detail `h1`). `KanbanCard.vue` and `ProjectTable.vue` currently bypass it with inline `v-if/v-else` templates -- consolidating those to call `getTaskIdentifier()` is a small cleanup that should happen regardless.

Extending `getTaskIdentifier()` to accept a format string and perform placeholder substitution is approximately 10 lines. The user-supplied string is never executed -- only processed by `.replace()` against a fixed set of known-safe values (`task.id` and `task.index` are numbers; `task.identifier` is already rendered in the UI today). There is no parser, no expression evaluator, and no XSS surface.

User preference storage has two options:

- **Local storage only** (no backend changes): a `useLocalStorage` composable, not synced across devices. Sufficient for a first implementation.
- **User API settings** (synced across devices): requires a backend migration adding a column to the user settings table plus an API change -- roughly a half-day of additional backend work.

A companion addition to `frontend/src/i18n/lang/en.json` would be needed for the settings label and placeholder documentation.

---

*Related: #2596 (by-index endpoint + unique constraint), #2694 (by-ref endpoint proposal), #476 (original identifier design discussion)*

javabrett:
## Motivation

Closes #2698.

The task identifier shown in the UI (`TAL-5`, `#5`) is the per-project index form. This is useful for human navigation but problematic for integrations and API consumers, because the project index is not stable over time: tasks pick up a new index when moved between projects, and indexes are reassigned as tasks are deleted, so a reference like `TAL-5` can silently point to a different task than it did previously. The only stable identifier is the global database `id`, which appears in every API response and URL but is not currently shown anywhere in the UI.

This PR introduces a server-side configuration key that lets admins choose how task identifiers render throughout the frontend, including exposing the stable global `id` alongside the existing per-project display.

Screenshots (task, Table view, Kanban view) when running with `VIKUNJA_SERVICE_TASKIDENTIFIERDISPLAY={identifier} (ID #{id})` :

Screenshot 2026-04-28 at 08 44 52
Screenshot 2026-04-28 at 08 44 13
Screenshot 2026-04-28 at 08 44 32

### Backend

A new config key `service.taskidentifierdisplay` is added (env: `VIKUNJA_SERVICE_TASKIDENTIFIERDISPLAY`, default: `{identifier}`). The value is a format string exposed verbatim via `/api/v1/info` as `task_identifier_display` so the frontend can read it. No backend interpretation of the format string occurs beyond passing it through.

A startup warning (at `WARN` level) is emitted when the configured string contains `{prefix}-{index}`, which renders as `-5` for projects with no identifier prefix set (see footgun note below).

### Frontend

`getTaskIdentifier(task, format?)` in `models/task.ts` now supports format strings. When `format` is omitted the function reads from `useConfigStore().taskIdentifierDisplay`, falling back to `{identifier}`. The substitution is plain `String.prototype.replace()` against fixed placeholder patterns — no parser, no expression evaluation, no XSS surface.

Supported placeholders:

| Placeholder | Value |
|---|---|
| `{identifier}` | Per-project value: `TAL-5` if the project has an identifier prefix, otherwise `#5`. Handles the empty-prefix case automatically. **Recommended.** |
| `{id}` | Global database id (stable, never reassigned). |
| `{index}` | Per-project numeric index (same number as in `{identifier}`, without the prefix). |
| `{prefix}` | Project identifier prefix (e.g. `TAL`); empty string if not set. |

`KanbanCard.vue` and `ProjectTable.vue` are updated to call `getTaskIdentifier(task)` in place of their inline `v-if identifier === ''` templates, consolidating the display logic in one place.

`stores/config.ts` adds `taskIdentifierDisplay` to `ConfigState`, populated automatically by the existing `objectToCamelCase` step on the `/info` response. A `console.warn` is emitted on first load when the footgun pattern is detected.

### Footgun: `{prefix}-{index}`

A format string using `{prefix}-{index}` directly renders as `-5` (a leading dash) for projects that have no identifier prefix set. This is not blocked — an admin who only uses projects with prefixes will get correct output — but both the backend (startup log) and frontend (`console.warn`) warn when this pattern is present in the configured string. The recommended alternative is `{identifier}`, which encapsulates the `prefix-index`/`#index` conditional automatically.

### Default behaviour

The default value of `{identifier}` is identical to the current hardcoded behaviour. Existing deployments that do not set this key see no change.

## Examples

| Format string | Project with prefix | Project without prefix |
|---|---|---|
| `{identifier}` *(default)* | `TAL-5` | `#5` |
| `#{id}` | `#42` | `#42` |
| `{identifier} (ID #{id})` | `TAL-5 (ID #42)` | `#5 (ID #42)` |
| `{prefix}-{index}` *(footgun)* | `TAL-5` | `-5` |

## Tests

14 Vitest unit tests added in `frontend/src/models/task.test.ts` covering: null/undefined task, `{identifier}` with and without prefix, `{id}`, combined formats, the footgun case (documented `-5` output), literal strings without placeholders, unknown placeholders left intact, and repeated placeholder substitution.

Existing E2E test assertions at `tests/e2e/task/task.spec.ts` (lines 281 and 331) remain valid because the default format `{identifier}` produces identical output to the previous hardcoded logic.

### DIFF
--- a/config-raw.json
+++ b/config-raw.json
@@ -139,6 +139,11 @@
                     "comment": "Allow using a custom logo for dark mode via external URL. If not set, the regular logo will be used for both light and dark modes."
                 },
                 {
+                    "key": "taskidentifierdisplay",
+                    "default_value": "{identifier}",
+                    "comment": "Format string controlling how task identifiers render in the UI.\nSupported placeholders:\n  {identifier} - the per-project value (e.g. 'TAL-5' if the project has an identifier prefix, otherwise '#5'). This is the recommended placeholder for the empty-prefix case.\n  {id}         - the global database id (stable, never changes; used by the API).\n  {index}      - the per-project numeric index.\n  {prefix}     - the project identifier prefix (e.g. 'TAL'); empty string if not set.\nExamples:\n  '{identifier}'             - default; shows 'TAL-5' or '#5'.\n  '#{id}'                    - shows the global database id (e.g. '#42').\n  '{identifier} (ID #{id})'  - shows both; renders as 'TAL-5 (ID #42)' or '#5 (ID #42)'. The 'ID' label disambiguates the database id from the project index when no prefix is set.\nNote: '{prefix}-{index}' renders as '-5' for projects without an identifier prefix; prefer {identifier} unless you control all projects."
+                },
+                {
                     "key": "enablepublicteams",
                     "default_value": "false",
                     "comment": "Enables the public team feature. If enabled, it is possible to configure teams to be public, which makes them\ndiscoverable when sharing a project, therefore not only showing teams the user is member of."
--- a/frontend/src/components/project/views/ProjectTable.vue
+++ b/frontend/src/components/project/views/ProjectTable.vue
@@ -198,12 +198,7 @@
 								>
 									<td v-if="activeColumns.index">
 										<RouterLink :to="taskDetailRoutes[t.id]">
-											<template v-if="t.identifier === ''">
-												#{{ t.index }}
-											</template>
-											<template v-else>
-												{{ t.identifier }}
-											</template>
+											{{ getTaskIdentifier(t) }}
 										</RouterLink>
 									</td>
 									<td v-if="activeColumns.done">
@@ -321,6 +316,7 @@ import type {SortBy} from '@/composables/useTaskList'
 import {useTaskList} from '@/composables/useTaskList'
 import type {ITask} from '@/modelTypes/ITask'
 import type {IProject} from '@/modelTypes/IProject'
+import {getTaskIdentifier} from '@/models/task'
 import AssigneeList from '@/components/tasks/partials/AssigneeList.vue'
 import type {IProjectView} from '@/modelTypes/IProjectView'
 import { camelCase } from 'change-case'
--- a/frontend/src/components/tasks/partials/KanbanCard.vue
+++ b/frontend/src/components/tasks/partials/KanbanCard.vue
@@ -29,12 +29,7 @@
 						:is-done="task.done"
 						variant="small"
 					/>
-					<template v-if="task.identifier === ''">
-						#{{ task.index }}
-					</template>
-					<template v-else>
-						{{ task.identifier }}
-					</template>
+					{{ getTaskIdentifier(task) }}
 					<span
 						v-if="showTaskPosition"
 						class="tw:text-red-600 tw:ps-2"
@@ -126,7 +121,7 @@ import Labels from '@/components/tasks/partials/Labels.vue'
 import ChecklistSummary from './ChecklistSummary.vue'
 import CommentCount from './CommentCount.vue'
 
-import {getHexColor} from '@/models/task'
+import {getHexColor, getTaskIdentifier} from '@/models/task'
 import type {ITask} from '@/modelTypes/ITask'
 import type {IProject} from '@/modelTypes/IProject'
 import {SUPPORTED_IMAGE_SUFFIX} from '@/models/attachment'
--- a/frontend/src/models/task.test.ts
+++ b/frontend/src/models/task.test.ts
@@ -1,0 +1,97 @@
+import {describe, it, expect} from 'vitest'
+import {getTaskIdentifier} from './task'
+import type {ITask} from '@/modelTypes/ITask'
+
+function makeTask(overrides: Partial<ITask> = {}): ITask {
+	// Test focuses on getTaskIdentifier which only reads id, index, identifier.
+	// Cast to ITask to avoid enumerating all unrelated required fields.
+	return {
+		id: 42,
+		index: 5,
+		identifier: 'TAL-5',
+		title: 'Some task',
+		...overrides,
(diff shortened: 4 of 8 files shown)


kolaente (changes requested):
Sorry for not getting back earlier, seems like this PR was lost in the huge pile of new PRs we're getting right now.

Why does this have to be a server setting if it only affects display?

javabrett:
Sorry for the late reply - yes, more than happy to let this go now.  It was born before I moved from self-managed to Cloud and am happy with life in the Cloud.  If later picked-up and used in Cloud yes it would need to be a per-user or per-tenant option.

I've trained my agent code to not use project-prefix based slugs and always use the immutable ID.

PR #2701, opened 2026-04-27
Add a small light border around the text editor input field

melroy89:
- Add a small light border around the text editor input field, since it was hard to find where to click and add text (especially on mobile)
- Works with both light and dark themes
- Add some margin between the toolbar & text area
  - **Note:** Adding a border around the full toolbar + text area doesn't make sense, since that will break the highlighting of the text area section (which is currently already in place)

---

See screenshot, **BEFORE** (no border / no indication of the text area):

no_indicator

And **AFTER** (both dark and light):

with_border_indicator

with_border_indicator_light

Improving also accessibility.

### DIFF
--- a/frontend/src/components/input/editor/EditorToolbar.vue
+++ b/frontend/src/components/input/editor/EditorToolbar.vue
@@ -395,6 +395,7 @@ function setLink(event: MouseEvent) {
 	border-radius: $radius;
 	display: flex;
 	flex-wrap: wrap;
+	margin-block-end: 2px;
 
 	> * + * {
 		border-inline-start: 1px solid var(--grey-200);
--- a/frontend/src/components/input/editor/EditorToolbar.vue
+++ b/frontend/src/components/input/editor/EditorToolbar.vue
@@ -395,7 +395,7 @@ function setLink(event: MouseEvent) {
 	border-radius: $radius;
 	display: flex;
 	flex-wrap: wrap;
-	margin-block-end: 4px;
+	margin-block-end: 3px;
 
 	> * + * {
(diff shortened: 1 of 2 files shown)


kolaente (changes requested):
Please make it one border around the whole thing, with your changes we now have 2px border between the toolbar and the actual input.

Otherwise great!

melroy89:
> Please make it one border around the whole thing, with your changes we now have 2px border between the toolbar and the actual input.
> 
> Otherwise great!

I see, **but** uh.. The toolbar already has its own border around it. And also highlighting the text field (even without this PR) will add a blue border around the input field:

image


Meaning, this change makes the most sense if you want to be compatible with the highlighting of the input field. My alternative solution would be to add maybe a tiny bit margin?

melroy89:
So like I said before, your solution is **NOT** recommended. If you want to keep the active blue highlighting when you selected the text area field.

Hence my proposed solution was to just add a small margin between the two sections (3px seems to be the sweet-spot):

image


When field is active (this blue was already in place):

image

---

This is also very similar to like how GitHub is solving this. 

image

melroy89:
Please?

kolaente:
Sorry, but I don't like the margin. I'll check what I can do here.

kolaente:
Thought about this some more and added a version of this in 5800f323a42fae5d22d6dfbb3445b8ed764b89c0 - I've experimented with a border around what you added here before, but found it to look too much out of place within the whole UI. The change adds a border around the whole editor on focus, without any spacing shenanigans. Just adding a border around the editor that's always there needs more thought and probably adjustment in more places, which is not something I'll do right now.

PR #2704, opened 2026-04-27
feat(frontend): preload assignee list on click with current user first

Tokra110:
## Summary

- Show all available project members in the assignee dropdown when the user clicks the search input, instead of requiring them to type first
- The current user is sorted to the top of the list for quick self-assignment
- Uses the existing `showEmpty` prop on `Multiselect` so no changes to shared components are needed

## How it works

The `EditAssignees` component now:
1. Passes `:show-empty="true"` to `Multiselect` so the dropdown renders even with an empty search query
2. Preloads all project members on the first click via `@click="preloadUsers"` (not on auto-focus, to avoid a jarring dropdown appearing far from where the user clicked "assign user")
3. Sorts results so the current user always appears first

## Test plan

- [x] Open a task detail view, click "assign user", then click the search input: dropdown shows all project members with current user on top
- [x] Type a name: results filter correctly
- [x] Assign someone, click the input again: the assigned user is excluded from the list
- [x] Test on a single-user project: only the current user appears (or empty if already assigned)


frontend/src/components/tasks/partials/EditAssignees.vue
```diff
@@ -7,11 +7,13 @@
 		:placeholder="$t('task.assignee.placeholder')"
 		:multiple="true"
 		:search-results="foundUsers"
+		:show-empty="true"
 		label="name"
 		:select-placeholder="$t('task.assignee.selectPlaceholder')"
 		:autocomplete-enabled="false"
 		@search="findUser"
 		@select="addAssignee"
+		@click="preloadUsers"
```
kolaente: This should be focus, not click

frontend/src/components/tasks/partials/EditAssignees.vue
```diff
@@ -68,6 +72,14 @@ const foundUsers = ref<IUser[]>([])
 const assignees = ref<IUser[]>([])
 let isAdding = false
 
+let hasPreloaded = false
+
+function preloadUsers() {
+	if (hasPreloaded) return
+	hasPreloaded = true
+	findUser('')
```
kolaente: do you have to pass `''`?

frontend/src/components/tasks/partials/EditAssignees.vue
```diff
@@ -117,6 +131,11 @@ async function findUser(query: string) {
 			u.name = getDisplayName(u)
 			return u
 		})
+		.sort((a, b) => {
+			if (a.id === currentUserId) return -1
+			if (b.id === currentUserId) return 1
+			return 0
```
kolaente: We should sort by display name alphabetically otherwise

kolaente (changes requested):

frontend/src/components/tasks/partials/EditAssignees.vue
Tokra110: Done, changed to `@focus`.

frontend/src/components/tasks/partials/EditAssignees.vue
Tokra110: Not needed, made the parameter default to `''` so the call is just `findUser()`.

frontend/src/components/tasks/partials/EditAssignees.vue
Tokra110: Added `localeCompare` on the display name for non-current-user entries.

kolaente (approved):

PR #2717, opened 2026-05-04
feat: Add filter quick keywords panel in the task filters

mdrkrg:
The user can select the keyword in the filter popup, under the input. The keywords, operators and join operators are coming from `frontend/src/helpers/filters.ts`.

Selecting a keyword to the input will bring a `=` after it, in some cases it is not the best, but I think this is acceptable.

The ordering of the keywords are not the best, ideally the most used keywords should come before the less frequently used ones, but this is the ordering of `AVAILABLE_FILTER_FIELDS`.

### DIFF
--- a/frontend/src/components/input/filter/FilterQuickKeywords.vue
+++ b/frontend/src/components/input/filter/FilterQuickKeywords.vue
@@ -1,0 +1,153 @@
+<template>
+	<div id="filter-quick-keywords" class="keywords mbe-2">
+		<div class="keywords-section">
+			<span class="keywords-label">{{ $t('filters.keywords.fields') }}</span>
+			<div class="keywords-buttons">
+				<BaseButton
+					v-for="field in filterFields"
+					:key="field"
+					v-tooltip="{
+						content: $t(`filters.query.help.fields.${field}`),
+						container: '#filter-quick-keywords',
+					}"
+					size="small"
+					variant="ghost"
+					class="keyword-btn"
+					@click.prevent.stop="$emit('insert', field + ' = ')"
+				>
+					{{ field }}
+				</BaseButton>
+			</div>
+		</div>
+		<div class="keywords-section">
+			<span class="keywords-label">{{ $t('filters.keywords.operators') }}</span>
+			<div class="keywords-buttons">
+				<BaseButton
+					v-for="op in filterOperators"
+					:key="op"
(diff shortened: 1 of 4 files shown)


frontend/src/components/input/filter/FilterQuickKeywords.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+	<div id="filter-quick-keywords" class="keywords mbe-2">
```
kolaente, line 2: Can you hide this block behind an `Expandable` so that it isn't as direct for the user?

frontend/src/components/input/filter/FilterQuickKeywords.vue
```diff
@@ -0,0 +1,153 @@
...
+		<div class="keywords-section">
+			<span class="keywords-label">{{ $t('filters.keywords.fields') }}</span>
+			<div class="keywords-buttons">
+				<BaseButton
```
kolaente, line 6: please also make these buttons slightly rounded (there should be a utility class that you can use)

frontend/src/components/input/filter/FilterQuickKeywords.vue
```diff
@@ -0,0 +1,153 @@
...
+					v-for="field in filterFields"
+					:key="field"
+					v-tooltip="{
+						content: $t(`filters.query.help.fields.${field}`),
+						container: '#filter-quick-keywords',
+					}"
+					size="small"
+					variant="ghost"
+					class="keyword-btn"
+					@click.prevent.stop="$emit('insert', field + ' = ')"
```
kolaente, line 16: I found it a little confusing to see a string like `dueDate = =`after pressing the button and then I directly pressed the `=` button without paying attention. Please don't add the `=` by default, I imagine it can trip up other people too. And the UX breaks a little the moment you want to use another operator anyways.

kolaente (changes requested):
Hey this is great! Thanks. Please check out the suggestions.

Sorry for the long silence, there are too many PRs right now and getting back to all of them takes a while.

Another thing we might need to have an answer for: this solves the "what fields to use" problem, but it does not help the user to add values to filter for. Not sure what to do there.

And please add some tests.

mdrkrg:
Sorry. Have been busy lately, will update later

PR #2721, opened 2026-05-04
fix(caldav): implement RFC 6578 sync-collection REPORT to fix iOS Reminders deletion sync

### LINKED ISSUE #2658: iOS Reminders CalDAV doesn't fetch new tasks
### Pre-submission checklist

### Description

iOS Reminders correctly pushes changes to Vikunja over CalDAV, but doesn't fetch changes made on the Vikunja side at all. 

iOS 26.4.1

### Vikunja Version

2.3.0

### Browser and version

_No response_

Yes

### Screenshots

_No response_

kallegrens:
# Fix CalDAV deletion sync for iOS Reminders (and other RFC 6578 clients)

## Fixes

Fixes https://github.com/go-vikunja/vikunja/issues/2658

## The bug

Apple Reminders on iOS — and any other RFC 6578-conformant CalDAV client — never learns about tasks that were deleted on the Vikunja side. The local Reminders database keeps stale entries forever, even though the same tasks have been deleted on the server.

The root cause is two-fold:

1. The CalDAV collection's ctag/sync-token did not reliably advance when tasks changed, so iOS never refetched the collection.
2. Even when iOS does refetch, there was no `sync-collection` REPORT handler that returns the canonical `404` per deleted href, so there was no protocol-level way for the client to learn that a task is gone.

## The fix

- **`STATUS:NEEDS-ACTION`** is always emitted for incomplete VTODOs (alongside the existing `STATUS:COMPLETED`) so iOS classifies them consistently.
- **RFC 6578 `sync-collection` REPORT handler**: returns changed tasks plus one `404 Not Found` `<response>` per deleted href together with a fresh sync-token. REPORT requests carrying a `sync-collection` element are intercepted before the generic CalDAV handler (which would answer 412 and make iOS silently stop syncing). Deletion records come from task soft-deletes (#3119) — soft-deleted rows are kept for 30 days, so no extra tombstone table, migration, or cron is needed (`models.GetDeletedTasksSince`).
- **ctag correctness**: every task change now bumps the project's `updated` timestamp — task create/update/delete and label add already did; relations, comments, attachments, and assignees (via the shared `HandleTaskUpdateLastUpdated` listener) plus `LabelTask.Delete` were the gaps. The ctag and sync token derive from `project.Updated` directly, with no extra per-project queries.
- **Token edge cases**: delta comparisons are inclusive (tokens have second granularity; re-reported items keep their etag so clients skip the download), and tokens older than the 30-day soft-delete retention get `403` + `valid-sync-token`, forcing a clean full resync instead of a delta that would silently miss purged deletions.

## Verification

- `go test ./pkg/models/ ./pkg/webtests/ ./pkg/caldav/` — all green, including `TestCaldavSync*` cases covering: empty token (full sync), recent token (delta), future token (empty delta), malformed/foreign-project/stale tokens (403), calendar-data rendering, and a real CalDAV DELETE followed by a delta sync reporting the 404.
- `golangci-lint` clean.
- Manually verified end-to-end against Apple Reminders on iOS: tasks deleted in the Vikunja UI now disappear from Reminders on the next refresh (verified on the earlier revision; the protocol surface is unchanged).

## Out of scope

- PostgreSQL/MySQL CI matrix for CalDAV (separate infra work).
- Refactoring the `caldav-go` fork.
- Replacing the strings-based REPORT-body inspection with a proper XML parse.


kallegrens:
Hello! Have you taken a look at this? Any feedback? I've been running this in my own fork for three weeks now without issues.

kallegrens:
If this fix is not wanted, please just close the PR. However, I think interoperability with iOS Reminders is highly wanted by many.

Promithius-DR:
I rebased this PR onto current `main` (it had drifted to a conflicting state) and independently verified it end-to-end against a **real iOS 27 device** — Apple Reminders ↔ Vikunja, through a TLS reverse proxy. Sharing results in case it helps move this along, since iOS Reminders interop is clearly wanted.

**The rebase** was clean — only three trivial list-neighbor conflicts (`models.go` `GetTables()`, `setup_tests.go` fixtures, and `tasks.go` `Task.Delete()`, where `main` now uses `doerFromAuth(s, a)`). It builds clean against `main`, `go vet` is clean, and `go test ./pkg/caldav/` passes.

**Verified behavior** (every cell tested on the device, not just locally):

| Direction | Create | Edit / Complete | Delete |
|---|---|---|---|
| Server → phone | ✅ | ✅ | ✅ — RFC 6578 `404` via `sync-collection` |
| Phone → server | ✅ (UID assigned) | ✅ | ✅ (removed + tombstone recorded) |

kallegrens:
Thank you @Promithius-DR ! Highly appreciated. Thanks for rebasing it! I’d like to keep the existing PR. Could you point me to the rebased branch in your fork? Couldn't find it. I’ll update the existing PR from that.

Promithius-DR:
Here's the pointer to the rebase: https://github.com/Promithius-DR/vikunja/tree/pr-2721-rebase

kolaente:
> HTML task descriptions being sent as raw markup in DESCRIPTION

This has been implemented in https://github.com/go-vikunja/vikunja/pull/2986

kallegrens:
Updated the branch with @Promithius-DR's rebase. Also, I resolved the merge conflict arising from #2986. Any feedback before this can be merged?

kolaente:
This is a bigger change so I'll need to find the time to review it properly. I hope to get to it next week but can't promise anything. Vikunja is getting a lot of PRs currently and I'm pretty busy with other stuff at the same time.

kolaente:
We now have soft deletes for tasks https://github.com/go-vikunja/vikunja/pull/3119 so I'll rework this to use that instead of the task tombstones.

pkg/caldav/caldav.go
```diff
@@ -325,6 +328,17 @@ func makeCalDavTimeFromTimeStamp(ts time.Time) (caldavtime string) {
...
 }
 
+// FallbackUID computes the CalDAV UID for a task that has no UID stored in the
+// database. It uses the same formula as ParseTodos so that CalDAV clients that
+// have already downloaded the task continue to see the same UID. The updated
+// time and title must be those that were in effect when the task was last served
+// to the client; if they have changed since then, the returned UID may not match
+// what the client has stored. Prefer saving this value back to the database on
+// first sight so that it stays stable across subsequent modifications.
+func FallbackUID(updated time.Time, title string) string {
```
kolaente: When is this actually required? All tasks should have a uuid in the database.

pkg/routes/caldav/listStorageProvider.go
```diff
@@ -144,13 +144,88 @@ func (vcls *VikunjaCaldavProjectStorage) GetResources(rpath string, withChildren
...
+			Find(&deletionTimes); err != nil {
+			_ = s3.Rollback()
+			return nil, err
+		}
+		if err := s3.Commit(); err != nil {
+			return nil, err
+		}
+		for _, dt := range deletionTimes {
+			if dt.DeletedAt.After(latestTaskTimes[dt.ProjectID]) {
+				latestTaskTimes[dt.ProjectID] = dt.DeletedAt
```
kolaente: Can't we rework this so that the project's updated time is always set correctly when a task is updated? That would save us the whole dance here.

kolaente:
auggie review

kallegrens:
Thanks @kolaente for picking this up and making the last portion of integrations. Will we be seeing 2.4.0 release soon?

kolaente:
Soon-ish, yes. There are still a few things I want to include, and doing them properly takes time.

PR #2741, opened 2026-05-09
chore: add mise.toml and update Renovate grouping to pin tool versions

javabrett:
## Summary

This PR adds a `mise.toml` at the repo root so that developers using [mise](https://mise.jdx.dev) can run `mise install` and immediately get the correct Node, pnpm, and Go versions -- without needing to discover and cross-reference the three separate files that currently carry these pins (`.nvmrc`, `package.json#packageManager`, `go.mod`).

It also tightens the Renovate config so that version-bump PRs update all files atomically.

## Problem it solves

mise is widely used in the Go and JS ecosystems as a unified tool version manager. Without a project-level `mise.toml`, mise falls back to the developer's global configuration -- meaning a developer with a different Node version set globally will silently build with the wrong version even though `.nvmrc` is present. Legacy version files (`.nvmrc` etc.) rank below all mise config, including the global one, so the project's pin is simply ignored.

With this change, `mise install` from the repo root activates the right versions of all three tools in one step, regardless of what a developer has configured globally.

## Changes

- `mise.toml` -- pins `node 24.13.0`, `pnpm 10.28.1`, `go 1.25.7`, each with a comment pointing to the canonical source file it mirrors
- `renovate.json` -- adds `packageRules` grouping node updates (`mise` + `nvm` managers) and pnpm updates (`mise` + `npm` managers) so Renovate raises a single PR per tool that updates all relevant files together rather than creating drift between them

**No behavioural changes** -- existing workflows (`pnpm install`, `go build`, CI) are unaffected. This is purely additive for developers who use mise.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,4 @@
+[tools]
+node = "24.13.0"  # keep in sync with frontend/.nvmrc
+pnpm = "10.28.1"  # keep in sync with frontend/package.json#packageManager
+go = "1.25.7"     # keep in sync with go.mod
--- a/renovate.json
+++ b/renovate.json
@@ -75,6 +75,26 @@
       "extends": [
         "schedule:weekly"
       ]
+    },
+    {
+      "groupName": "node",
+      "matchPackageNames": [
+        "node"
+      ],
+      "matchManagers": [
+        "mise",
+        "nvm"
+      ]
+    },
+    {
+      "groupName": "pnpm",
+      "matchPackageNames": [
+        "pnpm"
+      ],
+      "matchManagers": [
+        "mise",
+        "npm"
+      ]
     }
   ]
 }


kolaente:
We already have devenv - what makes mise better?

javabrett:
> We already have devenv - what makes mise better?

For me it's not a devenv-vs-mise, which is better and how, it's a question of: for potential new contributors to the Vikunja project, what would be the rough breakdown of:

- Already has mise installed, quite familiar with it, polyglot developer, uses it on a number of projects already, avoids Homebrew et. al. for dev-tooling
- Already has Nix+devenv installed etc., same for for Nix+devenv
- Has neither

The other axes are:

- how essential (or not) is devenv/Nix to making a successful contribution?
- how much does adding a mise config reduce build-environment set-up friction and first-build-ever error rates for new contributors?
- how much maintenance does adding mise add given Renovate config?

My suspicion is that Go developers will be quite heavy on the mise side of things.  Small amount of handwaving and not conclusive research:

- Stargazing: mise 28k, devenv 6.8k
- Homebrew: mise is in top-10 most downloaded formula consistently
- Thoughtworks tech radar: mise is at highest tier "Adopt"

kolaente:
Well I don't see myself using this, but if we can automate the updates and maintenance it should be fine. I won't make any promises to keep it around forever though

kolaente (approved):

PR #2743, opened 2026-05-09
fix(frontend): focus search input when opening quick actions modal

javabrett:
## Summary

The quick-actions search input was not being focused when the modal opened via button click or keyboard shortcut (`Cmd/Ctrl+K`). This was a silent regression introduced in cef03cb2a when the modal was refactored to use the native `<dialog>` element with `showModal()`.

**Root cause:** The `v-focus` directive fired on component mount, before `showModal()` had opened the dialog. Because browsers ignore `focus()` calls on elements inside a closed `<dialog>`, the directive silently no-oped every time the modal opened.

**Fix:**
- Add `autofocus` attribute to the search input. When `showModal()` is called, the browser processes `autofocus` as part of opening the dialog — this is spec-compliant and reliable.
- Remove `v-focus` from the input (dead code since the native dialog refactor).
- Scope the existing rAF retry loop back to `isQuickAddMode` only, restoring the original intent from before 01fff665c over-broadened it. The loop exists solely for the Electron quick-add overlay where the window may not yet be visible at `did-finish-load`; in a regular browser `autofocus` is sufficient.
- Bump hint text color from `--grey-light` (near-invisible in both light and dark modes) to `--grey-500`, and increase font size from `0.75rem` to `0.875rem` for legibility.

**Tests:** Three new Playwright E2E tests in `tests/e2e/misc/quick-actions.spec.ts` covering:
1. Focus via button click
2. Focus via `Cmd/Ctrl+K` keyboard shortcut
3. Focus after close and reopen

## Test plan

- [x] E2E tests pass locally: `mage test:e2e "tests/e2e/misc/quick-actions.spec.ts"` — 3 passed
- [x] Open quick-actions via the toolbar button — input is focused immediately
- [x] Open quick-actions via `Cmd+K` / `Ctrl+K` — input is focused immediately
- [x] Close and reopen — input is focused on the second open
- [x] Electron quick-add mode unaffected (rAF retry loop preserved for that path)

### DIFF
--- a/Dockerfile.frontend-test
+++ b/Dockerfile.frontend-test
@@ -1,0 +1,3 @@
+FROM nginx:1.27-alpine
+COPY frontend/dist /usr/share/nginx/html
+COPY Dockerfile.frontend-test.nginx.conf /etc/nginx/conf.d/default.conf
--- a/Dockerfile.frontend-test.nginx.conf
+++ b/Dockerfile.frontend-test.nginx.conf
@@ -1,0 +1,19 @@
+server {
+    listen 3457;
+
+    root /usr/share/nginx/html;
+    index index.html;
+
+    # Proxy API requests to the existing Vikunja backend.
+    # host.docker.internal resolves to the host machine on Docker Desktop.
+    location /api/ {
+        proxy_pass http://host.docker.internal:3456/api/;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+    }
+
+    # SPA catch-all: serve index.html for any path not matched above.
+    location / {
+        try_files $uri $uri/ /index.html;
+    }
+}
--- a/contrib/build-frontend-test-image.sh
+++ b/contrib/build-frontend-test-image.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# Build a lightweight nginx Docker image for testing frontend changes locally.
+# Proxies /api/ to the existing Vikunja backend at localhost:3456.
+#
+# Usage:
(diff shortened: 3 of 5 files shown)


kolaente:
What's this fixing that's not fixed by https://github.com/go-vikunja/vikunja/commit/01fff665c60e2b25e65205f706845517881db149?

javabrett:
> What's this fixing that's not fixed by https://github.com/go-vikunja/vikunja/commit/01fff665c60e2b25e65205f706845517881db149?

I'll take a look, but possibly nothing - I missed this change and it's not yet released.

javabrett:
Good catch, thank you - I had missed 01fff66 when I put this together.

**Compared to `01fff66`, the net diff in `QuickActions.vue` is:**

1. `v-focus` → `autofocus` — same effect, but browser-native via `showModal()` rather than the rAF retry loop
2. The rAF retry is scoped back to `isQuickAddMode` only — `01fff66` runs it for every open (including regular web), which is no longer needed once `autofocus` handles that case
3. CSS-only: hint text colour + size bump (unrelated to focus)

The E2E tests are new.

**Conclusion:** The user-visible focus regression is already fixed by `01fff66`. This PR's remaining value is: cleaner mechanism (`autofocus` instead of JS polling for web), restoring the rAF to its Electron-only intent, and the tests. Happy to close it if the `01fff66` approach is preferred - or I can rebase and drop the fix commit, leaving just the refactor, CSS, and tests as a follow-on cleanup.

kolaente:
Closing then since this is already fixed.

PR #2751, opened 2026-05-12
feat(hierarchical): Add option to view hierarchical presentation

simson:
Task and subbstask could be showed in hierarchical manner. This modification allows to drag and drop tasks and subtask to add easily tasks relation.
It could be enable through a tickbox.


kolaente:
What's the reason for this change? Why gate it behind a setting?

simson:
the point of the setting is to not activate it by default and to prevent UI changes for not interrested user.

i'd love to see subtasks under tasks.
moreover tasks relation implies a lot of click, so moving a tasks under another ease the creation of the subtasks relation.

kolaente:
Each additional setting is one more thing to maintain and one more thing to keep track of for users.

I feel like this is pretty complicated for such a simple feature. Can you outline why you built it that way?

PR #2761, opened 2026-05-16
fix(frontend): ensure text color inherits in filter autocomplete

bradmartin333:
Screenshot 2026-05-15 173804

Screenshot 2026-05-15 180222

use dark mode contrasting color for filter autocomplete by inhering color in scss


kolaente (approved):
Good catch!

PR #2767, opened 2026-05-18
feat(auth): sync teams from oidc groups claim

### LINKED ISSUE #2221: Implement OIDC groups claim
### Pre-submission checklist

### Description

I would be nice to see an implementation of oidc `groups` claim and scope.

This would enable further implementation of features like automatic group membership and global "whitelist" for specific groups.

Vikunja uses a custom claim `vikunja_groups` right now which requires customization of the provider which is not always possible. 

Using the de-facto standard `groups` for... well... _groups_ (almost all implementations i encountered use this for group mapping) would enable compatibility with a lot of providers out of the box, for example Google (via dex, this is my usecase).
https://dexidp.io/docs/configuration/custom-scopes-claims-clients/

_No response_

s-seidel:
This PR makes it possible to use Vikunja in environments where for example Keycloak is used as the only method for authentication and groups should be synced from Keycloak only. Closes #2221


config-raw.json
```diff
@@ -783,6 +783,11 @@
                                             "key": "requireavailability",
                                             "default_value": "false",
                                             "comment": "This option requires the OpenID Connect provider to be available during Vikunja startup. When set to `true`, Vikunja will crash if it cannot connect to the provider during initialization, allowing container orchestrators like Kubernetes to handle the failure by restarting the application. This is useful in environments where you want to ensure all authentication providers are available before the application starts serving requests. Allowed value is either `true` or `false`."
+                                        },
+                                        {
+                                            "key": "usegroupsclaim",
```
kolaente: why make this a new setting? why not use the groups claim if it is set in the `scope` config?

pkg/modules/auth/openid/openid.go
```diff
@@ -461,6 +488,10 @@ func mergeClaims(cl *claims, cl2 *claims, forceUserInfo bool) error {
 		cl.VikunjaGroups = cl2.VikunjaGroups
 	}
 
+	if (forceUserInfo && len(cl2.Groups) > 0) || len(cl.Groups) == 0 {
```
kolaente, line 557: Can we adjust this so that it either uses `vikunja_groups` or the `groups` claim and not both? No one will have configured both and this clears the way to deprecate the `vikunja_groups` claim at some point.

kolaente (changes requested):
Sorry for taking so long to get back here, there are way too many open PRs right now.

Please resolve the conflict

s-seidel:
@kolaente thanks for the review. I rebased from main and changed it so that if the `groups` claim is present, it will be used without any further configuration option needed.

pkg/modules/auth/openid/openid.go
```diff
@@ -260,9 +262,16 @@ func AuthenticateCallback(ctx context.Context, cb *Callback, providerKey string)
 		return nil, nil, err
 	}
 
-	teamData := getTeamDataFromToken(cl.VikunjaGroups, provider)
+	var teamData []*models.Team
+	teamNameSuffix := provider.Name
+	if slices.Contains(strings.Fields(provider.Scope), "groups") {
+		teamData = getTeamDataFromGroupsClaim(cl.Groups)
+		teamNameSuffix = ""
```
kolaente, line 269: Why reset the team name in this case?

pkg/modules/auth/openid/providers_test.go
```diff
@@ -279,6 +279,7 @@ func TestGetProviderFromMapStringBooleans(t *testing.T) {
 	}
 }
 
+
```
kolaente: nit: please don't introduce new empty lines

config-raw.json
```diff
@@ -762,7 +762,7 @@
                                         {
                                             "key": "scope",
                                             "default_value": "openid email profile",
-                                            "comment": "The scope necessary to use oidc.\nIf you want to use the Feature to create and assign to Vikunja teams via oidc, you have to add the custom \"vikunja_scope\" and check [openid.md](https://vikunja.io/docs/openid/).\ne.g. scope: openid email profile vikunja_scope"
+                                            "comment": "The scope necessary to use oidc.\nIf you want to use the feature to create and assign to Vikunja teams via oidc using the custom `vikunja_groups` claim, you have to add the custom \"vikunja_scope\" and check [openid.md](https://vikunja.io/docs/openid/).\ne.g. scope: openid email profile vikunja_scope\nAlternatively, you can also just add the standard OIDC `groups` claim: `scope: openid email profile groups`"
```
kolaente:
```suggestion
                                            "comment": "The scope necessary to use oidc.\nIf you want to automatically create and assign users to Vikunja teams via oidc using the custom `vikunja_groups` claim, you have to add the custom \"vikunja_scope\". For example: `scope: openid email profile vikunja_scope`. Check [the docs on this](https://vikunja.io/docs/openid/) for more details on the configuration.\nAlternatively, you can also just add the standard OIDC `groups` claim: `scope: openid email profile groups`"
```

kolaente (changes requested):

pkg/modules/auth/openid/openid.go
s-seidel: The assumption is that when groups are fetched using OIDC that it is authoritative in this case and it wouldn't make sense to get groups from OIDC and manage them locally or using another provider. Hence the added provider name is unnecessary and can be left out.

kolaente: Managing locally isn't possible due to permissions that we're enforcing elsewhere. Removing the name suffix and then not preventing manual team creation just makes this more confusing.

Please don't remove the suffix, let's handle the underlying problem in a follow up

PR #2788, opened 2026-05-21
fix(frontend): wrap notifications in Teleport to appear above modals

### LINKED ISSUE #2744: Dark theme: error notifications have near-zero contrast (barely visible)
## Summary

In dark theme, error notification toasts at the bottom of the screen are nearly invisible. The error text and border render in a dark red/maroon colour against the dark page background, with insufficient contrast to read comfortably.

Image

## Steps to reproduce

1. Enable dark theme.
2. Open **Edit This Project** for any project.
3. Enter a Project Identifier that is already in use by another project.
4. Click **Save**.

## Expected behaviour

The error notification ("A project with this identifier already exists.") is clearly readable, with sufficient foreground/background contrast.

## Actual behaviour

The notification appears at the bottom-left of the screen with very low contrast: dark red text inside a dark maroon/red border against the near-black page background. The message is difficult to read without knowing where to look.

## Environment

- Vikunja version: self-hosted (latest)
- Browser: Chrome
- Theme: Dark

## Notes

The same notification is presumably legible in light theme. The contrast issue is specific to the dark theme colour palette applied to error toast styling. A quick fix would be to use a brighter red (or white) for the text and/or a lighter background fill in the error toast when dark theme is active.

bradmartin333:
Fix #2744: keep notifications above modals

Detects the active open modal dialog and teleports notifications into it. When no modal is open, notifications continue rendering in body.

### DIFF
--- a/frontend/src/components/misc/Notification.vue
+++ b/frontend/src/components/misc/Notification.vue
@@ -1,65 +1,95 @@
 <template>
-	<Notifications
-		position="bottom left"
-		:max="2"
-		:ignore-duplicates="true"
-		class="global-notification"
-		role="status"
-		aria-live="polite"
-	>
-		<template #body="{ item, close }">
-			<!-- FIXME: overlay whole notification with button and add event listener on that button instead -->
-			<div
-				class="vue-notification-template vue-notification"
-				:class="[
-					item.type,
-				]"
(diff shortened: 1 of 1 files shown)


kolaente (approved):
Good catch, thanks!

PR #2791, opened 2026-05-24
fix(assignees): resolve 500 error when reading task assignees

### LINKED ISSUE #2740: GET /api/v1/tasks/{taskID}/assignees returns 500 with PostgreSQL
## Summary

`GET /api/v1/tasks/{taskID}/assignees` returns `500` with PostgreSQL.

It fails both when the task has no assignees and after assigning a user. The write endpoint works.

## Environment

- Vikunja: `v2.3.0`
- Database: PostgreSQL
- Deployment: Docker

## Reproduction

1. Create a task.
2. Call:

```bash
curl -i "$VIKUNJA_URL/api/v1/tasks/$TASK_ID/assignees" \
  -H "Authorization: Bearer $TOKEN"
```

Actual result:

```text
HTTP/1.1 500 Internal Server Error
```

Server log:

```text
method=GET uri=/api/v1/tasks/1618/assignees status=500 err="sql: no rows in result set"
```

3. Assign a user to the task:

```bash
curl -i -X PUT "$VIKUNJA_URL/api/v1/tasks/$TASK_ID/assignees" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"user_id":3}'
```

Actual result:

```text
HTTP/1.1 201 Created
```

4. Call the read endpoint again:

```bash
curl -i "$VIKUNJA_URL/api/v1/tasks/$TASK_ID/assignees" \
  -H "Authorization: Bearer $TOKEN"
```

Actual result:

```text
HTTP/1.1 500 Internal Server Error
```

Server log:

```text
method=GET uri=/api/v1/tasks/1618/assignees status=500 err="sql: expected 26 destination arguments in Scan, not 1"
```

## Expected Behavior

- No assignees: return `200 OK` with `[]`
- With assignees: return `200 OK` with the assignee list

## Suspected Cause

The count query in `TaskAssginee.ReadAll` seems to select `users.*` while calling `Count(&user.User{})`:

```go
numberOfTotalItems, err = s.Table("task_assignees").
    Select("users.*").
    Join("INNER", "users", "task_assignees.user_id = users.id").
    Where("task_id = ? AND users.username LIKE ?", la.TaskID, "%"+search+"%").
    Count(&user.User{})
```

This appears to produce a non-scalar select during count, which fails with PostgreSQL.

The count query probably should not use `Select("users.*")`.

nithinvarma411:
## Summary
Fixes the GET /api/v1/tasks/{taskID}/assignees endpoint returning 500 errors on both SQLite and PostgreSQL.

## How it's fixed
The count query in `TaskAssginee.ReadAll()` was using `Select("users.*")` which produced non-scalar SQL. This causes:
- PostgreSQL: "expected 26 destination arguments in Scan, not 1" (with assignees) or "no rows in result set" (without)
- SQLite: similar count mismatch error

**Fix:** Remove the `Select()` from the count query and count on the driving table (TaskAssginee) instead. The INNER JOIN on users is kept because the WHERE clause filters by username.

## Test coverage
Added `TestTaskAssignees` with tests for:
- Reading assignees when task has assignees (returns 200 with user list)
- Reading assignees when task has no assignees (returns 200 with empty array)
- Permission checks

Tests: FAIL without fix → PASS with fix

Closes/Resolves #2740

### DIFF
--- a/pkg/models/task_assignees.go
+++ b/pkg/models/task_assignees.go
@@ -334,10 +334,9 @@ func (la *TaskAssginee) ReadAll(s *xorm.Session, a web.Auth, search string, page
 	}
 
 	numberOfTotalItems, err = s.Table("task_assignees").
-		Select("users.*").
 		Join("INNER", "users", "task_assignees.user_id = users.id").
 		Where("task_id = ? AND users.username LIKE ?", la.TaskID, "%"+search+"%").
-		Count(&user.User{})
+		Count(&TaskAssginee{})
 	return taskAssignees, len(taskAssignees), numberOfTotalItems, err
 }
 
--- a/pkg/webtests/task_test.go
+++ b/pkg/webtests/task_test.go
@@ -563,3 +563,36 @@ func TestTaskDuplicate(t *testing.T) {
 		assert.Contains(t, getHTTPErrorMessage(err), "Forbidden")
 	})
 }
+
+func TestTaskAssignees(t *testing.T) {
+	testHandler := webHandlerTest{
+		user: &testuser1,
+		strFunc: func() handler.CObject {
+			return &models.TaskAssginee{}
+		},
+		t: t,
+	}
+
+	t.Run("ReadAll", func(t *testing.T) {
+		t.Run("With assignees", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


kolaente (approved):
Good catch, thanks!

Issue #2754, opened 2026-05-14
Impossible to import calendar on MacOS


remilapeyre:
### Pre-submission checklist

- [x] I have searched for existing open or closed issue reports with the same problem.

### Description

I am trying to import the tasks in my calendar using the CalDav integration but it fails with the Calendar app on MacOS.

Looking at the logs there is this error:

```logs
{"time":"2026-05-14T20:04:22.094883694Z","level":"INFO","component":"http","remote_ip":"127.0.0.1","method":"PROPFIND","uri":"/dav/principals/remi/","status":207,"latency":143632060}
{"time":"2026-05-14T20:04:22.288514033Z","level":"INFO","component":"http","remote_ip":"127.0.0.1","method":"OPTIONS","uri":"/dav/principals/remi/","status":200,"latency":131502222}
{"time":"2026-05-14T20:04:22.485488555Z","level":"INFO","component":"http","remote_ip":"127.0.0.1","method":"PROPFIND","uri":"/dav/projects/","status":207,"latency":138770157}
{"time":"2026-05-14T20:04:22.680723917Z","level":"ERROR","component":"http","remote_ip":"127.0.0.1","method":"MKCALENDAR","uri":"/dav/projects/E6948AA3-86CC-40A1-874D-3B0A02FAA781/","status":500,"latency":130530526,"err":"strconv.ParseInt: parsing \"E6948AA3-86CC-40A1-874D-3B0A02FAA781\": invalid syntax"}
```

which seems to be coming from https://github.com/go-vikunja/vikunja/blob/main/pkg/routes/caldav/handler.go#L186 but I don't know the CalDav protocol and I'm not sure why Calendar is making this request.

### Vikunja Version

2.3.0

### Browser and version

MacOS Calendar application (latest version)

### Can you reproduce the bug on the Vikunja demo site?

Yes

remilapeyre:
Ok, I understand now that I needed to add the account in System Settings and not from the calendar app and it seems like it's only working with the Reminder app, not Calendar.

whiskey:
Hit the same on Vikunja 2.3.0. Adding some signal in case it helps triage:

[...]

PR #2796, opened 2026-05-26
fix(caldav): return 404 when trying to access a project that cannot exist with CalDAV

remilapeyre:
Related to https://github.com/go-vikunja/vikunja/issues/2754


kolaente (approved):
Good catch, thanks!

PR #2804, opened 2026-05-27
fix: preserve login username on tab unfocus

bdamokos:
## Summary

Fixes the login form losing an already-entered username when the login view remounts after the browser tab loses and regains focus.

The username field intentionally remains DOM-driven so the existing browser autofill workaround is preserved. The patch adds a small session-scoped mirror of the username, restores it when the login form mounts again, and clears it after a successful login.

## Root Cause

The username field was intentionally not bound with `v-model` because some browsers do not reliably propagate autofilled values through Vue bindings. That meant the value only lived in the DOM. If the login form remounted while still on the login page, the DOM node was recreated and the typed username was lost.

## Screenshot

The UI is visually unchanged. 

Screenshot 2026-05-27 at 11 45 58

## Tests

- `./node_modules/.bin/vitest --run src/views/user/Login.test.ts`
- `./node_modules/.bin/eslint src/views/user/Login.vue`
- `./node_modules/.bin/eslint --no-ignore src/views/user/Login.test.ts`

## Manual Verification

- Started the frontend dev server with `DEV_PROXY=https://try.vikunja.io`.
- Filled the username field on `/login`.
- Switched focus to another browser tab and back.
- Confirmed the username remained populated.

### DIFF
--- a/frontend/src/views/user/Login.test.ts
+++ b/frontend/src/views/user/Login.test.ts
@@ -1,0 +1,130 @@
+import {beforeEach, describe, expect, it, vi} from 'vitest'
+import {mount} from '@vue/test-utils'
+
+import Login from './Login.vue'
+
+const login = vi.fn()
+
+vi.mock('vue-i18n', () => ({
+	useI18n: () => ({
+		t: (key: string) => key,
+	}),
+}))
+
+vi.mock('vue-router', () => ({
+	useRouter: () => ({
+		push: vi.fn(),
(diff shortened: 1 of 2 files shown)


kolaente:
I'm unable to reproduce the original issue. Please provide exact steps to reproduce.

bdamokos:
Hm, strange - this was happening to me repeatedly before, but now does not anymore. Can close for now

Issue #2594, opened 2026-04-10
Allow toggle for recurring tasks to stay in their respective kanban bucket following pre-2.3.0 behaviour


acenomad:
### Pre-submission checklist

- [x] I have searched for existing open or closed issue reports with the same problem.

### Description

The latest version introduces a change that automatically [moves recurring tasks to the default bucket when completed](https://github.com/go-vikunja/vikunja/pull/2574).

This breaks workflows in cases where, for example, an entire board is made up of recurring tasks (eg. for tracking habits or regular maintenance-type work) and ultimately has no default or done buckets, or cases where a specific bucket in a given board has been designated exclusively for recurring tasks (eg. for visibility or clarity purposes).

I understand that this was a result of a [feature request](https://github.com/go-vikunja/vikunja/issues/2573), however it's quite an annoying regression in my case that means each time a recurring task is completed I must move it back into its intended bucket, else it gets lost in a long list of ingress tickets or misplaced in an incorrect bucket altogether. It would be nice to be able to toggle this back to its pre-2.3.0 behaviour to have the recurring tasks remain in their bucket at the time of completion.

Otherwise, I've been a long-time user of Vikunja and have adored and relied upon it for a very long time to organise my work, life, and projects. Thanks for the wonderful work :)

### Vikunja Version

2.3.0

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
[Relevant XKCD](https://xkcd.com/1172/)

I understand your concern, but I don't think we can add a toggle here. Every toggle adds optionality that makes Vikunja harder to maintain and to use, that's not the way I want it to go.

A workaround I can offer for your specific use-case: Create a kanban view where each bucket is based on a filter, add a label to each kind of task, then set the filters on the buckets such that they show up in the way you need them.

acenomad:
Haha interesting perspective with the comic. I gather that means your opinion is that the old behaviour was not intentional and was incorrect the first place. That's a shame because the new behaviour is not intuitive for me at all, nor has it been my default experience with other management platforms in the past either.

Unfortunately this is a pretty large regression for a substantial number of my boards and I will now need to evaluate whether or not Vikunja is the appropriate tool for me moving forward.

I understand all decisions come with trade-offs, but I had hoped that a toggle would strike a decent middle ground between the two without much complexity. Either way, thanks for taking the time to respond.

ringmountain:
[...]

PR #2806, opened 2026-05-28
fix(task): recurring tasks return to original bucket when no default is set

### LINKED ISSUE #2786: Bug: Recurring tasks are recreated in first bucket when no default bucket is set
### Pre-submission checklist

### Description

**Description**
When using recurring tasks in a project with multiple buckets, completed recurring tasks are recreated in the project's first bucket rather than in the bucket they originally belonged to.

This only happens when no bucket is configured as the project’s default bucket.

**Expected behaviour**
A recurring task should be recreated in the same bucket it was originally created in.

**Actual behaviour**
After completing a recurring task:
* the new recurring instance is created
* but it is placed into the first bucket in the project
* instead of remaining in its original bucket

**Steps to reproduce**
1. Create a project with multiple buckets
2. Do not set any bucket as the default bucket
3. Create recurring tasks in different buckets
4. Complete one of the recurring tasks
5. Observe that the recreated task appears in the first bucket instead of the original bucket

**Additional notes**
This feels unintuitive because the recurring task loses its workflow/category placement every time it is completed.

I have recorded a clip of my experience.

### Vikunja Version

2.3.0

### Browser and version

Firefox

Yes

### Screenshots

https://github.com/user-attachments/assets/ef05d58f-a8c1-4ae8-9e0a-3cc05579d2d7

nithinvarma411:
## PR Title
fix: recurring tasks return to original bucket when no default is set

### Summary
Fixes recurring tasks moving to the first bucket instead of their original bucket when a project has no default bucket configured. Tasks now maintain their workflow placement (e.g., "In Progress") across recurrences.

### How it's fixed
When a repeating task is marked done, two code paths handle bucket placement:

1. **API path** (`moveTaskToDefaultBuckets` in tasks.go): Now checks if `view.DefaultBucketID` is set. If it is, uses the default (existing behavior). If not, queries the task's current bucket and preserves it.

2. **Drag-to-done path** (`updateTaskBucket` in kanban_task_bucket.go): Now checks if `view.DefaultBucketID` is set. If it is, calls `getDefaultBucketID`. If not, restores the original bucket (`oldTaskBucket.BucketID`).

Both paths fall back to the first bucket only if the task has no current bucket assigned.

### Test coverage
Added test case: "moving a repeating task from a non-default bucket when no default bucket is configured preserves the original bucket"
- Temporarily removes default bucket from view
- Pre-positions task in middle bucket
- Drags to done bucket
- Asserts task returns to original bucket (not first bucket)

All existing tests pass. New test validates the fix prevents regression.

### Files changed
- `pkg/models/tasks.go` - Updated `moveTaskToDefaultBuckets()`
- `pkg/models/kanban_task_bucket.go` - Updated repeating task logic in `updateTaskBucket()`
- `pkg/models/kanban_task_bucket_test.go` - Added regression test

Closes/Resolves #2786

### DIFF
--- a/pkg/models/tasks.go
+++ b/pkg/models/tasks.go
@@ -1523,15 +1523,35 @@ func (t *Task) moveTaskToDoneBuckets(s *xorm.Session, a web.Auth, views []*Proje
 // and is used when a repeating task is marked done: repeating tasks
 // don't stay in the done bucket, so they should be routed back to
 // the default ("To-Do") bucket so the next iteration is visible there.
+// When no explicit default bucket is configured, the task is restored
+// to its original bucket so it remains in its original workflow column.
 func (t *Task) moveTaskToDefaultBuckets(s *xorm.Session, a web.Auth, views []*ProjectView) error {
 	for _, view := range views {
-		defaultBucketID, err := getDefaultBucketID(s, view)
-		if err != nil {
-			return err
+		var bucketID int64
+		var err error
+
+		if view.DefaultBucketID != 0 {
+			bucketID = view.DefaultBucketID
+		} else {
+			// No explicit default: preserve the task's current bucket.
+			currentTaskBucket := &TaskBucket{}
+			_, err = s.Where("task_id = ? AND project_view_id = ?", t.ID, view.ID).
+				Get(currentTaskBucket)
+			if err != nil {
+				return err
+			}
+			bucketID = currentTaskBucket.BucketID
+			if bucketID == 0 {
+				// Task not in any bucket yet — fall back to first by position.
+				bucketID, err = getDefaultBucketID(s, view)
+				if err != nil {
+					return err
+				}
+			}
 		}
 
 		tb := &TaskBucket{
-			BucketID:      defaultBucketID,
+			BucketID:      bucketID,
 			TaskID:        t.ID,
 			ProjectViewID: view.ID,
 			ProjectID:     t.ProjectID,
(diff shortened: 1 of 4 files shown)


pkg/models/tasks.go
```diff
@@ -1528,15 +1528,35 @@ func (t *Task) moveTaskToDoneBuckets(s *xorm.Session, a web.Auth, views []*Proje
...
+		var bucketID int64
+		var err error
+
+		if view.DefaultBucketID != 0 {
+			bucketID = view.DefaultBucketID
+		} else {
+			// No explicit default: preserve the task's current bucket.
+			currentTaskBucket := &TaskBucket{}
+			_, err = s.Where("task_id = ? AND project_view_id = ?", t.ID, view.ID).
+				Get(currentTaskBucket)
```
kolaente: Why not just do nothing in that case? In your solution we're fetching the current bucket and reinserting it, if we did nothing, the task should stay in its current bucket?

kolaente (changes requested):

pkg/models/tasks.go
nithinvarma411: You are right, I have updated the code.

kolaente:
Just to be clear, does this PR implement this proposal? https://github.com/go-vikunja/vikunja/issues/2594#issuecomment-4322990377

nithinvarma411:
@kolaente yes, this PR implements the same proposal

kolaente (approved):

PR #2832, opened 2026-06-03
refactor(frontend): centralize menu shortcuts

### LINKED ISSUE #542: Define shortcuts in one central place under constants
_Original issue by dpschen on 2023-04-01T16:22:05.000Z_

Instead of writing in a components template:

```vue
v-shortcut="'g o'"
```

we should instead import the shortcuts from a central place and write something like.

```vue
v-shortcut="SHORTCUTS.HOME"
```

This would also enable us in a second step to user define shortcuts.

[Original issue on Gitea](https://kolaente.dev/vikunja/vikunja/issues/2040)

kergma-lw:
Part of #542

This is a small first step toward centralizing shortcut definitions.

It moves the menu and navigation shortcut bindings into shared frontend constants and makes the shortcuts help reuse the same definitions, so the displayed keys and the actual bindings stay in sync.

I also added a small e2e test for the menu shortcut.

Checked with frontend lint, stylelint, unit tests, and the new e2e test. Also verified manually in the browser.

If this direction looks good, I can follow up with the remaining parts needed to fully close #542.


kergma-lw:
Friendly ping on this one in case it got buried.

No rush at all — this is a small first step for #542, and I’d be happy to follow up with the remaining parts if this direction looks good.

kergma-lw:
Hi @kolaente! Sorry for another ping — just want to make sure this didn’t get buried.  

Could you please do a quick sanity check that this direction (centralizing shortcut definitions) is acceptable? If yes, I’d be happy to continue and work on the remaining parts to fully close #542.  

I’m planning to keep contributing here, so I’d really appreciate a quick direction check to avoid spending time on the wrong approach. Thanks!

frontend/src/components/misc/keyboard-shortcuts/shortcuts.ts
```diff
@@ -1,8 +1,9 @@
 import type {RouteLocation} from 'vue-router'
 
+import {SHORTCUTS} from '@/constants/shortcuts'
 import {isAppleDevice} from '@/helpers/isAppleDevice'
 
-const ctrl = isAppleDevice() ? '⌘' : 'ctrl'
+const ctrl = SHORTCUTS.toggleMenu.keys[0]
```
kolaente: Do we still need the const here if the logic now lives in the central place?

frontend/src/components/misc/keyboard-shortcuts/shortcuts.ts
```diff
@@ -24,11 +25,11 @@ export const KEYBOARD_SHORTCUTS: ShortcutGroup[] = [
 		shortcuts: [
 			{
 				title: 'keyboardShortcuts.toggleMenu',
-				keys: [ctrl, 'e'],
+				keys: SHORTCUTS.toggleMenu.keys,
 			},
 			{
 				title: 'keyboardShortcuts.quickSearch',
-				keys: [ctrl, 'k'],
+				keys: [SHORTCUTS.toggleMenu.keys[0], 'k'],
```
kolaente: Why is this not centrally defined?

frontend/src/constants/shortcuts.ts
```diff
@@ -0,0 +1,47 @@
...
+	keys: string[]
+	combination?: 'then'
+}
+
+const ctrl = isAppleDevice() ? '⌘' : 'ctrl'
+
+export const SHORTCUTS = {
+	toggleMenu: {
+		binding: 'Mod+KeyE',
+		keys: [ctrl, 'e'],
```
kolaente: Binding and keys kind of duplicate each other, I wonder if we can make this simpler? But then again, they were that before, centralizing just makes it obvious. Not sure if we should keep it or change it.

kolaente (commented):
Hey sorry for the late reply. I've been busy with other work and Vikunja is getting many PRs right now.

The general direction here is good, I just have some comments below.

frontend/src/components/misc/keyboard-shortcuts/shortcuts.ts
kergma-lw: Good point — deriving it from toggleMenu was unnecessary. I kept the alias for readability since these help entries still use keys arrays.

frontend/src/components/misc/keyboard-shortcuts/shortcuts.ts
kergma-lw: Agreed — I was a bit too conservative there. quickSearch is now defined centrally as well. It probably also makes sense to move more of these entries toward referencing shared shortcut definitions instead of inline keys arrays.

frontend/src/constants/shortcuts.ts
kergma-lw: Agreed — but I’d rather not change both things at once. I’d prefer to finish the centralization first, then revisit the binding/keys duplication.

 That feels a bit too easy to turn into a broader refactor.

kergma-lw:
I pushed a follow-up that completes the v-shortcut centralization across the frontend.

Task detail shortcuts are now also defined in SHORTCUTS, and the related help entries were synced where they already map cleanly to those definitions.

I intentionally did not extend this into the useTaskDetailShortcuts string matching yet, since that seems like a slightly different layer than the declarative v-shortcut bindings.

kergma-lw:
I cleaned up the branch history and rebased this PR to make it easier to review.

kergma-lw:
Hi @kolaente — I’m still interested in contributing to Vikunja long-term, but long pauses like this are honestly quite demotivating.

I’d really appreciate a clear status on this PR. If this direction still makes sense, I’m happy to continue the remaining shortcut cleanup. If not, please tell me directly, so I can avoid spending time on work that is unlikely to be merged.

A brief yes / no / needs changes would already help a lot. Thanks.

kolaente:
Hey 👋 

Sorry about the lack of replies, I understand that the pauses are not very motivating. My main problem is that I get ~200 notifications from GitHub every day, so in order to keep my sanity I filter them away and look at PRs and issues when I have the time. Due to the rate of new PRs, that unfortunately means older PRs slip through. Additionally, the last two weeks I was on vacation (and haven't really opened GitHub much during that time).

I'll give this a proper review tomorrow, the direction seems to be still fine though.

kolaente (commented):
Looks good overall.

I do wonder, could we use the binding to derive the keys for display? The current solution means that we'll have to maintain both.

kergma-lw:
Thanks — I addressed that now.

The centralized shortcuts now use binding as the canonical source of truth, and the keyboard shortcuts help derives its display keys from that instead of keeping separate display definitions.

I also added focused tests around the display conversion, shortcut rendering, and the derived keyboard help data.

frontend/src/constants/shortcuts.ts
```diff
@@ -0,0 +1,73 @@
+import {isAppleDevice} from '@/helpers/isAppleDevice'
+
+export const PRIMARY_MODIFIER_KEY = isAppleDevice() ? '⌘' : 'ctrl'
+
+export const SHORTCUTS = {
+	toggleMenu: {
+		binding: 'Mod+KeyE',
```
kolaente: If we only ever have one child property `binding`, we might as well get rid of it entirely and move the string one level up

kolaente (changes requested):

kolaente:
Much better already!

frontend/src/constants/shortcuts.ts
kergma-lw: I removed the extra binding layer from SHORTCUTS and updated the related usages/tests.

kolaente (approved):
Thanks!

Issue #1875, opened 2025-11-25
Collapsible subtasks lists in task list view


sirlark:
### Description

It would be handy to be able to collapse and expand the sub-list of subtasks for tasks in the task list view, especially for tasks with lots of subtasks that end up larger than a single screen (especially on mobile devices).

kolaente:
This seems to be similar to this thread in the forum: https://community.vikunja.io/t/can-we-have-a-feature-where-we-can-collapse-tasks-and-its-subtasks-in-list-view-similar-to-how-projects-and-sub-project-view-work/2102/11

sirlark:
Yes, exactly. I didn't see it as an issue in the open list, and didn't check the forum. I assumed the issue list was the correct place to add feature requests.

PR #2907, opened 2026-06-11
feat: collapsible subtasks in list view 🤖🤖🤖

oneclawbot-prog:
## Summary

Adds collapsible/expandable subtask functionality to the list view, addressing #1875.

## Changes

- **Single task collapse**: Each parent task with subtasks now shows a chevron button (▼/▶) on the right side. Clicking it collapses/expands its subtasks.
- **Global collapse/expand**: A button in the toolbar allows collapsing or expanding all subtasks at once.
- **Responsive design**: On narrow screens (<768px), only the right-side chevron is shown to save space. On wider screens, chevrons appear on both sides.
- **Conditional rendering**: Only tasks with actual subtasks display the collapse button.

## Files Changed

- `frontend/src/components/tasks/partials/SingleTaskInProject.vue` — Added collapse toggle button, `isCollapsed` state, and CSS styles
- `frontend/src/components/project/views/ProjectList.vue` — Added global collapse/expand button with `provide/inject` for state propagation

### 1. Initial state - all subtasks expanded
All expanded

### 2. All subtasks expanded (wide view)
All expanded wide

### 3. Single task collapsed (parent task 2 collapsed, parent task 1 expanded)
Single collapse

### 4. Wide screen (>768px) - chevrons on both sides
Wide screen

### 5. Narrow screen (<768px) - only right-side chevron
Narrow screen

## Testing

Tested locally with:
- Parent tasks with multiple subtasks → collapse/expand works correctly
- Tasks without subtasks → no collapse button shown
- Global collapse button → all subtasks collapse/expand
- Responsive behavior → left chevron hidden on narrow screens

## Notes

- Uses `chevron-down` icon with CSS `rotate(-90deg)` for collapsed state (consistent with existing icon set)
- No backend changes required — uses existing `relatedTasks.subtask` data structure
- Zero new dependencies


frontend/src/components/tasks/partials/SingleTaskInProject.vue
```diff
@@ -597,6 +633,41 @@ defineExpose({
...
+		transform: rotate(-90deg);
+	}
+}
+
+.collapse-toggle-left {
+	margin-inline-start: 0;
+	margin-inline-end: 0.5rem;
+	order: -1;
+
+	// 窄屏隐藏左侧按钮，只保留右侧
```
kolaente: please remove this

kolaente (changes requested):
Please fix the lint

oneclawbot-prog:
Thanks for the review! Pushed two commits:

**`7d60e655` — lint + review feedback**

- removed the comment you flagged
- fixed the lint errors: unused `hasSubtasks` computed, `defineEmits` placement, attribute order, and `media-feature-range-notation`
- the toggle labels were hardcoded English strings — they now go through i18n (`task.collapseSubtasks` and friends in `en.json`)
- the collapse-all state was provided under a string key as a plain boolean, so a task the user expanded on its own could fall out of sync with the next toggle. It now uses a typed `InjectionKey` carrying a token that is bumped on every click, following the pattern in `commentReplyContext.ts`.
- the chevron rotation was applied to the whole button, rotating its border and hover background along with it — it is on the svg only now

**`9ae7ee69` — e2e coverage**

Three cases in `project-view-list.spec.ts`: the single toggle, collapse-all including the resync case above, and the narrow-screen layout.

Locally `pnpm lint` and `lint:styles` are clean, `test:unit` passes (1353), and the full `project-view-list` e2e suite passes (16). The branch is also merged up to current main, so it should no longer be out of date.

The CI runs are sitting at `action_required` — they need your approval to start.

One open question: the collapse state isn't persisted, so it resets on reload or when switching views. Is that wanted for #1875, and if so should it live in localStorage or in the user's frontend settings? Happy to add it either way.

frontend/src/components/tasks/partials/SingleTaskInProject.vue
oneclawbot-prog: Removed, thanks.

oneclawbot-prog:
Added the persistence in `6772ee88` — collapsed tasks are kept in local storage under `collapsedSubtasks`, following the pattern `ProjectsNavigationItem.vue` already uses for the child projects in the navigation. Only collapsed ids are stored so the entry does not grow with every task that was ever expanded.

The shared state let me drop the provide/inject plumbing entirely. Two things fell out of that:

- collapse-all now derives its label from whether everything is actually collapsed, so expanding a single task no longer takes two clicks to collapse everything again
- the button is hidden when no task in the view has subtasks

E2E is up to 5 cases now (18 passing in the `project-view-list` suite), including the reload and the empty-project case.

oneclawbot-prog:
One more in `f2450757` — the collapse all button was a bare `<button>` with its own box styles, which did not line up with the sort and filter buttons next to it (visible border, slightly shorter). It is an `XButton` with `variant="secondary"` now, the same as those two, so the hand rolled box is gone and only the rotation and the active colour are left: -23/+4 lines.

PR #2916, opened 2026-06-15
feat: filter tasks by creator username

BenLocal:
### Summary
  Adds a `creator` filter field that matches tasks by the **creator's username**, mirroring the existing `assignees` filter. Until
  now, tasks could only be filtered by their creator via `created_by_id` (a numeric user ID) — filtering by `creator` was rejected
  with `ErrInvalidTaskField` (4016). This adds the symmetric, human-friendly form.

  ### Motivation
  `assignees = 'demo'` already filters by username, but there was no equivalent for the task creator — only `created_by_id = 5`
  (numeric). This closes that gap with `creator = 'demo'`.

  ### What changed

  **Backend (`pkg/models`, shared by `/api/v1` and `/api/v2`)**
  - Whitelist `creator` as a filterable field (`validateTaskField`).
  - Resolve its value as comma-separated usernames (`getNativeValueForTaskField`), reusing the `assignees` code path (no numeric
  coercion).
  - Add a `users` sub-table `EXISTS` subquery in `subTableFilters`:
    ```sql
    EXISTS (SELECT 1 FROM users WHERE tasks.created_by_id = users.id AND username IN (...))
    ```
    `!=` / `not in` produce `NOT EXISTS`.
  - Precise-only (`=`, `!=`, `in`, `not in`); `like` is ignored, consistent with `assignees`.

  Since the filter logic lives in the shared model layer, **both API versions get it automatically** — no route changes.

**Frontend**
  - Register `creator` as a recognized filter field (syntax highlighting + query parsing).
  - Wire **username autocomplete** for `creator`, reusing the assignee user-search path (project members in a project context, all
  users for saved filters).
  - Document `creator` in the filter help panel, with an example.
  - Usernames pass through to the API unchanged (the backend resolves them), exactly like `assignees`.

  ### Usage
  ```
  creator = 'demo'
  creator in 'alice,bob'        # or: creator ?= 'alice,bob'
  creator != 'demo'
  done = false && creator = 'alice'
  ```
  `created_by_id = 5` (numeric) keeps working unchanged.

  ### Notes
  - **No database migration.** No schema changes — `tasks.created_by_id` and `users.username` already exist; this only adds a query
  path. Cross-DB safe (MySQL/Postgres/SQLite); reuses the same `users.username` column the `assignees` filter already relies on.
  - **Non-breaking.** New field only; existing `created_by_id` numeric filtering is unaffected.
  - **Filter-only.** `creator` is filterable but not a sort key (sorting by username would require a join); `created_by_id` remains
  sortable.

  ### Tests
  - **Backend:** new cases in `TestParseFilter` (creator resolves to usernames, not an int64 id) and `TestTaskCollection_ReadAll`
  (`=`, `in`, `!=` → NOT EXISTS, and unknown-username → empty result). Full `pkg/models` suite passes.
  - **Frontend:** `eslint` clean; `vitest` `filters.test.ts` passes (including a `creator` username pass-through case).


frontend/src/components/input/filter/FilterAutocomplete.ts
```diff
@@ -338,7 +339,7 @@ export default Extension.create<FilterAutocompleteOptions>({
 
 					if (LABEL_FIELDS.includes(field)) {
 						fieldType = 'labels'
-					} else if (ASSIGNEE_FIELDS.includes(field)) {
+					} else if (ASSIGNEE_FIELDS.includes(field) || CREATOR_FIELDS.includes(field)) {
 						fieldType = 'assignees'
```
kolaente: This will set the field type to assignees for creators as well, should we make that field type creator? or more general like `users`

kolaente:
auggie review

frontend/src/components/input/filter/FilterAutocomplete.ts
BenLocal: I went with the more general users type rather than a dedicated creator one, since assignees and creators autocomplete against the exact same thing (users
  matched by username), so a separate type would just duplicate the identical fetch/display/insert branches.

  Renamed the autocomplete field type 'assignees' → 'users' across FilterAutocomplete.ts and FilterCommandsList.vue. The mapping now reads:

```
  } else if (ASSIGNEE_FIELDS.includes(field) || CREATOR_FIELDS.includes(field)) {
      fieldType = 'users'
```

pkg/models/task_collection_filter.go
```diff
@@ -360,7 +360,7 @@ func getNativeValueForTaskField(fieldName string, comparator taskFilterComparato
 
 	realFieldName := strings.ReplaceAll(strcase.ToCamel(fieldName), "Id", "ID")
 
-	if realFieldName == "Assignees" {
+	if realFieldName == "Assignees" || realFieldName == "Creator" {
 		vals := strings.Split(value, ",")
```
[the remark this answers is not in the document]
BenLocal: fixed. Each username is now TrimSpaced after the split, so multi-value filters like creator in 'user1, user6' match correctly (the leading space previously made the
  second username never match) — same way the numeric in path tolerates whitespace.

  I also drop empty entries while I'm at it, so stray/trailing commas like 'user1, , user6,' don't emit useless username = '' predicates:

```golang
  if realFieldName == "Assignees" || realFieldName == "Creator" {
        vals := strings.Split(value, ",")
        valueSlice := make([]string, 0, len(vals))
        for _, val := range vals {
                val = strings.TrimSpace(val)
                if val == "" {
                        continue
                }
                valueSlice = append(valueSlice, val)
        }
        return nil, valueSlice, nil
  }
```

pkg/models/task_collection_filter.go
```diff
@@ -428,9 +428,16 @@ func getNativeValueForTaskField(fieldName string, comparator taskFilterComparato
 
 	realFieldName := strings.ReplaceAll(strcase.ToCamel(fieldName), "Id", "ID")
 
-	if realFieldName == "Assignees" {
+	if realFieldName == "Assignees" || realFieldName == "Creator" {
```
kolaente: The field should match the field that's returned by the api which is `CreatedBy`

kolaente (changes requested):
Please also fix the failing lint.

BenLocal:
I’ve addressed both comments:

  - Renamed the filter field to match the existing API model: createdBy in the frontend, created_by in the API, and CreatedBy in Go.
  - Removed the unused initial assignment to userSuggestions, fixing the no-useless-assignment lint error.

pkg/models/task_collection_filter.go
BenLocal: Renamed the filter field to match the existing API model: createdBy in the frontend, created_by in the API, and CreatedBy in Go.

kolaente (approved):

PR #2920, opened 2026-06-16
fix: GCS signing issue

szuro:
There are signing issues when using Google Cloud Storage as an S3 backend.
For whatever reason, GCS modifies `Accept-Encoding` header.
This causes Vikunja to fail even with `files.s3.disablesigning: true`

The maintainers of aws-sdk-go refuse to address this, hence this PR.

More details can be found here: https://github.com/aws/aws-sdk-go-v2/issues/1816#issuecomment-4526752227

### DIFF
--- a/pkg/files/filehandling.go
+++ b/pkg/files/filehandling.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


kolaente (approved):
Thanks!

kolaente:
Please fix the lint

PR #2935, opened 2026-06-18
fix: blur quick add input on escape

probecat:
When the quick add input is focused, pressing Esc key now unfocuses the input. This frees up keyboard shortcuts without having to reach for mouse.


kolaente:
Sorry, what exactly are you trying to achieve? When I press esc while in the modal, the modal closes.

probecat:
This change only concerns "add task" input.

Steps to reproduce:
1. Open overview page
2. Focus the add task input
3. Press `Esc` key - does not unfocus

kolaente:
okay, makes sense to fix this then

kolaente (approved):

PR #2939, opened 2026-06-19
feat: include subproject tasks (rebase of #2272)

mds08011:
This PR revives @JamesBatten's work from #2272  so it is ready for merge. Supersedes and closes #2272 .

Adds an `include_subprojects` flag to the task collection, so a project's views can show the tasks of its subprojects alongside its own.

**Flag**

It is a field on `TaskCollection`, exposed as a query parameter on both `/api/v1` and `/api/v2` (`TaskListQueryParams`). The per-view default is stored inside the view's existing filter JSON with `omitempty`, so views that never set it serialise exactly as before and no database migration is needed.

The filter popup does not write to the view. Toggling it there sets an `?include_subprojects=` query param, and setting it back to the view's default drops the param again, so the popup stays local to the url like the rest of its controls.

**Resolving subprojects**

Descendants are resolved in a single recursive CTE that walks down from the requested project, similar to `SetArchiveStateForProjectDescendants`, intersected with the existing `accessibleProjectIDsSubquery`. Archived subtrees are pruned during the recursion. Only ids are selected, since every caller reads nothing but `Project.ID`.

Read permissions are inherited down `checkPermissionsForProjects` resolves a project's permission, so a descendant of a project the user can read is always readable.


frontend/src/components/project/views/ViewEditForm.vue
```diff
@@ -189,13 +190,21 @@ function handleBubbleSave() {
...
 		</div>
 
-		<div class="field mbe-3">
-			<FancyCheckbox
-				v-model="view.filter.filter_include_nulls"
-			>
-				{{ $t('filters.attributes.includeNulls') }}
-			</FancyCheckbox>
-		</div>
+<FancyCheckbox
```
kolaente: this needs indention

pkg/models/project_view.go
```diff
@@ -161,8 +161,11 @@ type ProjectView struct {
 
 	// The filter query to match tasks by. Check out https://vikunja.io/docs/filters for a full explanation.
 	Filter *TaskCollection `xorm:"json null default null" query:"filter" json:"filter" doc:"The filter query used to match tasks shown in this view. See https://vikunja.io/docs/filters."`
-	// The position of this view in the list. The list of all views will be sorted by this parameter.
-	Position float64 `xorm:"double null" json:"position" doc:"The position of this view in the project's list of views. Views are sorted ascending by this value."`
+	// The position of this view in the list. The list of all views will be sorted by this position.
+// The position of this view in the list. The list of all views will be sorted by this position.
```
kolaente: This is duplicate, please remove one comment

kolaente:
auggie review

frontend/src/components/project/views/ProjectGantt.vue
```diff
@@ -109,7 +127,38 @@ const {
 	isLoading,
 	addTask,
 	updateTask,
-} = useGanttFilters(route, viewId)
+} = useGanttFilters(route, viewId, includeSubprojects)
+
+async function updateIncludeSubprojects(newValue: boolean) {
```
kolaente: This appears duplicated with the function of the same name in the filter popup component, move it to the project store and call it from both places

frontend/src/components/project/views/ProjectList.vue
```diff
@@ -200,7 +204,7 @@ onMounted(async () => {
 	ctaVisible.value = true
 })
 
-const canDragTasks = computed(() => canWrite.value || isSavedFilter(project.value))
+const canDragTasks = computed(() => (canWrite.value || isSavedFilter(project.value)) && !includeSubprojects.value)
```
kolaente: Why not allow dragging when tasks from sub projects are shown?

frontend/src/components/project/views/ViewEditForm.vue
kolaente: Please also wrap it in a field, like the include nulls checkbox

frontend/src/components/tasks/partials/SingleTaskInProject.vue
```diff
@@ -292,6 +292,14 @@ const currentProject = computed(() => {
 	} : baseStore.currentProject
 })
 
+const projectLabel = computed(() => {
+	if (!project.value) {
+		return ''
```
kolaente: If there is no project, we shoul hide the router link instead of not showing an empty label.

pkg/models/project_view.go
```diff
@@ -161,8 +161,11 @@ type ProjectView struct {
...
+	Position float64 `xorm:"double null" json:"position" doc:"The position of this view in the list. The list of all views will be sorted by this position."`
+	// If enabled, this view also includes tasks from all descendant subprojects.
+	IncludeSubprojects bool `xorm:"default false not null" json:"include_subprojects"`
```
kolaente: This should be part of the filter settings and stored as such. Conceptually it belongs in the Task Collection, not in the view.

This also means that we don't need the migration.

pkg/models/task_collection.go
```diff
@@ -201,7 +207,53 @@ func getRelevantProjectsFromCollection(s *xorm.Session, a web.Auth, tf *TaskColl
...
+
+		relevantProjects = append(relevantProjects, &Project{ID: projectID})
+	}
+
+	return relevantProjects, nil
+}
+
+func getProjectAndDescendantIDs(s *xorm.Session, rootProjectID int64) (projectIDs []int64, err error) {
+	err = s.SQL(`
+WITH RECURSIVE descendant_projects (id) AS (
```
kolaente: Please check if we could instead use the query that fetches all projects the user has access to with a parameter.

kolaente (changes requested):

kolaente:
Please take a look at the failing ci.

kolaente:
Thanks for taking the effort to rebase this! Does that mean you're also taking over the PR?

frontend/src/components/project/views/ViewEditForm.vue
mds08011: Should be fixed in latest commit.
https://github.com/go-vikunja/vikunja/pull/2939/changes/f833300ebc2c86ff8d89ef8a5971fd7556a3703b

pkg/models/project_view.go
mds08011: Fixed in latest commit.

frontend/src/components/project/views/ProjectGantt.vue
mds08011: Fixed in latest commit.

frontend/src/components/project/views/ProjectList.vue
mds08011: Fixed in latest commit.

frontend/src/components/tasks/partials/SingleTaskInProject.vue
mds08011: Fixed in latest commit.

mds08011:
> Thanks for taking the effort to rebase this! Does that mean you're also taking over the PR?

Hey @kolaente and @JamesBatten, the UI and syntax fixes are done, so the CI should pass. The database changes are going to stretch my abilities, but I'll dive into moving the variable to the Task Collection this weekend to give it a try.

kolaente:
Unfortunatly, CI is failing again.

Happy to assist on the DB changes.

pkg/models/project_view.go
mds08011: Moved the variable to the Task Collection filter and removed the migration.

pkg/models/task_collection.go
mds08011: I removed the recursive CTE and swapped for project fetch query instead.

frontend/src/components/project/views/ViewEditForm.vue
```diff
@@ -13,6 +13,7 @@ import FancyCheckbox from '@/components/input/FancyCheckbox.vue'
 import FilterInputDocs from '@/components/input/filter/FilterInputDocs.vue'
 import FilterInput from '@/components/input/filter/FilterInput.vue'
 import FormField from '@/components/input/FormField.vue'
+import FancyCheckbox from '@/components/input/FancyCheckbox.vue'
```
Mushkrem: Duplicate import

mds08011: Fixed, thank you! https://github.com/go-vikunja/vikunja/pull/2939/commits/3653fd3d75a18ae928c61ed440cf1575bd07e3a3

Mushkrem:
I tried building this PR myself two days ago, and after removing the duplicate line the project was built correctly, but the buttons ,,Include subprojects" were turning off milliseconds after interacting with them.

I have no knowledge of this codebase, so I tried stitching together whatever seemed to match, but there seemed to be many inconsistencies, namely in files:
- `frontend/src/modelTypes/ISavedFilter.ts` (line 35) - the IFilters interface (which seems to be meant for exactly this kind of per-view filter data) has no include_subprojects field at all.
- `frontend/src/models/projectView.ts` (line 18) - sets a default includeSubprojects = false at the top level instead of inside the nested filter object.
- `pkg/models/project_view.go` (line 163) - Filter *TaskCollection seems to be the only place IncludeSubprojects is actually reachable from a view (Filter.IncludeSubprojects), but there's no top-level IncludeSubprojects field on ProjectView.
- `pkg/models/task_collection.go` (line 35) - IncludeSubprojects is declared here as a nested field, which doesn't match how the frontend components are reading/writing it.
- `frontend/src/components/project/views/ProjectGantt.vue, FilterPopup.vue, ProjectTable.vue, ProjectList.vue, ProjectKanban.vue, ViewEditForm.vue` - all read/write includeSubprojects as a top-level property instead of nested under filter.

Given all other toggleable filters append a ?query=value to the URL and this one didn't, I assumed this was supposed to be handled client-side rather than persisted to the DB, so I moved the whole thing over to work that way instead (as is shown above), but that didn't work either and it just stopped doing anything whatsoever, but buttons did persist (didn't turn off)

I have no idea what the "correct" intended design is here (persisted view setting vs. client-side query param), and honestly at this point looking at the PR, I'm not sure the PR itself has settled on one either.

mds08011:
Thanks for taking the time to build and test this! The intended design (per Kolaente's request) is a persisted view setting that lives inside the Task Collection filter. I pushed a revision so that the view components now correctly read and write the nested `filter.include_subprojects`. It seems to be building cleanly now, and I included a gif below testing it out.

**View with filter setting**





**Setting**




**Toggle in action**

mds08011:
Small follow-up: moved the duplicated `updateIncludeSubprojects` into the project store, and changed the Gantt view to show the source project as a muted label on the bar like list/kanban do.

HarshPatel5940:
damnm, i was looking for this feature.

did we ensure that members of parent project can see tasks of child projects and its own, while child project members can only see their own

mds08011:
> did we ensure that members of parent project can see tasks of child projects and its own, while child project members can only see their own

Yes this is the case. I added a test to check it as well.

dodzod:
Hi - just bumping this in case anyone has time to finalise it - it looks like the only required test failure is "file is not properly formatted" which hopefully should be an easy fix, in fact it looks like it is just a surplus blank line. Sadly I don't have the practical development skills to add a commit, run the tests etc. or I would do so.

mds08011:
@dodzod I removed the blank line you noted. I think I've addressed all the feedback in this PR, but let me know if you see anything else I missed.

dodzod:
Thanks very much! I guess it's now for the maintainers to review?

frontend/src/components/project/views/ViewEditForm.vue
```diff
@@ -65,9 +65,15 @@ onBeforeMount(() => {
 		// match the API query param format. We check both key forms here to handle
 		// data coming from either the API response (camelCased by assignData) or
 		// from a freshly constructed filter object (snake_case).
-		filter.filter_include_nulls = filterInput.filter_include_nulls
+		filter.filterIncludeNulls = filterInput.filter_include_nulls
 			?? (filterInput as Record<string, unknown>).filterIncludeNulls as boolean
 			?? false
+		filter.filter_include_nulls = filter.filterIncludeNulls
```
kolaente: why did you add `filterIncludeNulls` and `filter_include_nulls`? Similarily for other properties

frontend/src/components/project/partials/FilterPopup.vue
```diff
@@ -89,6 +93,26 @@ function showResults() {
...
+const isProjectView = computed(() => Boolean(props.projectId && props.projectId > 0 && props.viewId))
+
+const includeSubprojects = computed(() => currentView.value?.filter?.includeSubprojects ?? currentView.value?.filter?.include_subprojects ?? false)
+
+async function updateIncludeSubprojects(newValue: boolean) {
+	if (!currentView.value) {
+		return
+	}
+
+	await projectStore.updateViewIncludeSubprojects(currentView.value, newValue)
```
kolaente: When toggling the checkbox, the value should only be persisted in a query string (use https://vueuse.org/router/useRouteQuery/#useroutequery). Updating the view the user is in is inconsistend with the rest of the filter popup and confusing.

pkg/models/task_collection.go
```diff
@@ -31,6 +31,8 @@ import (
 type TaskCollection struct {
 	ProjectID     int64 `param:"project" json:"-"`
 	ProjectViewID int64 `param:"view" json:"-"`
+	// If set to true, tasks from all descendant subprojects will also be returned.
```
kolaente: this comment is redundant

pkg/models/task_collection.go
```diff
@@ -209,7 +244,41 @@ func getRelevantProjectsFromCollection(s *xorm.Session, a web.Auth, tf *TaskColl
 		}
 	}
 
-	return []*Project{{ID: tf.ProjectID}}, nil
+	if !tf.IncludeSubprojects || tf.ProjectID <= 0 {
+		return []*Project{{ID: tf.ProjectID}}, nil
+	}
+
+	allProjects, _, _, err := getRawProjectsForUser(s, &projectOptions{
```
kolaente: Fetching all projects to then look for sub projects is pretty wasteful. Please use one of the existing CTEs for that.

pkg/models/task_collection_test.go
```diff
@@ -1572,6 +1573,84 @@ func TestTaskCollection_ReadAll(t *testing.T) {
...
+			args: args{
+				a: &user.User{ID: 1},
+			},
+			want: []*Task{
+				task39,
+			},
+			wantErr: false,
+		},
+		{
+			name: "project tasks including subprojects with no project access",
```
kolaente: What does "no access" mean here?

If the flag is provided, it should return all tasks from all sub projects the user has access to, not fail when there is one sub project the user does not have access to.

frontend/src/components/project/views/ProjectGantt.vue
```diff
@@ -38,6 +38,15 @@
 					>
 						{{ $t('task.show.noDates') }}
 					</FancyCheckbox>
+					<FancyCheckbox
```
kolaente, line 41: When I click on this checkbox, it resets to unchecked immediatly, making it unusable

kolaente (changes requested):
Thanks for the updates, sorry for the long silence from me.

I noticed a few things:

- Please rebase on main, that should make the lint issue go away. If not, look into it and fix it.
- When moving a task from a sub project to the parent project via drag and drop, it disappears instead of staying visible. Please fix that.
- In Kanban, showing tasks from sub projects does not really work because the buckets are different. An easy way out for now would be to just not make that option available there? Open for other suggestions.
In my testing, tasks from the sub project showed up in the Done bucket, even though they were not done. That's wrong.
- Updating positions in the list view needs some thought. When I see tasks from sub projects and drag them around, the position should be updated, but what should it be anchored to? The surrounding tasks or the neighbouring tasks from any project that is currently visible?

frontend/src/components/project/views/ViewEditForm.vue
mds08011: I removed this and the other duplicated keys. `ProjectViewModel` now sets the filter back to snake_case, same as `SavedFilterModel` already does for its own filters.

frontend/src/components/project/partials/FilterPopup.vue
mds08011: Fixed. The toggle has a `?include_subprojects=` query, and no longer writes to the view. The flag stays as the per-view default.

That also settles the duplicate helper from earlier: `updateViewIncludeSubprojects` is gone from the project store, since nothing writes the flag to the view any more.

pkg/models/task_collection.go
mds08011: Removed

pkg/models/task_collection.go
mds08011: Replaced with a recursive CTE that walks down from the requested project, in the same shape as `SetArchiveStateForProjectDescendants`, intersected with `accessibleProjectIDsSubquery` so subprojects the user cannot see are skipped. Archived subtrees are pruned in the recursion. Only the ids are selected, since every caller reads nothing but `Project.ID`.

pkg/models/task_collection_test.go
mds08011: Yes, the name was misleading. That case is about the requested project, not its subprojects: user 14 has no access to project 32 at all, so the request is denied the same as it would be without the flag. Renamed to say so.

On subprojects, `checkPermissionsForProjects` resolves a project's permission by walking up to its ancestors, so a descendant of a project you can read is always readable too. A subproject only drops out of the result when it is archived.

Both cases are covered:
- `TestTaskCollection_ReadAll_IncludeSubprojectsInheritedAccess` for a subproject reached only through its parent being included.
- `TestTaskCollection_ReadAll_IncludeSubprojectsPartialAccess` for one that cannot be reached being skipped while the request still succeeds.

frontend/src/components/project/views/ProjectGantt.vue
mds08011: Fixed, there were two causes. The checkbox read the view from the base store while the update wrote to the project store, so the optimistic update never showed. That path is gone now the toggle is a query param. On top of that, `useRouteFilters` rebuilds the gantt url from the gantt filters alone and dropped every other query param on the next filter change, so the flag never survived. It now carries over the params it doesn't own.

mds08011:
**Drag and drop from a subproject**

With subprojects included the `ProjectList` removal is skipped and the list is reloaded after the move, so the task stays visible under its new project, and still disappears when it is moved somewhere this view doesn't cover.

**Kanban**

Buckets belong to a single view, so a subproject task has no bucket in the parent's view and lands in the default or done bucket regardless of its state. The toggle is hidden for kanban views in both the view settings and the filter popup, and `normalizeIncludeSubprojects` clears the flag server side so a request that sets it anyway is ignored.

**Reordering in list view**

I disabled reordering while subprojects are included, rather than anchoring.

Positions are stored per view (`task_positions.project_view_id`), so a task in a subproject only has a position in its own project's views. The API already skipped the position sort when `include_subprojects` is set. Sorting within the list is now turned off in that mode, so the list no longer has an order it cannot store.

Anchoring to neighbours from the same project gives an order within each project but not between them, since positions from different projects are drawn from unrelated number lines, so a drop between two tasks from different subprojects can't be expressed. Anchoring to any visible neighbour needs a position for every subproject task in the parent's view, which means writing rows for tasks that don't belong to that view and keeping them in sync as tasks move, get created, or the subproject set changes. If you'd rather have cross-project ordering, materialising positions in the parent view behind a migration would be the deliberate way to do it.

Dragging itself stays enabled. Only sortable's own sorting is off, so a task can still be dragged onto another project in the sidebar while subprojects are shown.

frontend/src/components/project/partials/Filters.vue
```diff
@@ -29,6 +29,14 @@
 			>
 				{{ $t('filters.attributes.includeNulls') }}
 			</FancyCheckbox>
+			<FancyCheckbox
+				v-if="showIncludeSubprojectsToggle"
```
kolaente, line 33: when a view has "include subprojects" configured in the view settings and I uncheck this checkbox, setting `?include_subprojects=false` via query, the subprojects are still shown. To make this better from a UX pov, disable the checkbox with an explaining tooltip when include subprojects is enabled in the view settings.

frontend/src/components/project/views/ProjectList.vue
```diff
@@ -174,10 +183,14 @@ watch(
 
 const isPositionSorting = computed(() => 'position' in sortByParam.value)
 
-const baseStore = useBaseStore()
+// Positions are stored per view, and a subproject's task has no position in this
+// project's view - the api drops the position sort for that reason. Dragging stays
+// enabled so a task can still be moved to another project, but it cannot be
+// reordered into a position that could never be stored.
+const canReorderTasks = computed(() => isPositionSorting.value && !includeSubprojects.value)
```
kolaente, line 187: yes, but this also prevents reordering tasks that are not coming from a subprojects. That's confusing.

pkg/models/task_collection.go
```diff
@@ -209,7 +252,52 @@ func getRelevantProjectsFromCollection(s *xorm.Session, a web.Auth, tf *TaskColl
...
+		relevantProjects = append(relevantProjects, &Project{ID: id})
+	}
+
+	return relevantProjects, nil
+}
+
+// getAccessibleSubprojectIDs walks down the project hierarchy from parentProjectID and
+// returns every descendant the auth can access. Subprojects without access are skipped
+// rather than failing the whole request.
+func getAccessibleSubprojectIDs(s *xorm.Session, a web.Auth, parentProjectID int64) (ids []int64, err error) {
```
kolaente, line 267: Please check what https://github.com/go-vikunja/vikunja/pull/3655 added and don't reinvent the wheel here

kolaente (changes requested):
Also cleanup the code comments. Almost all of them are too long and don't add value.

frontend/src/components/project/views/ProjectList.vue
mds08011: I agree that split is confusing. I think saved filter views could work for this. `ensureTaskPositionsForSavedFilterView` gives a position to tasks with no row in the view it is handed, and it is generic over `projects` and `view` , only the call site is saved-filter specific. Running it when `include_subprojects` is set would give subproject tasks a position in the parent's view. That is not touched by the new validation, since it inserts directly rather than going through `CanUpdate`.

Dragging and `validate position target views` has positioning a task in a view whose project is not the task's is now refused, saved filters aside. So with positions in place, the project's own tasks would drag again, while a subproject's task would not.

Would you rather it stayed that way, or should subproject tasks drag too, with a second exception for a task whose project is a descendant of the view's?

mds08011:
You were right that Filter popup checkbox did nothing so it's now disabled when the view enables the flag, and clearing the filters leaves it alone.

For #3655 I switched to `accessibleProjectIDsCond`, so the access half comes from the per-session memo instead of the recursive subquery.

I also replied to your `ProjectList.vue` comment with a question.

sylphia-nox:
hi team, found this issue as I was looking for a way to do this, is there anything I could do to help push this forward?

Issue #2826, opened 2026-06-01
Feature Request: Improve Upcoming View for Scheduled Tasks


wizand0:
### Pre-submission checklist

- [x] I have searched for existing open or closed issue reports with the same problem.

### Description

**Problem**
The Upcoming view currently only displays tasks that have a Due Date assigned.

Tasks that are planned using Start Date and End Date remain invisible in Upcoming, even when:
they have a clearly defined schedule,
they are shown in Gantt view,
they have an upcoming End Date,
they are recurring tasks.

As a result, users who organize work using Start Date and End Date may miss important upcoming tasks unless they manually assign a Due Date to every task.

**Suggested Improvement**
Consider adding an optional setting or alternative Upcoming mode that can also include:
tasks with an End Date when Due Date is empty;
recurring tasks that have a scheduled Start Date or End Date but no Due Date.

This would preserve the current behavior while providing better support for users who schedule tasks using date ranges rather than deadlines.

Example
Task:
Start Date: June 1
End Date: June 10
Due Date: not set

Currently:
visible in Gantt view
not visible in Upcoming

Expected:
visible in Upcoming because it has an approaching planned completion date.
**Benefits**
Better support for project planning workflows
Better visibility of recurring scheduled tasks
Reduced need to duplicate information by setting both End Date and Due Date

### Vikunja Version

2.3.0

### Can you reproduce the bug on the Vikunja demo site?

Yes

kolaente:
What about using a saved filter for this?

Issue #2695, opened 2026-04-25
Upcoming tasks should show start date tasks


melroy89:
### Pre-submission checklist

- [x] I have searched for existing open or closed issue reports with the same problem.

### Description

When a task has only a **start date** (and/or end date), this task is currently _not_ being displayed in the "Upcoming" overview. Which is strange and causing issues by users who are now forgetting tasks to complete. 

Apparently only "due date" is getting displayed under Upcoming. However, this needs to be extended to either due date or start date (depending which date is filled-in, and then give prio on the due date in case both are provided). Possibly even end-date can be added to the search query.

As you can imagine, users are not going to add both start _and_ due dates for each task. I believe a start date (and/or end date) is also a date.

### Rationale

Start & end dates are mainly used for tasks on the gantt chart. Users are _not_ going to fill in a 3rd date (due date). Meaning, you want at least the start date to be used as well in the Upcoming items overview (in case due date is missing).

### Vikunja Version

2.3.0

### Browser and version

Firefox 150.0

### Can you reproduce the bug on the Vikunja demo site?

Yes

### Screenshots

**Incorrectly** showing nothing to do:

<img width="990" height="158" alt="Image" src="https://github.com/user-attachments/assets/dc148d82-f0c3-4dc0-892f-c0c53e85f2d1" />

(yes the item will show when checking "Show tasks without date", but the Task DOES has a date)

melroy89:
I thought it was a bug, since people expected to see the tasks listed in "upcoming" when you assign a date to it. But yes, you can also see it as "new feature" I guess.

Issue #1364, opened 2025-08-31
Week and month calendar view in overview, like todoist


9165:
### Description

<img width="1180" height="1100" alt="Image" src="https://github.com/user-attachments/assets/62c79aa1-b069-4c8e-acf4-0b3d9db91bcb" />

netkela:
+1

IaMs12:
Agreed. To add on, being able to drag and move around the tasks as well in the week and monthly view would be great. Similar to Ganntt we have now but instead weekly and monthly versions.

Ultimately being able to pull in other calendars would be awesome to help plan when tasks can be moved around in between.

ECidno:
Calender View would be great but still not implemented?

PR #2957, opened 2026-06-23
Feature: Planner, adds cross-project calendar planner view

Marlon154:
## Summary

This adds a standalone /planner page where you can actually block out your week. The existing views (List, Kanban, Gantt, Table) are great for organizing tasks, but none of them let you say "I'll do this from 2–3pm on Tuesday." This fills that gap.

It's frontend-only — no API or DB changes. It reuses the existing start_date/end_date/due_date fields and the v2 tasks endpoint, so a scheduled block is just a task with a start and end time.

## Features

- Week and day views on an hour grid, with a zoom control and a current-time line
- Drag an unscheduled task from the sidebar onto a time slot; drag blocks to move them (across days too), resize them, or drag them back to the sidebar to unschedule
- An all-day row for due-only / all-day tasks, with drag in and out of it
- Filter the sidebar with the normal Vikunja filter (it's cross-project)
- View settings — working hours, slot size, default duration, full week vs. next 7 days, show/hide done — saved per device
- Honors the user's week-start and 12h/24h time-format settings

## Example

Screenshot 2026-06-23 at 08-29-20 Planner Vikunja

## Notes

This is a first pass and I'm very open to feedback on the naming, the UX, and the implementation — happy to change direction if it doesn't fit where you want the calendar view to go.

Heads up: I used Claude heavily while building this.

## Related issues
- #1364 
- #2826
- #2695
- maybe #531 
- Label area/calendar-view


carlosferaraujo:
Amazing! That’s great! 

It would be useful if we could view the calendar not only by selected date ranges but also by the week, the coming days and the following week.

Thank you for all your hard work 💚🌠

Marlon154:
Glad to hear that! 😃 

Do you mean to change the number of days to show?
I guess the problem with two weeks to show is that it will be to narrow for a good UX.

The coming days is already supported / it shows the rolling week.

carlosferaraujo:
Would it make sense to you to view the coming days and the tasks for each of them? Would that be useful? 
📆 ✔️ 🤩

Marlon154:
Sorry, I do not understand what you want.

carlosferaraujo:
> Sorry, I do not understand what you want.

We're thinking the same thing: a look at the next few days – it's going to be awesome! ✔️🎉

Marlon154:
### Features added by latest commit
- Create tasks by gesture — double-click / long-press an empty slot or drag to paint a range (timed grid + all-day row); opens a prefilled quick-add modal and drops the task in without leaving the planner.
- Configurable rolling day count (1–31) when "full week" is off; arrows page by the window, horizontal swipe slides a day.
- Sortable unscheduled sidebar (priority/title/created/%/random + default server order); sort is optional.
- Persist view mode and zoom across reloads.

### Fixes
- Date-range filter dropped tasks on the last visible day (date-only upper bound parsed as start-of-day).
- Recurring tasks (timed and all-day) now project into navigated weeks instead of only showing on their stored date.
- App header kept the previous project's title when entering the planner.
- Sidebar sort crash from a stale localStorage value; filter modelValue Boolean warning.

### Display polish
- Readable text on colored blocks/chips via getTextColor; project name, priority tag, and % progress on cards; description excerpt tooltip on blocks.

Expanded the planner E2E spec (gesture create, day count, recurrence projection, sort). Typecheck/lint/styles pass.

Screenshot 2026-06-25 at 22-52-11 Planner Vikunja

Marlon154:
### Bug fixes
- **Deleting a task from the detail popup no longer dumps you in the project list.** When the detail view is opened as an overlay (planner / kanban / gantt / related tasks), delete now defers to the modal's own close handler, returning you to where you came from. Full-page deletes still go to the project. *(`TaskDetailView.vue`)*
- **Deleted tasks now disappear from the planner.** The store emits a `lastDeletedTask` signal and the planner drops the task from the grid/sidebar (it stays mounted under the modal, so it previously showed a stale block). *(`stores/tasks.ts`, `usePlannerTasks.ts`)*
- **Recurring tasks that started long ago now project correctly.** Projection fast-forwards to the visible window instead of stepping from the stored start, so daily repeaters older than the iteration cap no longer silently vanish. *(`expandOccurrences.ts` + unit test)*
- **Creating multiple tasks at once no longer drops all but the first** — each emitted task is scheduled into the painted slot before the modal closes. *(`PlannerView.vue`)*

### Robustness / correctness
- Loading spinner + error banner on fetch failure (was a silent blank grid).
- Latest-wins guard on grid/sidebar loads to avoid stale data when navigating quickly.
- Drag `pointermove`/`pointerup` listeners are torn down on unmount (no leaks mid-drag).
- Slot size / default duration clamped to ≥5 min (a typed `0` produced `NaN` positions).

### Cleanup
- Extracted the duplicated task-colour logic into one `plannerTaskColor` helper.
- All-day items resolved once per render instead of re-filtering inside the template loop.
- Removed a dead sort branch.

### Testing
- Planner unit tests pass; `typecheck`/`lint` clean for touched files.
- Full planner Playwright spec: **9 passed**.

kolaente:
Hey thanks for this! This is quite a big PR, I'll have to set aside some time to do a proper review. I'm pretty busy right now so this might take a moment until I get to you.

Marlon154:
Thanks for taking a look! Pushed a follow-up commit with fixes and a couple of small additions since your comment.

TLDR: I did a bit more testing and found and fixed a few bugs. Also I added an overdue feature.

### New: overdue tasks in the sidebar
- Opt-in setting adds a red, count-badged "Overdue" section listing all not-done tasks anchored before today, oldest first
- Draggable into the grid like any unscheduled task — rescheduling one also defers its missed deadline
- The grid itself is untouched by the option; day headers are now all-day drop targets as well

### Bug fixes
- Several drag/resize/click interaction issues (unschedule no-op, wrong duration committed on drag/resize, accidental reschedule on multi-day block clicks)
- Recurrence projection rewritten to match backend semantics (month-overflow, fixed-interval stepping, from-current-date repeats)
- Stale project header/background when opening a task over the planner

### Polish
- Locale-aware dates/times throughout, midnight-surviving now-line -> tasks work over two days
- Filter null-handling and start-only/end-only task visibility fixed

### Testing
- Fixed drop-onto-all-day-row gesture (was hit-testing the wrong edge)
- Day headers now act as all-day drop targets
- New opt-in "Show overdue tasks in the sidebar" section
- Unit tests, lint, and typecheck: clean


Screenshot overdue column:
Screenshot 2026-07-03 at 11-27-43 Planner Vikunja

Marlon154:
TL;DR: Added a feature to quick-access existing saved filters from the filter popup. Found a bug while testing it, fixed that too.

### Saved filters in the filter popup

Added a **"Or use a saved filter:"** section to the filter popup: it lists the user's saved filters as buttons, and clicking one loads that filter's query into the input so it can be applied (and further tweaked) in the current view. Enabled via a new `showSavedFilters` prop on `Filters.vue`, passed through from `FilterPopup.vue`.

`frontend/src/components/project/partials/Filters.vue`, `FilterPopup.vue`, `frontend/src/i18n/lang/en.json`

### Fix: saved filter resolving to its own pseudo-project

While testing the above, selecting a saved filter whose name matches a real project (e.g. a "Birding" saved filter alongside the "Birding" project) returned an empty task list.

**Cause:** the filter popup round-trips the query `id → title → id`. A saved filter storing `project = 6` displays as `project = Birding`, but when re-encoded for the API, `findProjectByExactname` matched the *first* project titled "Birding" — the saved filter's own pseudo-project (id `-5`) — so the API received `project = -5`. No task belongs to a pseudo-project, hence "This project is currently empty."

**Fix:** `findProjectByExactname` now only matches real projects (`id > 0`), since a `project = X` filter value can never refer to a saved filter. This also hardens the four other callers that resolve a project name for a task context (`FilterInput.vue`, `ViewEditForm.vue`, `QuickActions.vue`, `tasks.ts`).

`frontend/src/stores/projects.ts`

Verified: API request went from `filter=project = -5` (empty) → `filter=project = 6` (tasks render).

Marlon154:
@kolaente did a few fixes in the last commit.
Further, I create a [branch](https://github.com/Marlon154/vikunja/tree/feat-planner-keyboard) with keyboard support. There are still some open issues for it to be great.
Looking forward to your review!

PR #2958, opened 2026-06-24
fix(auth): allow api tokens to access global v2 task list endpoint

13rac1:
I was getting 401s using the listing all tasks with the new v2 API, this seems like a reasonable solution.

*AI generated description follows*
<hr/>

The tasks.read_all special case in `CanDoAPIRoute` only covered v1 paths. Both GET `/api/v2/tasks` and GET `/api/v2/projects/:project/tasks` normalize to the same `tasks.read_all map` key, but only one `RouteDetail` survives — the project-scoped path overwrites the global one. The exact path comparison then rejects the global endpoint with 401.

Extend the special case to include the v2 paths.


kolaente (approved):
Good catch, thanks

PR #2959, opened 2026-06-24
fix(auth): preserve permission group names when creating API tokens

13rac1:
Saving a Bot API key with Time Entries permission causes a "Permission of group is invalid" error. This fix seems hacky, but is functionally correct and regression tested. Another solution would be to rename  "time-entries" → "time_entries", but that's a much larger change and will need user data migrations.

image
image

_AI Generated Description Follows_
<hr/>
`objectToSnakeCase` recursively transforms all object keys, including the permissions map. This mangles "time-entries" → "time_entries", causing the backend to reject the permission as invalid. Override `autoTransformBeforePut` and manually restore the original permissions map after the `snake_case` transform in `beforeCreate`.

### DIFF
--- a/frontend/src/services/apiToken.test.ts
+++ b/frontend/src/services/apiToken.test.ts
@@ -1,0 +1,35 @@
+import {describe, it, expect} from 'vitest'
+
+import {objectToSnakeCase} from '@/helpers/case'
+
+// Regression test: objectToSnakeCase mangles hyphenated permission group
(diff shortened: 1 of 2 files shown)


kolaente:
Implemented the fix in the api in https://github.com/go-vikunja/vikunja/pull/2970, will close this in favor of that PR.

13rac1:
Works for me! Thank you

PR #2961, opened 2026-06-24
feat(projects): make gantt chart zoom in if there are space available

BlackFuffey:
Title. Just a little change that I personally find useful to have.


kolaente (approved):
Good one, thanks!

PR #3062, opened 2026-06-30
fix(api): return 200 instead of 500 when listing attachments on a task with no attachments

13rac1:
Not much to say, fixing a 500.

_AI Generated Description Follows_
<hr/>

ReadAll used a bare return when len(attachments) == 0, which returned nil for the interface{} result. The v2 handler's type assertion then failed on nil, producing an untyped error that Huma surfaced as 500.

Return the empty slice explicitly so the assertion succeeds.


kolaente (approved):
Good catch!

PR #3070, opened 2026-07-01
fix: assign bot-created projects to the bot's owner

13rac1:
Problem: A bot account on my account created a project by API, but I cannot see the project.

Solution: Assign Project ownership to the bot owner, then share admin access with the bot.

I considered sharing the Project with the bot owner instead of assigning ownership, but this makes more sense. Either solution works for me though.

_AI Generated Below_
<hr/>
When a bot user creates a project, the bot's owner (the human who created the bot) now owns the project instead of the bot itself. The bot retains admin access via an explicit project share. This ensures the human can always see and manage projects their bots create.

### DIFF
--- a/pkg/models/project.go
+++ b/pkg/models/project.go
@@ -1046,8 +1046,17 @@ func CreateProject(s *xorm.Session, project *Project, auth web.Auth, createBackl
 	}
 
 	project.ID = 0
-	project.OwnerID = doer.ID
-	project.Owner = doer
+	if doer.IsBot() {
+		project.OwnerID = doer.BotOwnerID
+		owner, err := user.GetUserByID(s, doer.BotOwnerID)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


pkg/models/project.go
```diff
@@ -1061,6 +1070,18 @@ func CreateProject(s *xorm.Session, project *Project, auth web.Auth, createBackl
 		return
 	}
 
+	// Give the bot continued access to the project it created.
+	if doer.IsBot() {
+		_, err = s.Insert(&ProjectUser{
```
kolaente: Please use the Create function of the ProjectUser struct here

kolaente (changes requested):
Good catch, thanks

pkg/models/project.go
13rac1: Done!

kolaente (approved):

PR #3115, opened 2026-07-06
fix: use correct label for imported tasks 'project'

taiwithers:
Changes the label in the dropdown menu for column mapping CSV task imports from "Title" to "Project"


kolaente (approved):
Good catch, thanks

PR #3154, opened 2026-07-11
feat: allow subscribing another user to a project or task

mnadalin:
Subscriptions previously only supported subscribing the currently authenticated user. PUT/POST /subscriptions/{entity}/{entityID} now accepts an optional user_id in the body to subscribe someone else, mirroring how /tasks/{id}/assignees works.

An empty body still subscribes the authenticated user (unchanged default, backwards compatibility). Subscribing another user requires the caller to have write access to the entity, and the target user needs read access to it - enforced in both Subscription.CanCreate/Create (pkg/models) and the v2 Huma route (pkg/routes/api/v2/subscriptions.go).

### DIFF
--- a/pkg/models/subscription.go
+++ b/pkg/models/subscription.go
@@ -103,8 +103,10 @@ type Subscription struct {
 	// The id of the entity to subscribe to.
 	EntityID int64 `xorm:"bigint index not null" json:"entity_id" param:"entityID" readOnly:"true" doc:"The numeric id of the subscribed entity; taken from the request path."`
 
-	// The user who made this subscription
-	UserID int64 `xorm:"bigint index not null" json:"-"`
+	// The user who made this subscription. If left empty when creating a subscription, the currently
+	// authenticated user is subscribed. To subscribe a different user, both the authenticated user and
+	// the user to subscribe need read access to the entity.
+	UserID int64 `xorm:"bigint index not null" json:"user_id,omitempty" doc:"The id of the user to subscribe. If omitted, the currently authenticated user is subscribed. To subscribe someone else, both users need access to the entity."`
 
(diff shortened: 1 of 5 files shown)


kolaente:
What is the reason behind this feature? What's your use case?

mnadalin:
I have a "support" user who writes as Tasks in Vikunja e-mails received in the support@ mailbox, in the relevant project (defined mainly by sender). He is a regular user, but treated as a service user. The task is assigned to the user responsible for the project, but the user who wrote the email is only subscribed to the task itself.
The best solution would be to use the API keys of the user writing the email, but this becomes very cumbersome to manage. Hence the service user.

kolaente:
Why not assign the user who reported additionally? That would auto-subscribe them.

And it seems like you want to use a bot user for this: https://vikunja.io/docs/bot-users/

pkg/models/subscription.go
```diff
@@ -148,9 +153,27 @@ func (sb *Subscription) Create(s *xorm.Session, auth web.Auth) (err error) {
...
+	// subscriber defaults to the currently authenticated user; CanCreate already made sure the caller
+	// is allowed to subscribe someone else if sb.UserID is set to a different user.
+	subscriber := auth
+	if sb.UserID != 0 && sb.UserID != auth.GetID() {
+		subscriber, err = user.GetUserByID(s, sb.UserID)
+		if err != nil {
+			return err
+		}
+
+		canRead, readErr := canReadSubscriptionEntity(s, sb.EntityType, sb.EntityID, subscriber)
```
kolaente, line 166: this should be done in the Subscription's CanCreate

kolaente (changes requested):

PR #3155, opened 2026-07-11
fix(caldav): compute VTODO DURATION components correctly

TowyTowy:
`formatDuration` derived its minute/second components with `duration.Minutes() - duration.Hours()*60` and `duration.Seconds() - duration.Minutes()*60`. Because `Minutes()` is exactly `Hours()*60` (and likewise for seconds), both were always 0, and `FormatFloat(..., 'f', 0, ...)` additionally rounded the total hours. So the DURATION emitted for a VTODO was wrong for any non-whole-hour value:

- 1h30m -> PT2H0M0S  (should be PT1H30M0S)
- 30m   -> PT0H0M0S  (dropped entirely; should be PT0H30M0S)
- 45m   -> PT1H0M0S  (should be PT0H45M0S)

Switched to integer arithmetic to derive the hour/minute/second remainders, and added a `Test_formatDuration` table test covering these cases (it fails on current `main` and passes with this change). `go test ./pkg/caldav/`, `go vet`, and `gofmt` are clean.

AI disclosure: written with AI assistance (Claude). I reviewed every line, reproduced the bug, and verified the fix with `go test ./pkg/caldav/`.

### DIFF
--- a/pkg/caldav/caldav.go
+++ b/pkg/caldav/caldav.go
@@ -119,12 +119,14 @@ COLOR:` + color
 }
 
 func formatDuration(duration time.Duration) string {
-	seconds := duration.Seconds() - duration.Minutes()*60
-	minutes := duration.Minutes() - duration.Hours()*60
+	seconds := int64(duration.Seconds())
+	hours := seconds / 3600
+	minutes := (seconds % 3600) / 60
+	seconds %= 60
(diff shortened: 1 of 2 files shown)


kolaente (approved):
Good catch, thanks

PR #3158, opened 2026-07-11
feat(projects): include child tasks in parent projects

Pegoku:
## Summary

- add a project-level setting to show tasks from only the current project, all child projects, or selected child projects
- apply the configured scope consistently across project views
- include nested descendants recursively while ignoring selected projects outside the hierarchy
- expose the setting in the project edit panel next to the existing project metadata

### Project task scope setting

image

### Parent project with descendant tasks

image


kolaente:
Hey thanks for the PR!

There is also https://github.com/go-vikunja/vikunja/pull/2939 which solves the same problem and is further along the implementation, so I'll close this PR in favor of https://github.com/go-vikunja/vikunja/pull/2939.

PR #3164, opened 2026-07-12
feat(filters): filter tasks by their relations (closes #2183)

### LINKED ISSUE #2183: Filter option for task relations - filtering out blocking tasks
### Pre-submission checklist

### Description

When creating a filter for something like a View, you can filter out certain tasks but there is no filtering option/property for filtering task relations like is it 'blocked by' another task. And currently the relations feel more like a side note than actually a good visual indicator. Especially when you can use filter for that exact thing.

Couldn't find any.

grantemsley:
Adds two filter fields so tasks can be filtered by their relations. Closes #2183.

```
done = false && open_relations != blocked   # open tasks, minus anything blocked by an unfinished task
```

`relations` matches a task with at least one relation of the given kind (`subtask`, `blocked`, `related`, ...). `open_relations` only counts a relation when the task on the other end isn't done, so a task comes back as soon as its blocker is completed. Operators: `=`, `!=`, `in`, `not in`. Works anywhere `filter=` does.

Additive — no new routes, no `/api/v1` changes. Implemented as an `EXISTS` subquery on the existing sub-table filter mechanism in `pkg/models/task_search.go`, plus the frontend field list and filter help.

Tests added in `pkg/models` and `pkg/webtests`; `mage test:web` and the filter tests pass. Also tested by hand through the API and UI (every kind, cross-project, toggling a blocker done/undone).

AI-assisted: written with Claude, reviewed and tested by me. I'm not a Go dev, so the join in `task_search.go` is worth a close look.

<details>
<summary>Details</summary>

**Valid kinds:** `subtask`, `parenttask`, `blocked`, `blocking`, `related`, `duplicateof`, `duplicates`, `precedes`, `follows`, `copiedfrom`, `copiedto`. The web UI uses `openRelations` and converts to snake_case like other multi-word fields.

**Where the code is:**
- `task_collection_sort.go` / `task_collection.go` — property constants, valid-field set.
- `task_collection_filter.go` — parses values as `RelationKind`s, restricts to equality operators.
- `task_search.go` — `EXISTS` against `task_relations` with an inner join to the related task, so soft-deleted relations are ignored and `open_relations` can check the done state. `SubTableFilter` gained optional join fields, which also replace the hard-coded users join for assignees.
- Frontend: `filters.ts`, `FilterInputDocs.vue`, `en.json`. `pkg/swagger/` untouched (CI regenerates it).

**Two pre-existing quirks you'll hit while reviewing:**
- In a **List** view, `relations = subtask` looks doubled — the view always requests `expand=subtasks`. Table/Kanban/Gantt show the exact result.
- `subtask`/`parenttask` are keyed from the holder's side (a task that *has* subtasks carries `subtask`), opposite to `blocked`/`blocking`. Matches the existing `related_tasks` map; happy to rename.
</details>

### DIFF
--- a/frontend/src/components/input/filter/FilterInputDocs.vue
+++ b/frontend/src/components/input/filter/FilterInputDocs.vue
@@ -33,6 +33,8 @@ const showDocs = ref(false)
 			<li><code>reminders</code>: {{ $t('filters.query.help.fields.reminders') }}</li>
 			<li><code>created</code>: {{ $t('filters.query.help.fields.created') }}</li>
 			<li><code>updated</code>: {{ $t('filters.query.help.fields.updated') }}</li>
+			<li><code>relations</code>: {{ $t('filters.query.help.fields.relations') }}</li>
+			<li><code>openRelations</code>: {{ $t('filters.query.help.fields.openRelations') }}</li>
 		</ul>
 		<p>{{ $t('filters.query.help.canUseDatemath') }}</p>
 		<p>{{ $t('filters.query.help.operators.intro') }}</p>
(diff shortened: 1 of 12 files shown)


grantemsley:
Rebased onto current main and re-tested. Trimmed the description too — it was much longer than it needed to be.

Happy to change anything if it makes it easier for you to review and include.

grantemsley:
Fixed merge conflicts again.

The conflict was with #2916 (filter by creator username), which needed the same users join this PR was already generalizing: the new JoinTable/JoinCond fields now carry it, replacing the hardcoded if sf.Table == "task_assignees" special case.

PR #3180, opened 2026-07-14
feat(tasks): show the task identifier in the list view 🤖🤖🤖

WilliamLowther:
Renders each task's identifier before the title in the List view, reusing the existing `getTaskIdentifier()` helper so the format matches the task detail heading (and honours a project identifier when one is set).

Scope: only `SingleTaskInProject.vue` — one template line, one import, and a small `.task-id` style reusing the existing `--grey-400` token. No new dependencies or logic changes.

`pnpm lint` and `pnpm lint:styles` pass.


kolaente:
Sorry, but this is pretty intrusive for most users, we won't merge this.

PR #3183, opened 2026-07-15
fix(caldav): parse ISO 8601 week durations in reminder triggers

TowyTowy:
`ParseISO8601Duration` ignored the week designator: `P1W` (and `-P2W`) matched only the bare `P` and returned `0`. Per RFC 5545, `dur-week` is a valid VALARM `TRIGGER` (e.g. `TRIGGER:-P1W` for one week before), so importing such an alarm silently produced a zero-offset reminder. The same parser also backs the TickTick migration importer, which was affected too.

Fix: add a week group to the duration regex and fold it into the total. Added regression tests for `P1W` / `-P2W`. Month-vs-minute disambiguation is unchanged (`P1M` = 1 month, `PT1M` = 1 minute).

Verified with `go test ./pkg/utils/ ./pkg/caldav/` (the new week cases fail before / pass after); gofmt/vet clean.

_Prepared with Claude Code assistance (see `Co-Authored-By` trailer)._

### DIFF
--- a/pkg/utils/duration.go
+++ b/pkg/utils/duration.go
@@ -32,12 +32,13 @@ func ParseISO8601Duration(str string) time.Duration {
 
 	years := parseDurationPart(matches[2], time.Hour*24*365)
 	months := parseDurationPart(matches[3], time.Hour*24*30)
-	days := parseDurationPart(matches[4], time.Hour*24)
-	hours := parseDurationPart(matches[5], time.Hour)
-	minutes := parseDurationPart(matches[6], time.Second*60)
-	seconds := parseDurationPart(matches[7], time.Second)
+	weeks := parseDurationPart(matches[4], time.Hour*24*7)
+	days := parseDurationPart(matches[5], time.Hour*24)
+	hours := parseDurationPart(matches[6], time.Hour)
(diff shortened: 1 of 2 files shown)


kolaente (changes requested):
Can you add a test that checks a whole string with the `W` and other time things? Something like `P1W2D`

TowyTowy:
Thanks for the review! Added a combined-units regression test as suggested: `P1W2D` (→ `216h`) and `P1W2DT3H4M5S` (→ `219h4m5s`), so the week designator is exercised alongside days and the `T` time components in one string. Both cases fail against the pre-fix parser (they resolved to `0` because the ungrouped `W` caused the whole match after `P` to be dropped, not just the week) and pass with the regex change. `gofmt`/`go vet` are clean and `go test ./pkg/utils/ ./pkg/caldav/` is green.

(For reference, the failing `docker` check is the `pull_request_target` fork-checkout policy, which fails for any fork PR and is unrelated to this diff — all the `Test /` jobs pass.)

_Prepared with Claude Code assistance (also noted via the `Co-Authored-By` trailer)._

kolaente (approved):
Thanks!

PR #3190, opened 2026-07-15
fix(notifications): strip Markdown escapes from plain-text mail part 🤖🤖🤖

### LINKED ISSUE #3179: 2.3.0 regression: escaped characters in text part of mail body
### Pre-submission checklist

### Description

I'm running v2.3.0 and configured mail reminders for certain tasks. The mail body contains the text `This is a friendly reminder of the task "TITLE" (PROJECT).`, with TITLE and PROJECT replaced accordingly. For a task with title `Test - xyz - 123|456@789!` I get `Test \- xyz \- 123\|456@789\!`, though.

With previous versions, the backslashes did not appear.

This seems to be limited to the text part of mails (I'm using a text-only client).

### Vikunja Version

v2.3.0

### Browser and version

-

No

### Screenshots

_No response_

maximilize:
## What

Fixes #3179. Since v2.3.0 the plain-text part of notification e-mails shows CommonMark backslash escapes: a task titled `Test - xyz - 123|456@789!` renders as `Test \- xyz \- 123\|456@789\!` in text-only clients.

## Root cause

`EscapeMarkdown` (added for GHSA-45q4-x4r9-8fqj to neutralize Markdown injection in the HTML part) is applied to user-controlled values at line-construction time in `pkg/models/notifications.go`, so the escapes become part of the canonical `line.Text`. The HTML part needs them — the text is fed through goldmark — but the plain-text template renders `{{ .Text }}` verbatim, so the backslashes end up in the `text/plain` MIME part. Every notification mail (reminder, assigned, overdue, …) is affected.

## Fix

The plain-text template data now always goes through `convertLinesToPlain`, which drops the escapes via a new `UnescapeMarkdown` — the precise inverse of `EscapeMarkdown`. It removes a backslash only in front of a character `EscapeMarkdown` would have escaped, so unrelated text stays untouched. The HTML part is unchanged and stays escaped.

## Tests

- `TestUnescapeMarkdown` and `TestUnescapeMarkdown_InvertsEscape` cover the round-trip identity `Unescape(Escape(x)) == x`.
- `TestRenderMail/markdown escapes are stripped from the plain text part` reproduces #3179: it fails on the current code and passes with the fix. The full `pkg/notifications` suite stays green.

### DIFF
--- a/pkg/notifications/mail_render.go
+++ b/pkg/notifications/mail_render.go
@@ -295,7 +295,9 @@ func convertLinesToPlain(lines []*mailLine) []*mailLine {
 	plain := make([]*mailLine, 0, len(lines))
 	for _, line := range lines {
 		if !line.isHTML {
-			plain = append(plain, line)
+			// Drop the CommonMark escapes added for the HTML part (see EscapeMarkdown);
+			// in plain text they would otherwise show up verbatim, e.g. `Test \- 1\|2`.
+			plain = append(plain, &mailLine{Text: UnescapeMarkdown(line.Text)})
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


kolaente:
Hey thanks for this, but this is not the right fix. I'll close this one in favor of https://github.com/go-vikunja/vikunja/pull/3219 which takes a different approach.

PR #3205, opened 2026-07-17
chore(deps): replace vendored humaecho5 with upstream humaecho adapter

kolaente:
## Summary

Replaces the vendored `humaecho5` adapter with the upstream `humaecho` adapter from Huma v2.39.0, which now supports Echo v5. The Vikunja-specific glue (stashing the echo context on request context and handling group-prefix path rewrites) is preserved in a new `humabridge` module that wraps the upstream adapter.

## Key Changes

- **Removed**: `pkg/modules/humaecho5/` — the vendored Echo v5 adapter implementation
- **Added**: `pkg/modules/humabridge/` — thin wrapper around upstream `humaecho` that:
  - Stashes `*echo.Context` on the request context via middleware so handlers can retrieve it without per-handler wiring
  - Wraps the adapter to rewrite internal Huma dispatches (e.g., autopatch's GET+PUT round trip) with the group prefix, ensuring relative paths resolve correctly when mounted on a sub-group
- **Updated**: All imports across the codebase to use `humabridge` instead of `humaecho5`
- **Updated**: `go.mod` to Huma v2.39.0 (which includes the upstream `humaecho` adapter for Echo v5)

## Implementation Details

The `humabridge` module provides:
- `EchoContextKey` — a context key for retrieving the underlying `*echo.Context` from a Huma handler's `context.Context`
- `NewWithGroup()` — mounts a Huma API on an Echo group with middleware that stashes the echo context and a custom adapter that rewrites internal dispatch paths with the group prefix
- `groupPrefixAdapter` — wraps the upstream adapter to prepend the group prefix to paths that don't already have it, enabling autopatch and other internal Huma operations to work correctly under a sub-group

Tests in `humabridge_test.go` verify:
- Round-trip request handling and echo context availability
- OpenAPI spec serving under the group prefix
- Autopatch PATCH operation with group-relative path resolution

https://claude.ai/code/session_01MRLcn9aa27AYDLe5rar5ya


kolaente:
auggie review

PR #2363, opened 2026-03-04
feat: allow sorting tasks within Kanban buckets by priority, due date, and other fields

justinan7:
## Summary

Kanban views currently only support manual position-based task ordering (drag-and-drop). Users want to sort tasks within buckets by priority, due date, etc. — see #2085, #2244, and [community discussion](https://community.vikunja.io/t/kanban-sort-done-bucket-by-most-recently-completed-tasks/2525).

This PR adds `bucket_sort_by` and `bucket_sort_order` fields to `ProjectView`, allowing users to configure how tasks are sorted within **all Kanban buckets** in a view.

- **Supported sort fields:** `position` (default), `priority`, `due_date`, `created`, `updated`, `title`
- **Sort directions:** `asc` (default), `desc`
- Default remains `position` (manual drag order) for **full backward compatibility**

Closes #2085

### Backend (Go)
- **`pkg/models/project_view.go`** — Added `BucketSortBy` and `BucketSortOrder` fields to the `ProjectView` struct; added both to the `Update()` method's `Cols()` whitelist so the API can persist them
- **`pkg/models/kanban.go`** — `GetTasksInBucketsForView()` now reads the view's sort settings and applies them instead of always using position-based ordering. Reuses the existing `sortParam`/`sortOrder` system from list/table views
- **`pkg/migration/20260304200000.go`** — Database migration to add both columns to `project_views` table (with `columnExists` checks for idempotency)

### Frontend (Vue 3 + TypeScript)
- **`frontend/src/modelTypes/IProjectView.ts`** — Added `bucketSortBy` and `bucketSortOrder` to the interface
- **`frontend/src/models/projectView.ts`** — Added default values for the new fields
- **`frontend/src/components/project/views/ProjectKanban.vue`** — Added a sort dropdown (using the existing `Dropdown`/`DropdownItem` components) in the Kanban header toolbar, next to the existing filter popup. Uses the FontAwesome `sort` icon as the trigger. Changing the sort persists the setting via `ProjectViewService.update()` and reloads buckets
- **`frontend/src/i18n/lang/en.json`** — Added translation strings for sort options

### Default Kanban view (manual/position order)
Default kanban view

### Sort dropdown open — all available sort options
Sort dropdown open

### Sorted by Priority (ascending) — Medium → Urgent → DO NOW
Priority ascending

### Sorted by Priority (descending) — DO NOW → Urgent → Medium
Priority descending

## Test plan

- [x] `go build ./...` passes
- [x] `mage build:build` passes
[...]

Issue #2085, opened 2026-01-11
Add sorting to Kanban buckets


DonateIfYouCan:
### Description

I'd like to be able to sort for eg. my todo kanban bucket by priority or due date. Or have a default sorting enabled or disabled. 

### Which alternatives did you consider using instead?

I use a list and filters, but that requires more work to setup.

kolaente:
Is this a duplicate of https://github.com/go-vikunja/vikunja/issues/1729?

DonateIfYouCan:
No, it's not duplicate. That one is about the list view. But there is a comment on the Kanban view in that issue as well.

I created this issue, to add emphasis to the kanban board sorting and that manually sorting should still be an option.

PR #3220, opened 2026-07-18
feat: allow sorting tasks within Kanban buckets by priority, due date, and other fields (new PR)

Rapha149:
This is my try at #2363 (adding sort options for tasks in Kanban buckets, see also #2085); since the original author does not seem active anymore and there were requested changes. So I implemented them:

- Removed the screenshots
- Replaced `Ascending` and `Descending` with appropriate names for each sort option (e.g. `Earliest due date first`)
- Moved the sort settings to the view settings

I also added a change that I find useful: it is now possible to select multiple sort options to apply them in order. For example, you can choose to sort by priority and within a priority sort by due date.

grafik

The code was generated by Claude Code.


pkg/models/project_view.go
```diff
@@ -173,6 +173,11 @@ type ProjectView struct {
 	// If tasks are moved to the done bucket, they are marked as done. If they are marked as done individually, they are moved into the done bucket.
 	DoneBucketID int64 `xorm:"bigint INDEX null" json:"done_bucket_id" doc:"The id of the done bucket. Tasks moved here are marked done, and tasks marked done are moved here."`
 
+	// The fields to sort tasks by within kanban buckets, applied in order. Valid values are `position`, `priority`, `due_date`, `created`, `updated`, `title`. Defaults to `position` (manual drag order).
+	BucketSortBy []string `xorm:"json null default null" json:"bucket_sort_by" doc:"The fields to sort tasks by within kanban buckets, applied in order. Valid values are position, priority, due_date, created, updated, title."`
```
kolaente: This should be done in the bucket, not in the project view.

frontend/src/i18n/lang/en.json
```diff
@@ -522,7 +522,37 @@
       "bucketLimitSavedSuccess": "The bucket limit been saved successfully.",
       "collapse": "Collapse this bucket",
       "bucketLimitReached": "You have reached the bucket limit. Remove tasks or increase the limit to add new tasks.",
-      "bucketOptions": "Bucket options"
+      "bucketOptions": "Bucket options",
+      "sortBy": "Sort tasks within buckets by",
+      "sortEmptyHint": "Tasks are ordered manually via drag & drop.",
+      "addSort": "Add sort",
+      "sortPriority": "Priority",
```
kolaente: There are sorting options for the list view already, please use the translation strings from there

kolaente (changes requested):
Hey thanks for this! Generally yes, but we should move the sorting setting from the project view to the bucket, allowing for bigger flexibility. This affects a bunch of things, and I'm not entirely sure how to make that work in the UI.

Rapha149:
I moved the sort options to the bucket settings as requested, they are accessible through the three-dot-menu of each bucket. For convenience, I also added a button to apply the sort options of the current bucket to all buckets in that Kanban view.

I now also use the translations from the existing list view sort options and also added three more sort options presen there.
While testing, I noticed that the German translations for start/end date sorting were wrong, so I fixed those as well.

Here's how it looks now:
grafik
grafik
grafik

Rapha149:
Since the first review came in quickly, just following up to make sure this hasn't slipped through the cracks.

I also fixed the merge conflicts that arose with a force push a few days ago, but the tests don't seem to run for some reason (except for the docker test whose failure seems intentional).

PR #3223, opened 2026-07-18
docs(swagger): fix @Success 200 -> 201 on 5 PUT/Create endpoints

netadvanced:
## What / why

Five `PUT`-routed "create" endpoints document `@Success 200`, but they all dispatch through the shared `CreateWeb` handler (`pkg/web/handler/create.go`), which always responds `201 Created` regardless of the model's own swagger comment. Every other `CreateWeb`-routed endpoint (labels, teams, filters, task relations/comments/assignees/duplicate, task-labels) already correctly documents `201` — these five are stragglers with the same stale annotation, confirmed by checking the generated `pkg/swagger/swagger.json` itself, not just the source comments.

Found while verifying API behavior against the source for an unrelated investigation.

## Endpoints fixed

- `PUT /tokens` (`pkg/models/api_tokens.go`)
- `PUT /projects/{id}/views/{view}/buckets` (`pkg/models/kanban.go`)
- `PUT /projects/{id}/webhooks` (`pkg/models/webhooks.go`)
- `PUT /{kind}/{id}/reactions` (`pkg/models/reaction.go`)
- `PUT /projects/{project}/views` (`pkg/models/project_view.go`)

## Scope

Docs only — no behavior change. The 5 annotation edits plus a regenerated `pkg/swagger/{docs.go,swagger.json,swagger.yaml}` via:

```
swag init -g ./pkg/routes/routes.go --parseDependency -d . -o ./pkg/swagger
```

(the same command `mage generate:swagger-docs` runs), so `mage check:gotswag` should pass and the generated files match the annotations rather than needing CI to regenerate them.

### DIFF
--- a/pkg/models/api_tokens.go
+++ b/pkg/models/api_tokens.go
@@ -84,7 +84,7 @@ func GetAPITokenByID(s *xorm.Session, id int64) (token *APIToken, err error) {
 // @Produce json
 // @Security JWTKeyAuth
 // @Param token body models.APIToken true "The token object with required fields"
-// @Success 200 {object} models.APIToken "The created token."
+// @Success 201 {object} models.APIToken "The created token."
 // @Failure 400 {object} web.HTTPError "Invalid token object provided."
 // @Failure 500 {object} models.Message "Internal error"
 // @Router /tokens [put]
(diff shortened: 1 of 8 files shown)


kolaente (approved):
Good catch, thanks!

PR #3229, opened 2026-07-19
fix(filters): resolve relative and absolute date filters in UTC (#3181) 🤖🤖🤖

### LINKED ISSUE #3181: Relative saved-filter day boundary ignores filter timezone with PostgreSQL
### Pre-submission checklist

- [x] I searched existing open and closed issues for the same problem.

### Description

Relative day boundaries in a saved task filter appear to be evaluated or serialized as UTC midnight even though the request explicitly supplies `filter_timezone=America/Los_Angeles` and both the service and user are configured for that timezone.

At approximately 2026-07-14 17:00 PDT, both of these saved-filter expressions excluded active tasks due later on July 14 local time:

```text
done = false && due_date < now/d+1d
done = false && due_date < now+1d/d
```

Adding the current PDT-to-UTC offset made the expected tasks appear:

```text
done = false && due_date < now/d+1d+7h
```

That workaround is DST-sensitive and would need `+8h` during Pacific Standard Time.

#### Configuration and runtime evidence

The running container has:

```text
VIKUNJA_SERVICE_TIMEZONE=America/Los_Angeles
VIKUNJA_DEFAULTSETTINGS_TIMEZONE=America/Los_Angeles
```

The authenticated user's `users.timezone` value is:

```text
America/Los_Angeles
```

The actual GET request logged by Vikunja contains:

```text
filter_timezone=America%2FLos_Angeles
```

The NAS effective local clock is `America/Los_Angeles`. PostgreSQL 16 has its normal session timezone of UTC and converts `current_timestamp` to Los Angeles time correctly.

Representative due timestamps are stored in PostgreSQL as UTC values such as:

```text
2026-07-15 01:00:00
2026-07-15 04:24:59
```

These correspond to the evening of July 14 PDT and should match a filter whose upper bound is the start of July 15 in `America/Los_Angeles` (2026-07-15 07:00:00 UTC). The behavior of the unadjusted filter is consistent with a 2026-07-15 00:00:00 UTC cutoff instead. Adding `+7h` makes the expected tasks appear.

The saved-filter JSON does not persist `filter_timezone`, but the v2.3.0 backend falls back to the authenticated user's timezone for saved filters, and the request also explicitly supplies the correct timezone.

#### Expected behavior

With `filter_timezone=America/Los_Angeles`, `now/d+1d` and `now+1d/d` should resolve to the next local midnight and compare against the UTC-backed task timestamps without requiring a manual UTC offset.

#### Actual behavior

Tasks due during the last seven hours of the local calendar day are excluded. Adding `+7h` to the expression during PDT makes them appear.

#### Environment

- Vikunja: v2.3.0 (`vikunja/vikunja` combined container)
- Database: PostgreSQL 16
- Browser: Google Chrome 150.0.7871.114
- Deployment: Docker Compose behind Caddy
- Host effective timezone: America/Los_Angeles
- Service timezone: America/Los_Angeles
- User timezone: America/Los_Angeles
- PostgreSQL session timezone: UTC

### Vikunja Version

v2.3.0

### Browser and version

Google Chrome 150.0.7871.114

Not tested. The reproduction uses controlled UTC due timestamps and server/request/database configuration that are available on the self-hosted instance.

### Screenshots

None. Request-log, configuration, and database timestamp evidence is included above.

maximilize:
## Description

Relative saved-filter day boundaries like `due_date < now/d+1d` ignored the filter timezone and effectively used UTC midnight, so tasks due in the last hours of the local day were dropped (#3181).

### Root cause

Task timestamps are stored in a naive `timestamp` column as UTC. A filter's date value was emitted in the **service timezone** — `getValueForField` (datemath) and `parseTimeFromUserInput` (absolute dates) both did `.In(config.GetTimeZone())`. When that value is bound as a query parameter, the database driver hands it over with its offset, but comparing against a `timestamp without time zone` column **drops the offset** and keeps the wall clock. So with a non-UTC service/filter timezone the effective cutoff was the local wall clock reinterpreted as UTC — off by the UTC offset.

Concretely, at 2026-07-14 17:00 PDT with `filter_timezone=America/Los_Angeles`, `now/d+1d` should resolve to 2026-07-15 07:00 UTC (local midnight), but was bound as `2026-07-15 00:00 -0700`, which Postgres coerces to the naive `2026-07-15 00:00:00` (`'2026-07-15 00:00:00-07'::timestamp` = `2026-07-15 00:00:00`). A task due 2026-07-15 03:00 UTC (the evening of July 14 local) is then excluded.

### Fix

Emit the resolved instant in UTC. The filter timezone still controls how the datemath rounds (`now/d` is the start of the day in the filter's timezone), so only the representation handed to the database changes. Applies to both the relative (datemath) and absolute date paths, since they share the same root cause.

### Verification

- Reproduced end-to-end against PostgreSQL 16 with the production engine config (`SetTZLocation(service)` + `SetTZDatabase(GMT)`, `timestamp` column): with the old code a task due 2026-07-15 03:00 UTC is excluded by `due_date < now/d+1d` (filter timezone `America/Los_Angeles`); with the fix it is included.
- New regression test `TestDateFilterTimezone` (relative and absolute) — red before, green after.
- Full `pkg/models` suite, `gofmt`, `go vet` and `golangci-lint` all clean.

`config.SetTimeZone` is added so the regression test can exercise a non-UTC service timezone (the parsed zone is cached on first use).

## User-facing changes (Release notes)

Fixed an issue where relative and absolute date filters (e.g. `due_date < now/d+1d`) ignored the configured timezone on instances whose service timezone is not UTC, which could hide tasks due in the last hours of the local day.

## Additional notes

Fixes #3181

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -297,9 +297,8 @@ func GetTimeZone() *time.Location {
 	return timezone
 }
 
-// SetTimeZone overrides the cached time zone. It is primarily useful in tests
-// that need to exercise a non-default service timezone, since GetTimeZone caches
-// the parsed location on first use.
+// SetTimeZone overrides the zone cached by GetTimeZone, for tests that need a
+// non-default service timezone.
 func SetTimeZone(loc *time.Location) {
 	timezone = loc
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -287,20 +287,15 @@ var timezone *time.Location
 // It is a separate function and not done through viper because that makes handling
 // it way easier, especially when testing.
 func GetTimeZone() *time.Location {
-	if timezone == nil {
-		loc, err := time.LoadLocation(ServiceTimeZone.GetString())
+	tz := ServiceTimeZone.GetString()
+	if timezone == nil || timezone.String() != tz {
+		loc, err := time.LoadLocation(tz)
 		if err != nil {
 			log.Fatalf("Error parsing time zone: %s", err)
 		}
 		timezone = loc
 	}
 	return timezone
-}
-
-// SetTimeZone overrides the zone cached by GetTimeZone, for tests that need a
-// non-default service timezone.
-func SetTimeZone(loc *time.Location) {
-	timezone = loc
 }
 
 // Set sets a value
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -297,6 +297,13 @@ func GetTimeZone() *time.Location {
 	return timezone
 }
 
+// SetTimeZone overrides the cached time zone. It is primarily useful in tests
+// that need to exercise a non-default service timezone, since GetTimeZone caches
+// the parsed location on first use.
+func SetTimeZone(loc *time.Location) {
+	timezone = loc
+}
+
 // Set sets a value
 func (k Key) Set(i interface{}) {
 	viper.Set(string(k), i)
--- a/pkg/models/task_collection_filter.go
+++ b/pkg/models/task_collection_filter.go
@@ -108,9 +108,7 @@ func parseTimeFromUserInput(timeString string, loc *time.Location) (value time.T
 		}
 		value = time.Date(year, time.Month(month), day, 0, 0, 0, 0, loc)
 	}
-	// Emit the instant in UTC (see the note in getValueForField): task timestamps
-	// live in a naive UTC column and the driver drops the parameter's offset, so a
-	// service-timezone wall clock would shift the comparison by the UTC offset.
+	// UTC, not service timezone — see getValueForField.
 	value = value.UTC()
 	value = adjustDateForMysql(value)
 	return value, err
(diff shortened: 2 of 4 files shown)


pkg/config/config.go
```diff
@@ -297,6 +297,13 @@ func GetTimeZone() *time.Location {
 	return timezone
 }
 
+// SetTimeZone overrides the cached time zone. It is primarily useful in tests
+// that need to exercise a non-default service timezone, since GetTimeZone caches
+// the parsed location on first use.
+func SetTimeZone(loc *time.Location) {
```
kolaente: if you only want to use this for tests, use `config.ServiceTimeZone.Set`

pkg/models/task_collection_filter.go
```diff
@@ -108,7 +108,10 @@ func parseTimeFromUserInput(timeString string, loc *time.Location) (value time.T
 		}
 		value = time.Date(year, time.Month(month), day, 0, 0, 0, 0, loc)
 	}
-	value = value.In(config.GetTimeZone())
+	// Emit the instant in UTC (see the note in getValueForField): task timestamps
+	// live in a naive UTC column and the driver drops the parameter's offset, so a
+	// service-timezone wall clock would shift the comparison by the UTC offset.
```
kolaente: Please tell your agent to emit shorter comments. Like wayyy shorter. If I want to read prose, I pull up a book, not agent comments. No-One reads that.

kolaente (changes requested):

PR #3240, opened 2026-07-19
fix: 5 CodeQL-flagged security findings (path validation, clear-text logging, allocation bounds)

kigiela:
## Summary

We run CodeQL on our fork and found these worth sending upstream - each is a small, independent fix (see individual commits):

1. **`pkg/routes/static.go`** - the static-file handler relied only on `path.Clean`'s implicit sandboxing before opening a file; added an explicit check that the resolved path stays within the dist root.
2. **`pkg/modules/auth/openid/openid.go`** - three `Debug`-level log calls printed the raw token-endpoint response body/full `oauth2.Token` (no `String()` override, so `%v` includes `AccessToken`/`RefreshToken`) on error paths. Now logs the RFC 6749 `error`/`error_description` fields instead.
3. **`pkg/modules/migration/csv/csv.go`** - `DetectCSVStructure`/`PreviewImport`/`MigrateWithConfig` each allocated a buffer sized directly from caller input with no upper bound. Capped to the existing configured max upload size.
4. **`pkg/modules/avatar/upload/upload.go`** - `GetAvatar` narrowed a caller-supplied `int64` to `int` with no bounds check before `imaging.Resize`. Added a bounds check.
5. **`pkg/models/project_team.go`** - CodeQL couldn't verify the cross-package `db.ILIKE()` helper as parameter-bound in `TeamProject.ReadAll`. Rewritten with the in-file `"col LIKE ?", value` idiom already used three lines below for the count query (search term is always a bound argument either way).

Two other CodeQL-flagged spots were investigated and are **not** included here since they're false positives / already safe (verified: `db.ILIKE`'s search argument is always parameter-bound in `pkg/models/teams.go` and `pkg/models/project_users.go`, and `pkg/db/dump.go`'s table name is already validated against a strict identifier allowlist before the flagged lines) - happy to share that analysis if useful.

## Verification
`go vet`/`go build` clean on every touched package (excluding `pkg/routes` itself, which needs a built frontend to embed - unrelated to this change; verified imports are all already present).

### DIFF
--- a/pkg/models/project_team.go
+++ b/pkg/models/project_team.go
@@ -24,6 +24,7 @@ import (
 	"code.vikunja.io/api/pkg/web"
 
 	"xorm.io/xorm"
+	"xorm.io/xorm/schemas"
 )
 
 // TeamProject defines the relation between a team and a project
@@ -191,13 +192,20 @@ func (tl *TeamProject) ReadAll(s *xorm.Session, a web.Auth, search string, page 
 
 	limit, start := getLimitFromPageIndex(page, perPage)
 
+	// The search term is always passed as a bound query argument below; the operator
+	// itself is chosen from a fixed, non-user-controlled set based on the db driver.
+	likeOperator := "LIKE"
+	if db.Type() == schemas.POSTGRES {
+		likeOperator = "ILIKE"
+	}
+
 	// Get the teams
 	all := []*TeamWithPermission{}
 	query := s.
 		Table("teams").
 		Join("INNER", "team_projects", "team_id = teams.id").
 		Where("team_projects.project_id = ?", tl.ProjectID).
-		Where(db.ILIKE("teams.name", search))
+		Where("teams.name "+likeOperator+" ?", "%"+search+"%")
 	if limit > 0 {
 		query = query.Limit(limit, start)
 	}
@@ -220,7 +228,7 @@ func (tl *TeamProject) ReadAll(s *xorm.Session, a web.Auth, search string, page 
 		Table("teams").
 		Join("INNER", "team_projects", "team_id = teams.id").
 		Where("team_projects.project_id = ?", tl.ProjectID).
-		Where("teams.name LIKE ?", "%"+search+"%").
+		Where("teams.name "+likeOperator+" ?", "%"+search+"%").
 		Count(&TeamWithPermission{})
 	if err != nil {
 		return nil, 0, 0, err
--- a/pkg/models/project_team.go
+++ b/pkg/models/project_team.go
@@ -24,7 +24,6 @@ import (
 	"code.vikunja.io/api/pkg/web"
 
 	"xorm.io/xorm"
-	"xorm.io/xorm/schemas"
 )
 
 // TeamProject defines the relation between a team and a project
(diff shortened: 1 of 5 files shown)


pkg/models/project_team.go
```diff
@@ -191,13 +192,20 @@ func (tl *TeamProject) ReadAll(s *xorm.Session, a web.Auth, search string, page
...
+	}
+
 	// Get the teams
 	all := []*TeamWithPermission{}
 	query := s.
 		Table("teams").
 		Join("INNER", "team_projects", "team_id = teams.id").
 		Where("team_projects.project_id = ?", tl.ProjectID).
-		Where(db.ILIKE("teams.name", search))
+		Where("teams.name "+likeOperator+" ?", "%"+search+"%")
```
kolaente: The whole point of the `db.ILIKE` helper is to avoid that dance in every function call. Do not pull it out again.

pkg/modules/auth/openid/openid.go
```diff
@@ -626,12 +602,15 @@ func exchangeOidcTokens(cb *Callback, providerKey string) (*Provider, *oauth2.To
 			details := make(map[string]interface{})
 			if err := json.Unmarshal(rerr.Body, &details); err != nil {
 				log.Errorf("Error unmarshalling token for provider %s: %v", provider.Name, err)
-				log.Debugf("Raw token value is %s", rerr.Body)
+				// Do not log rerr.Body: the raw token-endpoint response can carry
+				// sensitive request/response context. The RFC 6749 error fields are
+				// the intended human-readable diagnostic, not the raw body.
```
kolaente: True, but please remove the comment.

pkg/modules/migration/csv/csv.go
```diff
@@ -313,11 +313,22 @@ func parseCSV(data []byte, delimiter string) ([]string, [][]string, error) {
...
+	// #nosec G115 -- configured value won't exceed int64 max in practice.
+	return int64(config.GetMaxFileSizeInMBytes()) * 1024 * 1024
+}
+
 // DetectCSVStructure analyzes a CSV file and returns detection results
 func DetectCSVStructure(file io.ReaderAt, size int64) (*DetectionResult, error) {
 	if size == 0 {
 		return nil, &migration.ErrFileIsEmpty{}
 	}
+	if size < 0 || size > maxImportFileBytes() {
```
kolaente: Size can never be negative, do we need this check?

pkg/routes/static.go
```diff
@@ -147,7 +147,15 @@ func static() echo.MiddlewareFunc {
 			if err != nil {
 				return
 			}
-			name := path.Join(rootPath, path.Clean("/"+p)) // "/"+ for security
+			cleaned := path.Clean("/" + p) // "/"+ for security
```
kolaente: This is not needed. We're serving embedded assets here, not general filesystem things. There is no path to traverse since the files and folders this operates on is purely the `dist/` folder from the compiled frontend.

kolaente (changes requested):
Hey thanks. A few things are real issues, a few are false positives. Please fix as reviewed.

pkg/models/project_team.go
kigiela: Reverted to db.ILIKE. Also switched the count query to it - that one was using a plain case-sensitive LIKE, so on postgres it could disagree with the main query.

pkg/modules/auth/openid/openid.go
kigiela: Removed.

pkg/modules/migration/csv/csv.go
kigiela: Right, dropped it.

pkg/routes/static.go
kigiela: Reverted.

PR #3245, opened 2026-07-19
fix(modal): prevent scrollbar drag from closing the task detail modal 🤖🤖🤖

Alex-Blanes:
## Bug

On desktop, dragging the task-detail modal's vertical scrollbar and releasing the mouse outside the modal closes it unexpectedly. This makes it easy to lose the task view while simply scrolling through a long task.

## Cause

The scrollbar sits on `.modal-container`, whose `@mousedown.self` handler treats a press/release on the scrollbar as a backdrop click and closes the modal.

## Fix

On desktop (≥ `$tablet`) `overflow` moves to `.modal-content`, so the scrollbar lives inside the content area instead of the backdrop container. The `$tablet`/`$desktop` breakpoints are adjusted so scrolling stays within `100dvh`. Single file, SCSS only — no behaviour change on mobile.

## Repro / environment

Firefox 153.0 (desktop):

1. Open a task (task-detail modal).
2. Drag the modal's vertical scrollbar.
3. Release the mouse outside the modal.

Without the fix the modal closes; with the fix it stays open.

---

AI-assisted (Claude Code); reviewed, tested and authored by me.

### DIFF
--- a/frontend/src/components/misc/Modal.vue
+++ b/frontend/src/components/misc/Modal.vue
@@ -349,6 +349,17 @@ $modal-width: 1024px;
 
 // scrolling-content
 // used e.g. for <TaskDetailViewModal>
+.scrolling {
+	// On desktop the scrollbar lives on .modal-content so that
+	// clicking / dragging the scrollbar does not trigger
+	// @mousedown.self on .modal-container and close the modal.
+	@media screen and (min-width: $tablet) {
+		.modal-container {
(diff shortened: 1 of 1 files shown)


kolaente:
Sorry, but this makes the UI (and arguably the UX as well) worse, since there are two scrollbars now - one inside the modal and one on the right edge of the screen.

PR #3247, opened 2026-07-20
fix: 4 more security/robustness findings (DoS crash, plugin symbol exposure, silent 2FA/SSRF bypass)

kigiela:
## Summary

Continuing from #3240 - found these running our own audit on our fork, each independent (see individual commits):

1. **`pkg/modules/migration/ticktick/ticktick.go`** - `sortParentsBeforeChildren`'s cycle-following recursion only marked a task as placed *after* recursing into its parent, so a crafted CSV where two tasks list each other as parent (or one lists itself) recurses forever. This is a Go runtime stack overflow, which `panic`/`recover` (including Echo's `middleware.Recover()`) cannot catch - it kills the whole process, not just the request. Any authenticated user can trigger it via `PUT /migration/ticktick/migrate` with a 1-2 row CSV. Fixed with a tri-state visited map so a cycle is detected and stops recursion instead; unchanged for every acyclic input.
2. **`pkg/plugins/yaegi/loader.go`** / **`pkg/yaegi_symbols/vikunja_db.go`** - the generated yaegi symbol table (mechanical `yaegi extract` output, no curation) hands every plugin `db.WipeEverything`/`Dump`/`Restore`/`RestoreAndTruncate` and Go-test-only fixture helpers. No shipped plugin or yaegi test references any of these 9 symbols. Filtered them out of a copy of the symbol map before `i.Use()`, and added a startup warning documenting that plugin code runs with the server's full OS/DB privileges regardless (the real, load-bearing boundary - `db.NewSession` and yaegi's own stdlib access already grant that).
3. **`pkg/user/totp.go`** - `TOTPEnabledForUser` returned `false` from the `service.enabletotp` flag alone, without ever reading the user's own `totp.enabled` row - so an operator disabling that flag (incident response, working around a bug) silently drops the passcode requirement at login for every already-enrolled user, across all three callers that trust this function (password login, OIDC, CalDAV), with zero signal it happened. Now logs a warning when this bypass actually applies to an enrolled user; doesn't change the returned bool for any caller.
4. **`pkg/utils/httpclient.go`** - once `transport.Proxy` is set, the dialer's `Control` hook (the SSRF guard) only ever sees the proxy's own dial address - confirmed with a standalone reproduction. `config-raw.json` recommends `outgoingrequests.proxyurl` as the security control for multi-tenant instances, but nothing verified or warned that the configured proxy itself needs to be SSRF-aware for that to actually hold. Added a one-time warning + updated the doc comment.

All four are additive (warnings, cycle detection, symbol filtering) with no behavior change for any existing working flow - only the previously-broken/insecure paths change.

## Verification

`go build`/`go vet`/`gofmt` clean on every touched package. Extended/added tests for each: a TickTick cycle regression test (mutual + self-referencing), a yaegi symbol-filtering unit test, and the existing totp/httpclient suites re-run clean (added a `TestMain` to `pkg/utils` for logger init, needed once a `log.Warningf` call became reachable from that package's tests). `go test` passes for every touched package with no regressions.


pkg/user/totp.go
```diff
@@ -59,12 +59,25 @@ type TOTPPasscode struct {
...
+	if err != nil {
+		return false, err
+	}
+
 	if !config.ServiceEnableTotp.GetBool() {
+		// service.enabletotp is a global kill-switch: it doesn't clear or
+		// flag this user's own totp.enabled row, so an operator disabling it
+		// silently removes 2FA enforcement for every already-enrolled user
+		// without touching their data. Surface that here so it at least
+		// shows up in logs instead of being invisible.
```
kolaente: This feels redundant. If an admin disables it they can check in the database if this will also disable it for users.

pkg/utils/httpclient.go
```diff
@@ -54,6 +71,11 @@ func NewSSRFSafeHTTPClient() *http.Client {
 	proxyPassword := config.OutgoingRequestsProxyPassword.GetString()
 
 	if proxyURL != "" && proxyPassword != "" {
+		if !config.OutgoingRequestsAllowNonRoutableIPs.GetBool() {
+			warnProxyDefeatsSSRFGuardOnce.Do(func() {
```
kolaente: Check this on startup, not when doing the first request

pkg/utils/httpclient.go
```diff
@@ -21,19 +21,36 @@ import (
...
 // NewSSRFSafeHTTPClient returns an *http.Client with SSRF protection applied.
 // It blocks connections to non-globally-routable IP addresses (loopback,
 // private ranges, link-local, etc.) unless outgoingrequests.allownonroutableips
 // is set to true. It also configures proxy settings from outgoingrequests config.
 //
+// If outgoingrequests.proxyurl is also set, that REPLACES this destination-level
+// check rather than supplementing it: once a proxy is configured, the dialer's
+// Control hook only ever sees the proxy's own address, never the real
+// destination, since the proxy resolves and connects to it independently of
+// Go's http.Transport. The configured proxy itself must enforce SSRF policy.
```
kolaente: Toooo long comment. Wayy too long. Please tell your agent to be shorter.

pkg/plugins/yaegi/loader.go
```diff
@@ -53,11 +99,19 @@ func LoadPlugin(dir string) (plugins.Plugin, error) {
 //   - NewUnauthenticatedRouterPlugin() plugins.UnauthenticatedRouterPlugin (optional)
 //   - NewMigrationPlugin() plugins.MigrationPlugin                        (optional)
 func LoadPluginFull(dir string) (*plugins.LoadedYaegiPlugin, error) {
+	// yaegi's stdlib.Symbols grants full OS-level access (file read/write/
+	// delete, process spawning, env vars, raw sockets) and the vikunja
+	// symbols below grant a raw, unscoped DB session - a loaded plugin runs
+	// as a fully-trusted extension with the server process's own privileges,
+	// not a sandboxed third-party runtime. Only load plugin source you wrote
+	// or fully trust.
```
kolaente: remove the comment

pkg/plugins/yaegi/loader.go
```diff
@@ -20,15 +20,61 @@ import (
...
+const dbPackagePath = "code.vikunja.io/api/pkg/db/db"
+
+// unsafeForPlugins lists db symbols that yaegi extract mechanically included
+// but that only make sense for the Go test harness or a validated CLI restore
+// flow (db.WipeEverything's one real caller, pkg/modules/dump/restore.go),
+// never for interpreted plugin code: WipeEverything drops every application
+// table with no confirmation, dry-run, or audit hook. No shipped plugin or
+// yaegi test references any of these (verified: examples/plugins/example and
+// every pkg/plugins/yaegi/*_test.go), so removing them from what plugins can
+// resolve changes no working behavior.
```
kolaente: make this comment wayy shorter

kolaente:
Have you heard of https://github.com/JuliusBrussee/caveman ?

pkg/user/totp.go
kigiela: Removed - reverted the whole hunk.

pkg/utils/httpclient.go
kigiela: Done - checked once at startup now.

pkg/utils/httpclient.go
kigiela: Shortened.

pkg/plugins/yaegi/loader.go
kigiela: Removed.

pkg/plugins/yaegi/loader.go
kigiela: Shortened to two lines.

kolaente:
I have in the meantime implemented the fix for ticktick in https://github.com/go-vikunja/vikunja/pull/3321, please rebase (sorry for not coordinating this better)

kolaente:
lets also wait on https://github.com/go-vikunja/vikunja/pull/3240 before merging this.

PR #3255, opened 2026-07-20
feat(migration): add a ClickUp import module

kigiela:
## Why

ClickUp is one of the more popular task managers Vikunja doesn't have an importer for yet, alongside Trello, Todoist, TickTick, Wekan and Microsoft Todo. We migrated a real production ClickUp workspace (~400 tasks) with this module and it held up well, so figured it was worth contributing back.

## What

Adds `pkg/modules/migration/clickup`, following the same `Migrator` interface and `InsertFromStructure` pattern as the existing importers (closest in shape to `todoist` - hand-written HTTP + JSON structs, since ClickUp has no maintained Go SDK the way Trello does).

Mapping:
- Team (workspace) → a fixed pseudo-parent project ("Migrated from ClickUp"), matching the Todoist/Trello convention
- Space → child project
- Folder (optional in ClickUp) → child project of its Space
- List → a `Bucket` within its Space/Folder project (tasks live here) - the same List→Bucket mapping Trello uses for its own Lists
- Task → title, description, priority (ClickUp's 4 levels mapped to Vikunja's), due/start date, done status + done-at, tags → labels, attachments (downloaded and re-uploaded), and subtask relations via `parent`

**Auth is the one new pattern.** ClickUp doesn't offer OAuth for personal integrations - it's a personal API token pasted directly into the UI. None of the existing migrators have this (Todoist and Microsoft Todo are full OAuth with a client secret; Trello is closest - no secret exchange, but still round-trips through an OAuth redirect). `AuthURL()` returns `""` for ClickUp, which `MigrationHandler.vue` now treats as the signal to render a plain token input instead of a "connect" button or file upload, via a new `isTokenMigrator` flag on the frontend `Migrator` interface. Everything downstream of that (the POST body, the `code` JSON key, the `/migration/clickup/migrate` route) is unchanged from the existing OAuth-migrator plumbing - `Migration.Code` just holds the pasted token instead of an OAuth code.

Gated behind `VIKUNJA_MIGRATION_CLICKUP_ENABLE` (default `false`), same as the other optional migrators.

**Known limitation, left out deliberately to keep this PR reviewable:** assignees aren't mapped. None of the existing migrators do this either (there's no reliable way to match a foreign service's user to a Vikunja account without a lookup the `Migrate(u *user.User)` interface doesn't currently have a session for), so this isn't a regression - just flagging it in case there's appetite for designing that separately.

## Testing

`go test ./pkg/modules/migration/clickup/...` - covers priority/timestamp/status mapping, subtask relation resolution (including a task whose parent falls outside the token's accessible spaces), and attachment download against a local `httptest` server. `golangci-lint run`, frontend `lint`/`lint:styles` clean.

Ran it for real against a ~400-task ClickUp workspace (multiple spaces, folders, and folderless lists) migrating into a fresh Vikunja project - projects, buckets, tasks, labels, subtask relations, and attachments all came through correctly.


kolaente:
Good stuff! I'll have to do a deep review later, but from glancing over it, it seems like it does not need much adjustment.

The reason for the config for the other migrators is that they need more settings like api keys and such. If this migrator does not need that, we should just remove the setting and keep it always on.

pkg/modules/migration/clickup/clickup.go
```diff
@@ -0,0 +1,468 @@
...
+// @Success 200 {object} migration.Status "The migration status"
+// @Failure 500 {object} models.Message "Internal server error"
+// @Router /migration/clickup/status [get]
+func (m *Migration) Name() string {
+	return "clickup"
+}
+
+// AuthURL is empty for ClickUp - it uses a pasted personal token, not an OAuth redirect.
+// @Summary Get the auth url from ClickUp
+// @Description Empty for ClickUp: it authenticates with a personal API token pasted directly into the frontend, not an OAuth redirect.
```
kolaente: Can we make this explicit instead of relying on an empty auth url as signal?

pkg/modules/migration/clickup/clickup.go
```diff
@@ -0,0 +1,468 @@
...
+// @Produce json
+// @Security JWTKeyAuth
+// @Param migrationCode body clickup.Migration true "A personal ClickUp API token (Settings > Apps in ClickUp)."
+// @Success 200 {object} models.Message "A message telling you everything was migrated successfully."
+// @Failure 500 {object} models.Message "Internal server error"
+// @Router /migration/clickup/migrate [post]
+func (m *Migration) Migrate(u *user.User) (err error) {
+	log.Debugf("[ClickUp Migration] Starting migration for user %d", u.ID)
+
+	var pseudoParentID int64 = 1
```
kolaente, line 333: Is this ID just used to relate the projects? because if we're starting at 1, it might clash with existing projects, not sure where that could break.

kolaente:
auggie review

pkg/modules/migration/clickup/clickup.go
kigiela: Done - I split `AuthURL()` off the base `Migrator` interface into a new `OAuthMigrator` interface. `RegisterMigrator` now only registers the `/auth` route for migrators implementing it, so ClickUp (which doesn't) gets no auth route at all instead of signalling with an empty url. The v2 OAuth route registration was updated to take `func() migration.OAuthMigrator` so this stays compile-time checked.

pkg/modules/migration/clickup/clickup.go
kigiela: Yes, these IDs are only used to relate the entries of the hierarchy to each other - `InsertFromStructure` resets each one to 0 before saving (`p.ID = 0`) and rebuilds parent/child relations with the real, newly created IDs via its `projectsByOldID` map, so they can't clash with existing projects. It's the same scheme the Todoist and Trello migrators use (both also start at `pseudoParentID = 1`). I added a comment in the code spelling this out.

pkg/modules/migration/clickup/clickup.go
```diff
@@ -0,0 +1,468 @@
...
+
+func (m *Migration) getTeams() ([]*clickupTeam, error) {
+	resp, err := migration.DoGetWithHeaders(apiBase+"/team", m.authHeader())
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	r := &teamsResponse{}
+	if err := json.NewDecoder(resp.Body).Decode(r); err != nil {
```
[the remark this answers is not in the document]
kigiela: Good catch - fixed. All ClickUp API calls now go through a `decodeResponse` helper that checks the status code before decoding and surfaces the response body in the error (e.g. ClickUp's `{"err":"Token invalid","ECODE":"OAUTH_025"}`), so an invalid token fails the migration instead of silently importing nothing. Added tests for both paths.

pkg/modules/migration/clickup/clickup.go
```diff
@@ -0,0 +1,468 @@
...
+			Title:       ct.Name,
+			Description: ct.Description,
+			BucketID:    bucketID,
+		},
+	}
+
+	if ct.Status.Type == "closed" || ct.Status.Type == "done" {
+		task.Done = true
+		if closed, ok := parseClickupTimestamp(ct.DateClosed); ok {
+			task.DoneAt = closed
```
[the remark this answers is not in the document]
kigiela: Fixed - `DoneAt` is now converted with `.In(config.GetTimeZone())` like `DueDate` and `StartDate`.

frontend/src/views/migrate/MigrationHandler.vue
```diff
@@ -28,6 +28,36 @@
...
+						{{ migrationError }}
+					</Message>
+					<div class="field has-addons">
+						<div class="control is-expanded">
+							<input
+								v-model="migratorAuthCode"
+								class="input"
+								type="password"
+								:placeholder="$t('migrate.tokenPlaceholder', {name: migrator.name})"
+								@keyup.enter="migrate"
```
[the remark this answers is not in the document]
kigiela: Fixed - `migrate()` now returns early for token migrators when the token field is empty, so the Enter key path can't submit an empty token anymore.

kigiela:
Thanks for the review! All feedback addressed in the latest commit:

- **Config setting removed** - you're right, ClickUp needs no server-side settings since the personal API token is supplied by the user in the frontend. The migrator is now always enabled and always advertised in `/info`.
- **Token auth is now explicit** - `AuthURL` moved off the base `Migrator` interface into a new `OAuthMigrator` interface; migrators that don't implement it get no `/auth` route registered.
- Also addressed the three bot findings: HTTP status is checked before decoding responses (an invalid token now fails loudly instead of importing nothing), `DoneAt` gets the configured timezone, and the token input's Enter key can't submit an empty token.

PR #3315, opened 2026-07-27
Add rapid local iteration dev server with auto-restart on both API and UI

JohnStarich:
Adds `mage dev:start` to run auto-restarting servers when code changes.

The frontend already did this to an extent, so this change detects API
updates as well for an even faster development iteration.

This saved me a ton of time while working on [weighted sort](https://github.com/go-vikunja/vikunja/pull/3317) and I thought others could benefit from the automation I created!

All changes:

* Add "mage dev:start" to run both frontend and backend servers
    - Provisions temporary sqlite database until Ctrl+C
    - Automatic API restarts when Go files change
    - Automatic UI refresh (this was already built-in to `npm start`)
    - Registers test user (`test`/`test`)
    - Bump rate limit
    - Track `./frontend/dist` directory to allow automatic UI install + start
* Reduce API logs to configured log level
* Fix duplicate registered signal handler
    - Only one signal handler can be registered in a Go program at one time. One already exists in main.go, so use contexts elsewhere.
* Enable running cmd with args and a context to propagate OS signals and args
    - This one was primarily implemented to pass down the signal-based context. However, this could be extended to perform high level functional tests at the CLI level.


kolaente:
I am not very sure if we actually need this. It is 800 LOC for something that is nice to have and we'll have to maintain. Is there something off-the-shelf that we can use?

I personally never found the need to have hot-reload for the api but your mileage may vary.

JohnStarich:
I agree we need to be careful about what we choose to maintain, it's got to be worth it.

I can certainly cut this one down to a smaller size like removing the auto-open URL, and split it into smaller bits. Maybe one PR for pnpm + test user creation, and a second one for the fswatch API restarter?

I did find this incredibly helpful during my development iterations. I was losing roughly 1 minute per iteration with setting up the in-memory database again and again. This was particularly painful for trying (many) tweaks to the urgency score formula, since it required setting up a user, creating tasks, setting properties on them, creating a saved filter, and then finally viewing the urgency sort.

Regarding an off-the-shelf solution for watches, I'm not aware of one unfortunately. I have used this pattern successfully in another project using the same fsnotify library. I could probably dumb it down with some simplifying assumptions. Thoughts?

As lines of code go, I think the 800+ figure is mostly go.sum (+187), my usual thorough tests (+253), and some nice-to-haves like waiting for the UI to start and open the URL (~100). Without them,  it'd probably sit around `312` and still have the things that made this so useful.

(I'm out on holiday currently, but will return in ~1 week.)

JohnStarich:
And of course if I am unconvincing, it wouldn't bother me (too much :stuck_out_tongue:) to simply close this as "won't fix". It has to be useful to you too!

kolaente:
Let me challenge that more.

What did your process look like before having this?

> I was losing roughly 1 minute per iteration with setting up the in-memory database again and again. This was particularly painful for trying (many) tweaks to the urgency score formula, since it required setting up a user, creating tasks, setting properties on them, creating a saved filter, and then finally viewing the urgency sort.

This sounds like the proper solution would be TDD with an end-to-end test that set it up with all the needed fixtures :)

A bunch of the changes in the diff seem like cleanup and refactors that could go to separate PRs as well.

PR #3414, opened 2026-08-03
feat(filter): support @me and currentUser macros in task assignee queries

### LINKED ISSUE #3413: [Feature Request]: Support @me and currentUser macros in task assignee filter queries
### Summary
Currently, task filter queries require hardcoding specific usernames when filtering by assignees (e.g., `assignees = alice` or `assignees in alice, bob`). When sharing project views or saved filter templates across a team, this requires every team member to create and maintain their own individual filters to see tasks assigned to them.

### Proposed Solution
Support `@me`, `currentUser`, and `current_user` dynamic macros in task filter query strings.

When a query containing `assignees in @me` or `assignees = currentUser` is evaluated, the backend filter engine dynamically resolves the macro to the authenticated user's username (`web.Auth`).

### Use Cases
- Shared project views (e.g. "My Tasks" view in a shared project) where each team member views their own assigned tasks dynamically.
- Saved filter templates across organizations.

mcassaniti:
Resolves #3413

### What this PR does

1. **Backend Dynamic Macro Resolution** (`pkg/models/task_search.go`):
   - Added `resolveCurrentUserMacros` in `dbTaskSearcher.Search`.
   - Filter query terms matching `@me`, `currentUser`, or `current_user` in assignee queries dynamically resolve to the logged-in user's username (`web.Auth`). **UPDATE**: Now only `@me`.
   - The stored filter expression in the database retains `@me` so shared views resolve dynamically per user.

2. **Backend Unit Tests** (`pkg/models/task_search_test.go`):
   - Added `TestTaskSearchCurrentUserMacro` testing string, slice, and array macro replacement.

3. **Frontend UX & Docs** (`FilterAutocomplete.ts` & `FilterInputDocs.vue`):
   - Prepended `@me` (`Current User (@me)`) to autocomplete suggestions for `assignees`.
   - Added `assignees in @me` example to the filter help documentation.

### Testing
- Go unit test `TestTaskSearchCurrentUserMacro` passed.
- 1,084 frontend Vitest unit tests passed.

> **Disclaimer**: This PR was prepared with AI assistant pair programming. All code changes, database queries, unit tests, and commits have been reviewed, GPG-signed, and verified against the repository build pipeline.

### DIFF
--- a/frontend/src/components/input/filter/FilterAutocomplete.ts
+++ b/frontend/src/components/input/filter/FilterAutocomplete.ts
@@ -238,6 +238,15 @@ export default Extension.create<FilterAutocompleteOptions>({
 								} else {
 									assigneeSuggestions = await userService.getAll({} as IUser, {s: autocompleteContext.search}) as SuggestionItem[]
 								}
+								if (autocompleteContext.search === '' || '@me'.startsWith(autocompleteContext.search.toLowerCase())) {
+									const meMacro = {
+										id: 0,
+										username: '@me',
+										title: '@me',
+										name: 'Current User (@me)',
+									} as unknown as SuggestionItem
+									assigneeSuggestions = [meMacro, ...assigneeSuggestions]
+								}
 								// For assignees, show suggestions even with empty search, but limit if we have many
 								if (autocompleteContext.search === '' && assigneeSuggestions.length > 10) {
 									assigneeSuggestions = assigneeSuggestions.slice(0, 10)
--- a/frontend/src/components/input/filter/FilterInputDocs.vue
+++ b/frontend/src/components/input/filter/FilterInputDocs.vue
@@ -62,6 +62,7 @@ const showDocs = ref(false)
 				{{ $t('filters.query.help.examples.undoneHighPriority') }}
 			</li>
 			<li><code>assignees in user1, user2</code>: {{ $t('filters.query.help.examples.assigneesIn') }}</li>
+			<li><code>assignees in @me</code>: {{ $t('filters.query.help.examples.assigneesIn') }} (@me)</li>
 			<li>
 				<code>(priority = 1 || priority = 2) &amp;&amp; dueDate &lt;= now</code>:
 				{{ $t('filters.query.help.examples.priorityOneOrTwoPastDue') }}
--- a/pkg/models/task_search.go
+++ b/pkg/models/task_search.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"code.vikunja.io/api/pkg/db"
+	"code.vikunja.io/api/pkg/user"
 	"code.vikunja.io/api/pkg/web"
 
 	"xorm.io/builder"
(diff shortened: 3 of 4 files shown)


kolaente:
What happens when a user has `me` as their username?

mcassaniti:
Good question. For the frontend, the auto-complete will not match. Something like `@m` will however. This is why the frontend code uses `startsWith()`. The backend has exact matching regardless --> `'me' != '@me'`.

I guess the code should reserve these special values and not allow them as a username. I'm fine if you want to change what is matched (e.g.: only `currentUser` is a valid pattern).

ultinate:
Proposal: Only use `@me` as pattern. Not `currentUser` or `current_user` to avoid any conflicts with actual user names. This could extend to future patterns like `@all` or `@admins`.

This assumes that usernames starting with `@` are not allowed.

mcassaniti:
Done. The filter is restricted to `@me` only.

frontend/src/components/input/filter/FilterAutocomplete.ts
```diff
@@ -238,6 +238,15 @@ export default Extension.create<FilterAutocompleteOptions>({
 								} else {
 									assigneeSuggestions = await userService.getAll({} as IUser, {s: autocompleteContext.search}) as SuggestionItem[]
 								}
+								if (autocompleteContext.search === '' || '@me'.startsWith(autocompleteContext.search.toLowerCase())) {
```
kolaente, line 241: Does that mean it will filter to the current user when no user to filter for is provided?

frontend/src/components/input/filter/FilterAutocomplete.ts
```diff
@@ -238,6 +238,15 @@ export default Extension.create<FilterAutocompleteOptions>({
...
+									const meMacro = {
+										id: 0,
+										username: '@me',
+										title: '@me',
+										name: 'Current User (@me)',
```
kolaente, line 246: Please word this differently. It is viewed from the current user's pov, so at least make it "You" - arguably even better would be to use the current user's name and avatar but that's a bigger change.

frontend/src/components/input/filter/FilterInputDocs.vue
```diff
@@ -62,6 +62,7 @@ const showDocs = ref(false)
 				{{ $t('filters.query.help.examples.undoneHighPriority') }}
 			</li>
 			<li><code>assignees in user1, user2</code>: {{ $t('filters.query.help.examples.assigneesIn') }}</li>
+			<li><code>assignees in @me</code>: {{ $t('filters.query.help.examples.assigneesIn') }} (@me)</li>
```
kolaente, line 65: This also needs an update in the docs.

pkg/models/task_search.go
```diff
@@ -443,9 +444,67 @@ func buildParentSearchCondition(search string) builder.Cond {
 	return cond
 }
 
+func replaceUserMacro(val string, username string) string {
+	valTrim := strings.TrimSpace(val)
+	if valTrim == "@me" {
```
kolaente, line 449: we also need to make sure no user can be called `@me` and a migration path for existing users who are called that.

Please change the username validator so that it rejects everything that starts with `@`. Open to ideas on how to make the rejection work for current users who upgrade, that smells like a breaking change.

kolaente (changes requested):

frontend/src/components/input/filter/FilterAutocomplete.ts
mcassaniti: I had to check and no, it instead means that `@me` is added as a suggestion when nothing is typed in yet.

frontend/src/components/input/filter/FilterAutocomplete.ts
mcassaniti: So `name: 'You (@me)'` ? Let's get that agreed on first and I'll make the change.

mcassaniti:
There are two parts to the feedback wanted.

# Disallow users with `@` and handle migration

My suggestions would be:

1. Block user creation for a username that starts with `@` regardless of the source (that includes OIDC and LDAP).
1. Migrate user names from `@user` to `user` after checking the target name does not exist regardless of source. That means a silent migration of `@user` to `user` for an LDAP sync. This is migration only, so a new LDAP user starting with `@` is blocked.
1. Allow login of a user name of the form `@user` by looking for `user` instead, but only after trying `@user`.
1. The UI needs some adjustment to warn when logging in as a user starting with `@`, and an existing user who couldn't be migrated needs to know their user name is now deprecated.

Assuming a user of `@me` still exists post migration, the user cannot be selected in a UI drop-down. At least they can still login.

# The name hint for `@me`

I want to be very clear on what the hinted user name is when someone types `@me` for auto-fill. My current suggestion was `Current User (@me)` but I think you wanted it to be `You (@me)`. I don't really mind, but I want it to be very clear exactly what that value is because we cannot go back. I'm leaning towards `Current User` or `Logged In User`.

frontend/src/components/input/filter/FilterAutocomplete.ts
kolaente: Sounds good. It also needs a translation.

kolaente:
> Block user creation for a username that starts with @ regardless of the source (that includes OIDC and LDAP).

Yes. That will make this a breaking change, which means it will take a while until we can get this merged (other things need to go first).

Worth splitting out the new rule and the migration of existing usernames out first and then stacking the filter change on top.

mcassaniti:
Closing. See #3816 and #3817 for follow-up work.

PR #3433, opened 2026-08-04
feat: add default due time settings

### LINKED ISSUE #1907: Default Due Time
### Description

It's impossible for tasks to be due "some day". Only a day and a time. Having a task due today can be achieved by setting the due time to 23:59. However, the default due time is 12:00. This makes it so that you have to change the time for every task you create, which feels like an unnecessary extra step.

I'd suggest adding an option like "Day ends at" under General Settings > Projects & Tasks to set a custom due time. This keeps the flexibility for users to pick their own end of day, whether that is noon, after work, bedtime or 23:59.

I am aware has been [discussed](https://community.vikunja.io/t/default-due-time-configuration/1253/10) before but I couldn't find an issue about it on GitHub.

Setting the default time to 23:59 for newly created tasks.

ThiloSavary:
Resolves #1907.

## Summary
Adds a user-specific **Default Due Time** setting.

- Stored in frontend settings only; does not affect the API.
- **Quick Add Magic**: if a date is set without a time, uses the Default Due Time when configured, otherwise unchanged.
- **Task detail page**: "Set due date" time input is pre-filled with the Default Due Time when configured, otherwise unchanged.

## Screenshots

grafik

Creating a new task:
grafik
Leads to
grafik

## Testing
- Added tests for Quick Add Magic
- Added tests for new due date helpers

## AI Disclaimer
Used Copilot inline suggestions during development.

### DIFF
--- a/frontend/src/helpers/time/getDefaultDateForDay.ts
+++ b/frontend/src/helpers/time/getDefaultDateForDay.ts
@@ -34,7 +34,7 @@ export function getDefaultDueTimeParts(date: Date): {hours: number, minutes: num
 	}
 }
 
-export function getDefaultDueDateForDay(date: Date): Date {
+export function getDefaultDateForDay(date: Date): Date {
 	const newDate = new Date(date)
 	const defaultDueTime = getDefaultDueTimeParts(newDate)
 	newDate.setHours(defaultDueTime.hours, defaultDueTime.minutes, 0, 0)
--- a/frontend/src/helpers/time/getDefaultDueDateForDay.ts
+++ b/frontend/src/helpers/time/getDefaultDueDateForDay.ts
@@ -1,0 +1,44 @@
+import {calculateNearestHours} from '@/helpers/time/calculateNearestHours'
+import {useAuthStore} from '@/stores/auth'
+
+function parseUserDefaultDueTime(defaultDueTime?: string): {hours: number, minutes: number} | null {
+	if (!defaultDueTime) {
+		return null
+	}
+
+	const match = /^(\d{2}):(\d{2})$/.exec(defaultDueTime)
+	if (!match) {
+		return null
+	}
+
+	const hours = Number(match[1])
+	const minutes = Number(match[2])
+	if (hours > 23 || minutes > 59) {
+		return null
+	}
+
+	return {hours, minutes}
+}
+
+export function getDefaultDueTimeParts(date: Date): {hours: number, minutes: number} {
+	const defaultDueTime = useAuthStore().settings.frontendSettings.defaultDueTime
+	const parsedTime = parseUserDefaultDueTime(defaultDueTime)
+
+	if (parsedTime !== null) {
+		return parsedTime
+	}
+
+	return {
+		hours: calculateNearestHours(date),
+		minutes: 0,
+	}
+}
+
+export function getDefaultDueDateForDay(date: Date): Date {
+	const newDate = new Date(date)
+	const defaultDueTime = getDefaultDueTimeParts(newDate)
+	newDate.setHours(defaultDueTime.hours, defaultDueTime.minutes, 0, 0)
+	return newDate
+}
+
+export {parseUserDefaultDueTime}
--- a/frontend/src/views/user/settings/General.vue
+++ b/frontend/src/views/user/settings/General.vue
@@ -57,6 +57,15 @@
 				/>
 			</FormField>
 			<FormField
+				:label="$t('user.settings.general.defaultDueTime')"
+				layout="two-col"
+			>
+				<FormInput
+					v-model="settings.frontendSettings.defaultDueTime"
+					type="time"
+				/>
+			</FormField>
+			<FormField
 				v-if="hasFilters"
 				:label="$t('user.settings.general.filterUsedOnOverview')"
 				layout="two-col"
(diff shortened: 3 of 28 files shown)


frontend/src/helpers/time/getDefaultDueDateForDay.ts
```diff
@@ -0,0 +1,44 @@
...
+		return parsedTime
+	}
+
+	return {
+		hours: calculateNearestHours(date),
+		minutes: 0,
+	}
+}
+
+export function getDefaultDueDateForDay(date: Date): Date {
```
kolaente: That's used for many things, not just Due Dates. Please rename the function. Same everywhere else.

frontend/src/views/user/settings/General.vue
```diff
@@ -56,6 +56,15 @@
 					:options="minimumPriorityOptions"
 				/>
 			</FormField>
+			<FormField
```
kolaente, line 59: This needs an explanation of where the default time is used

kolaente (changes requested):

frontend/src/helpers/time/getDefaultDueDateForDay.ts
ThiloSavary: Renamed to `getDefaultDateForDay`

frontend/src/views/user/settings/General.vue
ThiloSavary: Added some help text

ThiloSavary:
Thanks for the feedback! I've addressed the review comments and pushed the updates.

ThiloSavary:
The e2e test showed, that i missed the old behavior of datepicker, which used 12:00 as default time, while my new implementaion used the nearest Hour. 

Fixed that, current behavior for the datepicker is now: 

- If there is a due date set by the user, use that
- Otherwise use 12:00 (as before)
- "Quick selections" as 'Today' and 'Tommorow' now check
  - If there is a due date set by the user, use that
  - Otherwise use the nearest-hour behavior (as before)

frontend/src/helpers/time/getDefaultDateForDay.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function getDefaultDateForDay(date: Date): Date {
+	const newDate = new Date(date)
+	const defaultDueTime = getDefaultDueTimeParts(newDate)
+	newDate.setHours(defaultDueTime.hours, defaultDueTime.minutes, 0, 0)
+	return newDate
+}
+
+export {parseUserDefaultDueTime}
```
kolaente: why the extra export?

frontend/src/helpers/time/getDefaultDateForDay.ts
```diff
@@ -0,0 +1,44 @@
+import {calculateNearestHours} from '@/helpers/time/calculateNearestHours'
+import {useAuthStore} from '@/stores/auth'
+
+function parseUserDefaultDueTime(defaultDueTime?: string): {hours: number, minutes: number} | null {
```
kolaente:
```suggestion
function parseUserDefaultTime(defaultDueTime?: string): {hours: number, minutes: number} | null {
```

frontend/src/helpers/time/getDefaultDateForDay.ts
```diff
@@ -0,0 +1,44 @@
...
+	const hours = Number(match[1])
+	const minutes = Number(match[2])
+	if (hours > 23 || minutes > 59) {
+		return null
+	}
+
+	return {hours, minutes}
+}
+
+export function getDefaultDueTimeParts(date: Date): {hours: number, minutes: number} {
```
kolaente:
```suggestion
export function getDefaultTimeParts(date: Date): {hours: number, minutes: number} {
```

frontend/src/helpers/time/getDefaultDateForDay.ts
```diff
@@ -0,0 +1,44 @@
...
+		return parsedTime
+	}
+
+	return {
+		hours: calculateNearestHours(date),
+		minutes: 0,
+	}
+}
+
+export function getDefaultDateForDay(date: Date): Date {
```
kolaente: This should be `addTimeToDate` or something like it because the current name implies that it always adds a default, but it also runs `caluclateNearestHours` which is unrelated to defaults

kolaente (changes requested):

kolaente:
Please don't force push, that makes it harder to review what actually changed.

frontend/src/helpers/time/getDefaultDateForDay.ts
ThiloSavary: Removed, not sure why i added it

ThiloSavary:
I renamed the functions. Not sure about the `getDateWithTime` name, but cant think of something better.

> Please don't force push, that makes it harder to review what actually changed.

Sorry, somehow I messed up the history while syncing with upstream, and force-pushing seemed like a quick fix. I didn’t realize it would cause extra work for you.

This is my first time contributing code on GitHub, so I’m still a bit clumsy with the workflow. Feedback like this is very welcome!

kolaente (approved):
Looks good now! Thanks

PR #3436, opened 2026-08-04
fix(tasks): don't skip a month when a monthly task repeats from the 29th-31st

TowyTowy:
`addOneMonthToDate` passes `d.Month()+1` with the original day into `time.Date`, which normalizes an out-of-range day forward rather than clamping it.

A task with repeat mode "Monthly" and a due date of Jan 31 gets a new due date of **Mar 3** — February is skipped, and the day of month stays shifted for every occurrence after that (Apr 3, May 3, ...). Same for Mar 31 -> May 1 and May 31 -> Jul 1. It applies to the reminders, start date and end date of the mode too.

This clamps the day to the last day of the target month instead, so Jan 31 repeats to Feb 28 (Feb 29 in a leap year). That matches the "Monthly" label in the UI, and the `FREQ=MONTHLY;BYMONTHDAY=<n>` rule the CalDAV export already emits for this mode — under RFC 5545 an occurrence can never land in the month after next.

Added a table test for `addOneMonthToDate` (short month, leap February, 30-day month, year boundary) and an `updateDone` case for the end-of-month due date. Both fail on main and pass with the fix; the rest of `pkg/models` is unaffected — no fixture uses `repeat_mode: 1`.

I'm aware of #3071 replacing this code path for 3.0.0. Filing anyway since the current behaviour ships until then — happy to close it if you'd rather not carry the change.

AI disclosure: written with Claude Code, reviewed and tested before opening.

### DIFF
--- a/pkg/models/tasks.go
+++ b/pkg/models/tasks.go
@@ -1718,8 +1718,22 @@ func (t *Task) moveTaskToDefaultBuckets(s *xorm.Session, a web.Auth, views []*Pr
 	return nil
 }
 
+// addOneMonthToDate advances d into the next calendar month, clamping the day
+// to that month's last day. time.Date normalizes an out-of-range day instead,
+// which made a monthly task due Jan 31 repeat to Mar 3 - skipping February
+// entirely and permanently shifting the day of month.
 func addOneMonthToDate(d time.Time) time.Time {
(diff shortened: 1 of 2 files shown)


kolaente:
I think this will be made obsolete by https://github.com/go-vikunja/vikunja/pull/3071

kolaente:
Closing as (and you noted this correctly) this will be superseeded by https://github.com/go-vikunja/vikunja/pull/3071, adding this now just means removing it later

PR #3460, opened 2026-08-07
add mailer no encryption support

lublak:
i use a smtp bridge (tunnel) which don't uses encryption

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -132,6 +132,7 @@ const (
 	MailerQueuelength   Key = `mailer.queuelength`
 	MailerQueueTimeout  Key = `mailer.queuetimeout`
 	MailerForceSSL      Key = `mailer.forcessl`
+	MailerDisableSSL    Key = `mailer.disablessl`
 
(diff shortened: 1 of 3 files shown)


lublak:
@kolaente is it possible to request a review for this? It would be nice if it gets into the next release.

pkg/config/config.go
```diff
@@ -132,6 +132,7 @@ const (
 	MailerQueuelength   Key = `mailer.queuelength`
 	MailerQueueTimeout  Key = `mailer.queuetimeout`
 	MailerForceSSL      Key = `mailer.forcessl`
+	MailerDisableSSL    Key = `mailer.disablessl`
```
kolaente, line 135: Now we have two settings that do SSL things, that should be one setting

kolaente (changes requested):

pkg/config/config.go
lublak: That's correct, but unfortunately it would be a breaking change, and I wanted to avoid that. Of course, if you say that's okay with you, I'd be happy to finish it up that way.

PR #3463, opened 2026-08-08
feat: Web Push notifications

Tokra110:
## What

Adds Web Push notifications for the installed PWA on Android, iOS/iPadOS 16.4+, and desktop, delivering the same notifications Vikunja already sends by mail and shows in the in-app feed.

## How it works

- **Backend**: a durable, restart-safe delivery worker (leased rows, exponential backoff, dedup) sends VAPID-encrypted Web Push messages. Two additive tables (`web_push_subscriptions`, `web_push_deliveries`) via migration. Config: `webpush.enabled` plus a VAPID key pair (`vikunja webpush generate-keys`); requires an HTTPS `service.publicurl`. The public key is exposed on `/info`. Prometheus outcome metrics and an SSRF-safe HTTP client are included.
- **Frontend**: a settings control to enable / test / disable per device, a service worker `push` / `notificationclick` handler, subscription reconcile on load, and local unsubscribe on logout.
- **Routed to push**: reminders, overdue, comments, assignments, mentions, task deletion, project created, team member added, and API-token-expiry.

## Design decisions worth a reviewer's eye

1. **Subscription lifetime — revoked only on real events.** A subscription is removed on explicit logout, account deletion, disabling push, or a 404/410 from the push service. It is deliberately **not** tied to login-session inactivity: push is most useful while the app is closed, so gating delivery on session activity would silently stop notifications after the JWT TTL.
2. **Notifications show even while a tab is focused.** The service worker shows the OS notification and also refreshes the in-app list. Easy to switch to suppress-when-focused if that is preferred.
3. **VAPID contact (`sub`).** Uses `mailer.fromemail` when it is a real address, otherwise synthesizes `webmaster@<public-host>`. `webpush-go` prepends `mailto:`, so a bare address is passed on purpose: a `mailto:` value would double to `mailto:mailto:`, which Apple rejects with `403 BadJwtToken` (FCM tolerates it). Happy to add an explicit `webpush.contactemail` key instead.
4. **Delivery timestamps compared in UTC.** Stored timestamps are UTC, but xorm does not convert a `time.Time` bound into a raw `Where` string to the database time zone. On SQLite in a non-UTC process time zone this made the worker treat every fresh delivery as already expired and never send it (masked on UTC/CI and on Postgres). The worker now normalizes its comparisons to UTC.

## New dependency

`github.com/SherClockHolmes/webpush-go` v1.4.0 (MIT), which pulls in `golang-jwt/jwt/v5`.

## Testing

- **Backend**: unit + huma endpoint tests — worker outcomes, dedup, lease recovery/release, subscriber formatting, and delivery surviving an idle/removed session. `mage test:feature` and `mage test:web` pass; `golangci-lint` clean on the changed packages.
- **Frontend**: vitest unit tests + a Playwright e2e spec.
- **Verified on real devices**: Android (Chrome PWA) and iOS 18 (installed Safari PWA).

## Notes

- iOS requires the site installed to the Home Screen (iOS/iPadOS 16.4+); the UI communicates this state.
- Docs in `docs/web-push.md`; config documented in `config-raw.json`.


kolaente:
This is pretty big and I'm unsure about the direction, will take a moment until I get to properly review it

PR #3465, opened 2026-08-09
fix(frontend): make the editor toolbar fit on a single line

ngoc-minh-do:
## Summary

The TipTap editor toolbar wraps onto a second line in narrow containers, pushing later buttons (undo/redo, table) off-screen. This compacts it to always fit on one line:

- Removed per-segment separators; segments become invisible layout containers (`display: contents`)
- Smaller buttons (2rem → 1.625rem) and icons
- `flex-wrap: nowrap` with `gap: .125rem`
- Table option buttons now flow properly in their own wrapped row

Old
image

New

image

## Testing

- `cd frontend && pnpm lint:fix && pnpm lint:styles:fix`
- Manually verified at task-detail width in the dev build — toolbar stays on one line

## AI Disclosure

Generated with AI assistance; reviewed and verified locally.

### DIFF
--- a/frontend/src/components/input/editor/TipTap.vue
+++ b/frontend/src/components/input/editor/TipTap.vue
@@ -1038,6 +1038,9 @@ watch(
 .tiptap__editor {
 	&.tiptap__editor-is-edit-enabled {
 		min-block-size: 10rem;
+		background: var(--grey-200);
+		border: 1px solid var(--grey-300);
+		border-radius: $radius;
 
 		.ProseMirror {
 			padding: .5rem;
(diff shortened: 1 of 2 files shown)


frontend/src/components/input/editor/TipTap.vue
```diff
@@ -1038,6 +1038,9 @@ watch(
 .tiptap__editor {
 	&.tiptap__editor-is-edit-enabled {
 		min-block-size: 10rem;
+		background: var(--grey-200);
```
kolaente, line 1041: Please don't add a background to the editor, I don't think that looks good

kolaente (changes requested):

PR #3479, opened 2026-08-10
fix(examples): exclude Yaegi plugin example from 'go build ./...'

### LINKED ISSUE #3478: examples/plugins/example: 'go build ./...' fails (function main is undeclared)
## Summary
`go build ./...` fails on the example plugin in `examples/plugins/example/main.go`:

```
# code.vikunja.io/api/examples/plugins/example
runtime.main_main·f: function main is undeclared in the main package
```

The file is `package main` but has no `func main()`. It is a Yaegi plugin entrypoint that is evaluated at runtime via its exported factory functions (`NewPlugin`, `NewAuthenticatedRouterPlugin`, …), so it intentionally has no standalone `main`. A plain `func main() {}` would fix `go build` but breaks the Yaegi loader (`pkg/plugins/yaegi` tests then fail with `undefined selector: NewPlugin`).

## Expected behavior
`go build ./...` should succeed (so tools that build the whole module, e.g. `hermes verify`, pass), while `pkg/plugins/yaegi` tests keep loading the example plugin correctly.

## Proposed fix
Add `//go:build ignore` to the plugin file. This excludes it from `go build ./...` (it is not a standalone binary) while Yaegi still compiles and loads it at runtime. Verified: `go build ./...` passes and `go test ./pkg/plugins/...` stays green.

## Environment
- Vikunja: v2.5.0 (current `main`)
- Go: 1.26.x

webcultist:
## What
Fixes `go build ./...` failing on `examples/plugins/example/main.go` with `function main is undeclared in the main package`.

The example is a Yaegi plugin entrypoint: it is `package main` but has no `func main()` because Vikunja loads it at runtime via its exported factory functions (`NewPlugin`, `NewAuthenticatedRouterPlugin`, …). A plain `func main() {}` would fix the build but break the Yaegi loader (`pkg/plugins/yaegi` tests then fail with `undefined selector: NewPlugin`).

## Change
Add `//go:build ignore` to the plugin file. This excludes it from `go build ./...` (it is not a standalone binary) while Yaegi still compiles and loads it at runtime.

## Verification
- `go build ./...` → passes
- `go test ./pkg/plugins/...` → `ok  code.vikunja.io/api/pkg/plugins/yaegi`

Closes #3478.

### DIFF
--- a/frontend/src/components/project/views/ProjectCalendar.vue
+++ b/frontend/src/components/project/views/ProjectCalendar.vue
@@ -1,0 +1,331 @@
+<script setup lang="ts">
+import {computed, onMounted, ref, watch} from 'vue'
+import {useRouter} from 'vue-router'
+import TaskService from '@/services/task'
+import type {ITask} from '@/modelTypes/ITask'
+import {useTaskStore} from '@/stores/tasks'
+
+const props = defineProps<{
+	projectId: number
+	viewId: number
+	isLoadingProject?: boolean
+}>()
+
+const router = useRouter()
+const taskStore = useTaskStore()
+const tasks = ref<ITask[]>([])
+const loading = ref(false)
+const currentMonth = ref(new Date())
+const viewMode = ref<'month' | 'week'>('month')
+
+function startOfPeriod(d: Date): Date {
+	if (viewMode.value === 'month') {
+		return new Date(d.getFullYear(), d.getMonth(), 1)
+	}
+	const startWeekday = (d.getDay() + 6) % 7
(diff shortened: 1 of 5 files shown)


frontend/src/components/project/views/ProjectCalendar.vue
kolaente, line 1: this is unrelated, please rebase and exclude

kolaente (changes requested):

kolaente:
Superseeded by https://github.com/go-vikunja/vikunja/pull/3575

PR #3497, opened 2026-08-14
fix: sort saved filters alphabetically in sidebar navigation 🤖🤖🤖

### LINKED ISSUE #3496: Saved filters in sidebar have no meaningful sort order
## Description

Filters shown in the left-hand navigation panel (under "Filters", separate from Projects and Favorites) are not sorted in any meaningful way — they appear in roughly insertion/DB-row order.

## Root cause

- `SavedFilter.ToProject()` (`pkg/models/saved_filters.go`) converts a saved filter into a pseudo-`Project` for the API response but never sets `Position`, so it defaults to `0` for every filter.
- `getSavedFiltersForUser` (`pkg/models/saved_filters.go`) queries saved filters with no `ORDER BY` clause.
- On the frontend, `frontend/src/stores/projects.ts` sorts the combined projects+filters array by `a.position - b.position`. Since all filters tie at position `0`, this sort is a no-op for them, and they end up ordered by whatever (unordered) order the DB returned rows in.
- Consistent with this, `Navigation.vue` renders the saved-filters section with `:can-edit-order="false"`, confirming there's no real support for manual reordering of filters today.

## Expected behavior

Filters should be sorted predictably (e.g. alphabetically by title as a baseline), and ideally support explicit user-defined ordering like projects do.

## Suggested fix

- Short-term: sort saved filters lexicographically by title in the frontend as a stopgap.
- Longer-term: give saved filters a real `position` field (backed by the DB) so they can be manually reordered like projects.

acwilsondev:
## Summary

Fixes #3496. Saved filters in the left-hand sidebar had no meaningful sort order — `SavedFilter.ToProject()` never sets `Position`, so every filter defaults to `0` and ties in the existing `a.position - b.position` sort in `projects.ts`, leaving them in effectively DB row order.

This sorts saved filters lexicographically by title as a stopgap, matching the `localeCompare` pattern already used elsewhere in the store (labels, teams, etc.). A real `position` field for filters (to support manual reordering like projects) is a larger change tracked separately in the linked issue.

## Test plan
- Added unit tests in `frontend/src/stores/projects.test.ts` covering: alphabetical ordering despite tied `position`, exclusion of archived filters, and exclusion of regular projects / the favorites pseudo-project
- `pnpm test:unit` — 1097/1097 passing
- `pnpm lint` / `pnpm typecheck` — no new errors introduced

## Disclosure
This change was written with AI assistance (Claude Code), reviewed and tested by me before opening this PR.

### DIFF
--- a/frontend/src/stores/projects.test.ts
+++ b/frontend/src/stores/projects.test.ts
@@ -147,6 +147,45 @@ describe('project store', () => {
 		})
 	})
 
+	describe('savedFilterProjects', () => {
+		it('should sort filters lexicographically by title regardless of position', () => {
+			const store = useProjectStore()
+			const filterC = createMockProject({id: -3, title: 'Charlie', position: 0})
+			const filterA = createMockProject({id: -4, title: 'Alpha', position: 0})
+			const filterB = createMockProject({id: -5, title: 'Bravo', position: 0})
+
+			store.setProject(filterC)
+			store.setProject(filterA)
+			store.setProject(filterB)
+
+			const titles = store.savedFilterProjects.map(p => p.title)
+			expect(titles).toEqual(['Alpha', 'Bravo', 'Charlie'])
+		})
+
+		it('should exclude archived filters', () => {
+			const store = useProjectStore()
+			const archivedFilter = createMockProject({id: -3, title: 'Archived', isArchived: true})
+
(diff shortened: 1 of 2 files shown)


kolaente (approved):
Thanks!

Issue #3483, opened 2026-08-11
CalDAV: saved filters (e.g. negative-id pseudo-projects) reject writes with 500 and report cross-collection hrefs


kzssc:
## Pre-submission checklist

- [x] I have searched for existing open or closed issue reports with the same problem.

## Description

Saved filters (e.g. the built-in "My Open Tasks" pseudo-project, negative project id) are exposed over CalDAV as if they were regular calendars, but they don't behave like one:

1. Writing a new task to a saved filter's CalDAV collection fails server-side with an unhandled error, instead of a clean rejection.
2. The filter's CalDAV collection reports child resources (task `.ics` files) whose `href` lives under a *different* collection's URL path than the one being queried — a violation of the WebDAV containment expectation that all `Depth: 1` children of a collection live directly under that collection's own URI. This is very likely why iOS Reminders/Calendar shows the list as completely empty on-device, even though the server has real data for it.

### Steps to reproduce

**Part A — write fails:**

1. Have a Vikunja instance with a saved filter (e.g. "My Open Tasks", project id `-2`).
2. `PUT` a new valid VTODO resource to `/dav/projects/-2/<new-uuid>.ics`.
3. Observe the response.

**Part B — cross-collection hrefs / empty on iOS:**

1. Query the filter's collection with `Depth: 1`:
   ```
   PROPFIND /dav/projects/-2/  (Depth: 1)
   ```
   or a `calendar-query` `REPORT` with a `VTODO` filter.
2. Inspect the `<D:href>` of each returned child resource.
3. Add the same account as a CalDAV account on iOS and look at the corresponding list.

### Expected behavior

- Part A: either the write should succeed by creating the task in some sensible real project, or it should be rejected with a clear 4xx (e.g. 403/405) explaining that filters aren't writable — not an unhandled 500.
- Part B: every child resource returned under `/dav/projects/-2/` should have an `href` that is actually nested under `/dav/projects/-2/...`, so CalDAV clients can treat it as a normal collection. Tasks should be visible in the iOS Reminders list matching what the server has.

### Actual behavior

**Part A:** `PUT /dav/projects/-2/<uuid>.ics` with a valid `VTODO` body returns `HTTP/2 500` with an empty response body. A subsequent `GET` on the same URL returns `404` — the task was never created anywhere.

[...]

PR #3530, opened 2026-08-18
Hide filters in caldav

cheriimoya:
Currently as discussed in #3483, pseudo projects are exposed via CalDAV. This PR hides the filters and adds a config option `enablecaldavpseudoprojects` which defaults to false to restore the current behavior.

It also fixes a lint issue I found in a test.

I used AI for this.

### DIFF
--- a/config-raw.json
+++ b/config-raw.json
@@ -64,6 +64,11 @@
                     "comment": "Enable the caldav endpoint, see the docs for more details"
                 },
                 {
+                    "key": "enablecaldavpseudoprojects",
+                    "default_value": "false",
+                    "comment": "Expose saved filters and the Favorites project as caldav calendars. Off by default: their tasks live in other projects, so most clients show these calendars as empty."
+                },
+                {
                     "key": "motd",
                     "default_value": "",
                     "comment": "Set the motd message, available from the /info endpoint"
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -51,6 +51,7 @@ const (
 	ServiceUnixSocketMode                 Key = `service.unixsocketmode`
 	ServicePublicURL                      Key = `service.publicurl`
 	ServiceEnableCaldav                   Key = `service.enablecaldav`
+	ServiceEnableCaldavPseudoProjects     Key = `service.enablecaldavpseudoprojects`
 	ServiceRootpath                       Key = `service.rootpath`
 	ServiceMaxItemsPerPage                Key = `service.maxitemsperpage`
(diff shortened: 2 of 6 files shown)


kolaente:
Hey thanks for this. I took a different approach in https://github.com/go-vikunja/vikunja/pull/3551 which fixes the same problem more broadly and for cases where projects are shared read only.

cheriimoya:
hmm i cherry picked the commits onto v2.5.0 and tried it out. on my tasks.org app, tasks show up twice now. i can mark items as completed (synced to vikunja) but i cannot uncomplete them (locally undone but still done in vikunja). when i update a task on vikunja which is also covered by a filter (using a filter `id = 123` for testing), it seems like the task is properly handled in the actual project but not updated in the filter project. it seems quite flaky.

do you think this PR is still valid for hiding pseudo items completely via caldav with an opt-in option?

cheriimoya:
@kolaente unsure, if you've seen my comment on this closed PR, do you think the hiding option as an opt-in would benefit upstream?

kolaente:
I'd like to avoid more config options if possible, because it makes things more complicated for users. Do you still have the flakyness with 2.6.0?

cheriimoya:
i took some time to further look into this and tried it out on 2.6.0

my structure is this:

image

and my filters are

- done today `(done = true) && (updated >= now/d)`
- modified today `updated >= now/d`
- only personal `done = false && project in personal, personal-waiting, personal-someday`
- only work `done = false&&project in work, work-waiting, work-someday`

when all filters are exposed as projects via caldav, the list of all my tasks looks like this:

Screenshot_20260905-093909

that leaves the part #3551 wasn't aimed at: all tasks that are represented in a project AND included in a filter are displayed twice which is undesirable for me as my task list is roughly double.

i take your point on config options, so i'd drop the server level setting i proposed.
what about a property on the saved filter itself instead, a "sync via caldav" checkbox in the filter's settings.
it's per user and per object rather than an instance wide setting, and it only appears to someone already editing a filter.
it is on by default so nothing changes for anyone? 

happy to redo the pr that way if you think it's worth having.

about the flakiness: my tasks.org client stores modifications locally and regularily (or by pulling down from the top) syncs the local changes with the upstream via caldav.
now when i move a task from one filter to another, the move works as all things are local first. when a sync happens, vikunja returns 403 which is correct from the current vikunja implementation and the client doesn't update anything => local state and vikunja state are out of sync. this explains the flakiness and it isn't a vikunja bug.

PR #3606, opened 2026-08-27
fix(editor): preserve subscript and superscript in pasted text

### LINKED ISSUE #1222: Pasting into the task description doesn't preserve <sub> and <sup> tags
### Description

Vikunja's task description editor is an HTML WYSIWYG thing. Is it intentional that it does not preserve `<sub>` and `<sup>` tags when pasting?

Even though most of the time I wish we had a simple markdown editor, the HTML thing has one big advantage to how I use Vikunja: sometimes, I copy-paste chunks from webpages that a task needs to reference instead of adding just a link with a fragment both to avoid having to go there again in the future and to get rid of the risk of having a broken link in a few months. When I do that, it basically **just works** and that's fantastic. Except for the `<sub>` and `<sup>` tags.

That can actually really impair legibility in some cases. Is this easy to change? Is there a reason not to change it? Having `<sub>` and `<sup>` buttons in the editor would also be nice, but I don't think I would use them often.

I chose the "Bug" type, but feel free to change it to a feature request, if that makes more sense to you.

### Vikunja Version

v0.24.6

### Browser and version

Chrome Version 138.0.7204.184 (Official Build) (64-bit)

Yes

### Screenshots

_No response_

kergma-lw:
Closes #1222

This adds general subscript and superscript support to the task description editor so pasted rich text keeps `<sub>` and `<sup>` formatting instead of flattening it into plain text.

I also added coverage for this in both editor-level and browser-level tests.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -67,6 +67,8 @@
     "@tiptap/extension-link": "3.30.3",
     "@tiptap/extension-list": "3.30.3",
     "@tiptap/extension-mention": "3.30.3",
+    "@tiptap/extension-subscript": "3.30.3",
+    "@tiptap/extension-superscript": "3.30.3",
     "@tiptap/extension-table": "3.30.3",
     "@tiptap/extension-typography": "3.30.3",
     "@tiptap/extension-underline": "3.30.3",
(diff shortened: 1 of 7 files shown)


kolaente (approved):
Thanks!

PR #3672, opened 2026-08-30
chore(dev): add local docker-compose stack + agent cost logging

oshratshabtai:
## What

Adds a local development stack so the repo can be brought up end-to-end with a single command:

- **`docker-compose.yml`** — Postgres 16 + the Vikunja `api` service built from the repo's own `Dockerfile`. Vikunja ships as one binary serving both the API and the built frontend, so there is no separate frontend container (matches upstream).
- **`.env.example`** — scaffolding for the Anthropic key used by the agent tooling.
- **`agent/cost_logger.py`** + **`agent/requirements.txt`** — a thin cost-logging wrapper around Anthropic API calls, in place from the start so spend is tracked.
- **`.gitignore`** — ignore `.env`.

A follow-up commit fixes two issues found bringing the stack up for the first time:

- `VIKUNJA_SERVICE_PUBLICURL` set to `http://localhost:3456` so the frontend talks to the right origin.
- `VIKUNJA_FILES_BASEPATH` pointed at `/tmp/files` — `/app/vikunja` is not writable by the container's non-root user, `/tmp` is (`--chown=1000:1000 --chmod=1777` in the Dockerfile).

## Why

Give reviewers and the agent phases a clean, reproducible way to run the app locally against Postgres.

## Verification

`docker compose up --build` from the repo root:

- image builds, `vikunja-db` reports healthy, migrations run clean
- API listens on `:3456`, `curl http://localhost:3456/api/v1/info` → `200`
- registered a user and logged in via the browser — end-to-end healthy

The `401` responses on `/api/v1/user/token/refresh` before login are expected (SPA fires a refresh on load with no session yet).

## Notes for reviewer

- `VIKUNJA_SERVICE_JWTSECRET` is a hardcoded throwaway value — local dev only.
- Registration is enabled (`VIKUNJA_SERVICE_ENABLEREGISTRATION: "true"`) for the same reason.


oshratshabtai:
Opened against the wrong repository by mistake — this belongs on my fork, not upstream. Closing. Apologies for the noise.

PR #3707, opened 2026-09-02
fix(db): redact PostgreSQL credentials from connection errors

reggi:
## Summary

- sanitize PostgreSQL initialization and ping errors before they reach logs or doctor output
- redact complete DSN userinfo without corrupting diagnostics when passwords are short
- add regression coverage for malformed PostgreSQL URLs that include credentials

## Security impact

PostgreSQL URL parsing errors can include the full credential-bearing DSN. Existing callers log or display the returned error directly, which can expose the database username and password.

## Testing

Not run locally.

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -17,10 +17,12 @@
 package db
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


kolaente (approved):
Good catch, thanks

Issue #2151, opened 2026-01-25
Improvements on dropdown for "New task relation"


pe1uca:
### Description

There are a few comments I want to make about the dropdown to select a task for a relationship, just minor improvements for the UX I'd fine useful, maybe someone else also would do so.  

- Have the tasks of the current project at the top of the options.  
- Have the ID of the task shown in the entry.  
- Not sure how useful is to have tasks mark as "done" in this list, so maybe remove them?  
- Add an always visible marker on the last entry to denote it creates a new task (right now you need to pay attention to a message which only appears when the entry is highlighted)  
- When searching only by ID, remove the option to create a new task (I've created several tasks with title ``#6``)  
- Be able to use ID directly instead of prefixing it with ``#``.  

<img width="667" height="350" alt="Image" src="https://github.com/user-attachments/assets/b5e05c83-c73a-4b93-8ee7-bcb7fe77c72d" />

kergma-lw:
@kolaente I suggest hiding the create option when searching by a task reference like #123, to prevent accidental task creation. Does that make sense?

kolaente:
I wouldn't want to hide the creation, I'd guess we'll get an issue complaining if we do that.

The creation has a plus icon now so I'd say it's clearly distinguishable now:

<img width="417" height="142" alt="Image" src="https://github.com/user-attachments/assets/d185a46a-41d9-44a0-8ba5-b4243d7ef1c9" />

I think the rest of this issue is done then with https://github.com/go-vikunja/vikunja/pull/3711 getting merged?

kergma-lw:
The plus icon makes the action visually distinguishable, but accidental creation is still easy: clicking anywhere on the #5556 create row - not just the plus - or pressing Enter creates a task with that title while I'm looking for an existing task by ID. This is exactly the kind of accident @pe1uca  reported.

A bare task reference like #5556 doesn't seem like a common intentional task title. In this context, I'd expect it to mean “find this task” much more often than “create a task with this exact title.”

Hiding creation for reference-only queries in the relation picker wouldn't prevent anyone from using such a title: they could still create the task through the normal creation flow. Given that, wouldn't preventing the reported accidental creations outweigh preserving this particular shortcut for what seems like an uncommon use case?

PR #3711, opened 2026-09-02
feat(task): improve relation task search results

kergma-lw:
Implements part of #2151.

Included in this PR:

- task IDs in relation dropdown results
- current-project results first, without lifting done tasks above open ones


kolaente (approved):
Thanks!

PR #3730, opened 2026-09-04
fix(frontend): open saved-filter settings after reload 🤖🤖🤖

### LINKED ISSUE #594: Ambiguous routes: `ProjectEdit` and `FilterEdit`
_Original issue by dpschen on 2025-01-19T14:04:27.000Z_

### Description

The routes to edit projects and filters are not ambiguous.
Consecutive loading the edit modal of a filter via it's url opens a buggy edit modal of a (non-existing) project instead.

While working on https://kolaente.dev/vikunja/vikunja/pulls/2950 this caught my attention, because the folderpaths are basically identical.

This probably wasn't caught because navigation via the RouteName works as intended (`'project.settings.edit'` and `'filter.settings.edit'`), so only a reload triggers the error.

Reason seems to be that both routes target the same address:

**ProjectSettingEdit:**

```json
		{
			path: '/projects/:projectId/settings/edit',
			name: 'project.settings.edit',
			component: ProjectSettingEdit,
			props: route => ({ projectId: Number(route.params.projectId as string) }),
			meta: {
				showAsModal: true,
			},
		},
```

**FilterEdit:**

```json
		{
			path: '/projects/:projectId/settings/edit',
			name: 'filter.settings.edit',
			component: FilterEdit,
			meta: {
				showAsModal: true,
			},
			props: route => ({ projectId: Number(route.params.projectId as string) }),
		},
```

Step by step reproduction:

1) Navigate to `/projects` and click on `New Saved Filter`.
2) Type `Example` and click `Create Saved Filter`.
3) The newly created filter opens
4) Click in the header bar on the three dots next to the filter name and select `Edit` in the menu that opens. The `FilterSettingsEdit` modal opens (title: 'Edit This Saved Filter').
5) Hard reload the browser.
6) The `ProjectSettingsEdit` modal opens (title: 'Edit This Project').

-------

Since https://kolaente.dev/vikunja/vikunja/pulls/2950 changes a lot of the routing I would wait for that branch to be merged.

In general we would need to check if the id number starts with a minus and behave accordingly.
We should probably also check if there are other routes that should be adjusted.
E.g. we could adapt the 'duplicate project' modal accordingly respectively add a 'duplicate saved project'.

### Vikunja Version

v0.24.1-696-ea6b141d42

### Browser and version

all

Yes

### Screenshots

[Original issue on Gitea](https://kolaente.dev/vikunja/vikunja/issues/3029)

---

_**@kolaente** commented on 2025-01-19T19:48:29.000Z_:

> In general we would need to check if the id number starts with a minus and behave accordingly.

So this would be the solution here? And then unify both components?

---

_**dpschen** commented on 2025-01-19T19:52:46.000Z_:

This could be **a** solution. But to be honest I think it would be better to separate projects from filters in the routes. Or is there some specific reason why they need to me together? Creating new filters also happens on a route that has 'filter' in it.

ryanchou1994:
## What this fixes

Hard-reloading a saved-filter edit or delete URL currently matches the project settings route first and opens the wrong modal.

- constrain project settings routes to non-negative numeric IDs
- route saved-filter IDs `<= -2` to filter settings while leaving Favorites (`-1`) unmatched
- use the correct `projectId` parameter for named delete navigation
- cover direct URL and named-route resolution

Fixes #594

## Testing

- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/router/index.test.ts` (10 passed)
- ESLint on the three changed production files (0 errors)
- the same focused test on Linux with Node.js 24 (10 passed)
- `git diff --check`

## Screenshots

Not applicable: this changes route resolution and reuses the existing modals; it does not change visual styling or layout.

## AI assistance

This pull request was prepared by an automated coding agent with AI assistance and opts into the repository's automated-agent process. The agent reproduced the bug on current `main`, reviewed the complete diff, and ran the tests above. Ryan authorized the submission and remains available for maintainer follow-up.


kolaente (approved):
Thanks!

PR #3819, opened 2026-09-08
fix(migration): preserve todoist task descriptions

blancqua:
## Summary
- The Todoist v1 sync API returns a `description` field on every item, but the migrator's `item` struct had no such field, so Go's JSON decoder silently dropped it — every task description was lost during migration.
- Adds `Description string \`json:"description"\`` to the `item` struct and maps it onto `task.Description` during conversion.
- Existing behaviour (Todoist notes appended to the description) is preserved: notes are now appended *after* the migrated description with a newline, same as before for note-only tasks.

## Root cause evidence
Live sync payload (`POST /api/v1/sync` with `resource_types=["items"]`) includes `description` per item; after migrating a real account, all five tasks that had descriptions came over with empty descriptions.

## Test plan
- [x] New `TestTodoistItemDecoding` proves the field decodes (failed to compile before the struct change — RED)
- [x] New `TestConvertTodoistToVikunjaPreservesDescriptions`: desc-only, desc+note, note-only (regression) — failed before fix, passes after
- [x] Full package `go test ./pkg/modules/migration/todoist/` green

### DIFF
--- a/pkg/modules/migration/todoist/todoist.go
+++ b/pkg/modules/migration/todoist/todoist.go
@@ -83,6 +83,7 @@ type item struct {
 	UserID         string      `json:"user_id"`
 	ProjectID      string      `json:"project_id"`
 	Content        string      `json:"content"`
+	Description    string      `json:"description"`
 	Priority       int64       `json:"priority"`
 	Due            *dueDate    `json:"due"`
 	ParentID       string      `json:"parent_id"`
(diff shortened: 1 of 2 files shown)


kolaente (approved):
Thanks.

Please tell your coding agent what a good PR description looks like (for example here: https://github.com/go-vikunja/vikunja/pull/3804)

kolaente:
I use this skill to create these: https://gist.github.com/kolaente/9e4c1859aa2611a8774f63941717ec2c

blancqua:
Thanks for merging, and for the pointers! We've adopted that PR description format going forward (installed your open-pr skill for our agent runs), so future PRs will follow it.

PR #3820, opened 2026-09-08
fix(migration): parse more todoist recurrence formats per language

blancqua:
Todoist exposes recurrence only as free text in the user's language (`due.string`), so `parseTodoistRepeat` — which only knew bare English phrases like "every 3 months" — left most recurring tasks non-repeating: in a real 56-task export, 26 of 31 recurring tasks lost their recurrence entirely.

## What changed

- **Parsed forms:** bare adverbs ("daily", "wekelijks"), intervals ("every 3 weeks", "elke 4 weken", "every other day"), date-anchored yearlies ("yearly 1st May", "elke 16e jul", "every 15/03"), day-of-month monthlies ("every 15th" → `TaskRepeatModeMonth` so the day stays anchored despite unequal month lengths), and weekdays ("elke maandag" → weekly; the due date already anchors the weekday). Anything unparsed still returns 0 with the existing debug log — no behaviour change for unknown input.
- **Lang-keyed structure (rework after review feedback):** the shared EN+NL word maps and regexes are replaced by per-language data packs (`langPacks`, selected by `due.lang` — en and nl today) running through a language-agnostic grammar. A new language is a new data block with zero grammar changes; unknown languages fall back to `en`, and a missing `lang` (which real exports never produce) tries every pack as a whole so lang-less callers keep the old behaviour. The table and the small `parseTodoistRepeat` entry point are deliberately swappable for the planned multilingual date parser.

## How to verify

1. Run the Todoist migration (Settings → Migrations → Todoist) with an account that has recurring tasks using strings like "elke 4 weken", "elke 3 maanden", "yearly 1st May", "every 15th", "every 15/03", "elke maandag".
2. Open the migrated tasks in Vikunja.
3. **Expected:** each task shows a matching repeat interval and mode — e.g. "elke 4 weken" → repeats every 4 weeks; "every 15th" → monthly repeat anchored to day 15; "elke maandag" → weekly repeat.
4. **Before this PR:** all of these migrated as non-repeating tasks.

### DIFF
--- a/pkg/modules/migration/todoist/todoist.go
+++ b/pkg/modules/migration/todoist/todoist.go
@@ -20,9 +20,11 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"net/http"
 	"net/url"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


kolaente:
We can't keep adding things like that. That's not practical at all and unmaintainable. Open for suggestions?
The goal is to make our own date magic parser multi-lingual at some point, maybe we could reuse it.

(To me it's wild that Todoist seems to just store the text the user entered instead of normalizing like we do)

blancqua:
Agreed that stacking per-language regexes doesn't scale — I've reworked the PR along those lines.

Two things that helped:

1. Todoist's due object includes a `lang` field (`"lang": "nl"`), always present in the exports I checked. So no language *detection* is needed, just a lookup: the grammar is now language-agnostic and each language is one data table (units, month/weekday names, ordinals, adverbs) keyed on `due.lang`, falling back to English. Adding a language is data-only — no new code paths.
2. I checked the v1 sync API for a normalized recurrence field: there is none, only `string` + `lang` + `is_recurring` — confirming your "wild that they store raw text" (text parsing really is the only option at migration time).

The entry point is deliberately small and swappable so this can be replaced by your multilingual date parser when that lands. Happy to help build that out — point me at where you want it to live.

pkg/modules/migration/todoist/todoist.go
```diff
@@ -263,62 +265,241 @@ const (
...
+	weekdays        map[string]struct{}
+	months          map[string]struct{}
+	ordinals        string // regex alternation of ordinal suffixes: "st|nd|rd|th"; nl: "e|ste|de"
+	ordinalDayRegex *regexp.Regexp
+}
+
+// langPacks is deliberately data-only: a new language is a new data block here, with zero
+// grammar changes. Todoist plans to move its date parsing upstream into a multilingual
+// parser, so keep this table and the small parseTodoistRepeat entry point swappable for it.
+var langPacks = compileRepeatLangs(map[string]*repeatLang{
```
kolaente: please move the type for this and the variable to its own file

kolaente (changes requested):

kolaente:
Sounds good on the approach. I'll check if we can add support for more languages.

pkg/modules/migration/todoist/todoist.go
blancqua: Done — moved the `repeatLang` type, the `langPacks` data and its helpers to `repeat_langs.go` (commit 77eb34126).

PR #3823, opened 2026-09-08
feat(openid): send nonce and PKCE in the authorization request

kiyoteruhosoda:
## What

The OIDC client sends only `client_id`, `redirect_uri`, `response_type`, `scope` and `state` to the provider. This adds the two parameters the specs ask a client to send:

- a **nonce**, kept for the callback and compared against the ID token's `nonce` claim
- a **PKCE** `code_challenge` (S256), with the verifier passed to the token exchange

## Why

Providers are free to require either one, and some do. Authelia's integration guide for Vikunja has to turn PKCE off for the client (`require_pkce: false`, `pkce_challenge_method: ''`), and a provider that requires a nonce rejects the authorization request before the user ever sees a login screen:

```
302 .../auth/openid/<provider>?error=invalid_request&error_description=nonce+is+required
```

Both are defined for exactly this: the nonce in OpenID Connect Core 1.0 §3.1.2.1, PKCE in RFC 7636. Vikunja's own OAuth 2.0 authorization server already requires `code_challenge` from the clients that talk to it.

## Compatibility

Nothing changes for providers that do not implement these:

- a provider that does not support PKCE **ignores** `code_challenge` and `code_challenge_method` (RFC 7636 §5)
- the backend only passes `code_verifier` to the exchange when the client sent one, and only compares the nonce when the client sent one — so a flow started before this change, or by a client that sends neither, behaves exactly as it did

⚠ One caveat worth knowing about: `crypto.subtle` is restricted to secure contexts, so an instance served over plain http cannot hash the verifier. Rather than send a challenge it could never redeem — which would lock those users out — the frontend leaves PKCE out of the request in that case. The nonce needs no hashing and is always sent.

## Tests

- `frontend/src/helpers/pkce.test.ts` — the verifier's shape and length, and the challenge checked against the test vector in RFC 7636 appendix B
- `frontend/src/helpers/redirectToProvider.test.ts` — the authorization request keeps every parameter it used to carry, gains nonce and challenge, and drops PKCE (only PKCE) when the browser exposes no subtle crypto

`pnpm lint`, the frontend unit tests and `go test ./pkg/modules/auth/openid/` pass locally. `pnpm typecheck` reports errors, but they are all pre-existing on `main` — none are in the files this PR touches.

The nonce comparison itself has no test: there is no provider double in `pkg/modules/auth/openid` to drive a token exchange through, and standing one up felt out of proportion for a two-line check. Happy to add one if you would rather have it.

## Note on the random values

`createRandomID` is backed by `Math.random`, which is not suitable for a nonce or a code verifier, so the new helper draws from `crypto.getRandomValues` instead. The existing `state` is left as it is — changing it is a separate concern from this PR.


kiyoteruhosoda:
Thanks for the review — all six are addressed in 3b91288.

**`state` from `Math.random` (MEDIUM).** Fixed. `state` now comes from the same CSPRNG through a `createState()` in `pkce.ts`. You are right that it is adjacent and that the fix was free, and the login-CSRF path you describe is real: predicting the next `state` is enough to log a victim into the attacker's account. Added a test that fails if `Math.random` is touched anywhere in the request.

**Optional enforcement (MEDIUM, needs input).** Kept opt-in for this PR, but I have made the field doc say so rather than imply otherwise — `Nonce` now reads "A client which sends none skips the comparison — this is a best-effort check, not one Vikunja enforces."

My reasoning for not enforcing here: turning either one on unconditionally breaks clients mid-upgrade — a browser holding a pre-upgrade flow in `localStorage`, the desktop app, anything driving the callback directly — and it would also lock out plain-http instances, which cannot produce a challenge at all (see the `crypto.subtle` caveat). A per-provider `requirepkce` / `requirenonce` flag is the right shape for operators with a compliant IdP, but it is a new config surface with its own docs and sample entries, so I would rather it be its own PR. Happy to write it either way — say the word and I will fold it in here.

**Spent values left in `localStorage` (LOW).** Fixed — both are cleared right after the callback reads them.

**Unhandled rejection from `digest` (LOW).** Fixed with a `try`/`catch` degrading to `undefined`, so a hash failure lands on the documented no-PKCE path instead of an unhandled rejection. Also dropped the unreachable `typeof crypto === 'undefined'` half of the guard — correct, `createNonce` calls `getRandomValues` unguarded and runs first.

**Test proving nothing (LOW).** Fixed — it seeds a stale verifier before the request now. Good catch: it passed with the `removeItem` line deleted.

**Comment volume (NIT).** Trimmed to your list, about 25 lines shed. Kept the RFC 7636 appendix B pointer since it pins a magic value to its source, and the empty-`Nonce` line since the bypass reads like a hole without it.

PR #3837, opened 2026-09-08
docs: add swagger annotations for the OAuth 2.0 endpoints

sjwasko:
`POST /api/v1/oauth/authorize` and `POST /api/v1/oauth/token` are served but missing from the generated OpenAPI docs: the handlers in `pkg/modules/auth/oauth2server` carry no swagger annotations, so neither path is in `swagger.json` — 126 paths, no `oauth` among them.

Found while building a third-party client. The API answers fine; the docs say the endpoints don't exist, so a client that validates its calls against them can't use OAuth at all.

Adds annotations for both handlers and regenerates `pkg/swagger`. No behaviour change — comments only, plus generated output. `mage generate:swagger-docs` produces 521 insertions and 0 deletions, so nothing else in the spec moved.

Checked against a 2.5.0 instance: `OPTIONS` on both paths returns `Allow: OPTIONS, POST`, and `POST /api/v1/oauth/token` with no grant type returns code 17007 naming `authorization_code` and `refresh_token`.

One judgement call worth a look: I annotated the token endpoint `@Accept json`, matching the comment on `TokenRequest` that v1 binds JSON and v2 takes the form-encoded body. In practice v1 binds a form body too — `Content-Type: application/x-www-form-urlencoded` with `grant_type=nope` returns the same 17007 — so if you'd rather document both, say the word and I'll widen it.

The `/api/v2` routes are untouched — happy to cover those too if you'd like them in.

AI-assisted: annotations drafted with Claude, reviewed and tested by me.

### DIFF
--- a/pkg/modules/auth/oauth2server/authorize.go
+++ b/pkg/modules/auth/oauth2server/authorize.go
@@ -46,6 +46,18 @@ type AuthorizeResponse struct {
 // HandleAuthorize handles POST /oauth/authorize.
 // It validates the OAuth parameters, creates an authorization code, and
 // returns it as JSON. Authentication is handled by the token middleware.
+// @Summary OAuth 2.0 authorize endpoint
+// @Description Creates an authorization code for an OAuth 2.0 client on behalf of the authenticated user. PKCE is required. API tokens cannot be used to authorize a client.
+// @tags auth
+// @Accept json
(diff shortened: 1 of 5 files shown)


kolaente (approved):
Good catch. Thanks!

PR #3883, opened 2026-09-14
fix(caldav): parse Z-suffixed timestamps as UTC

### LINKED ISSUE #3871: CalDAV: absolute VALARM triggers are parsed in the server timezone instead of UTC
## Summary

`caldavTimeToTimestamp()` treats the `Z` in a UTC timestamp as a literal layout
character and then parses the value in the configured server timezone. Any
CalDAV property that ends in `Z` and carries no `TZID` parameter gets shifted by
the server's UTC offset.

Absolute VALARM triggers are the ones that suffer. RFC 5545 §3.8.6.3 requires
them to be in UTC, so they can never carry a `TZID` and always take the broken
path. `DUE` and `DTSTART` mostly escape it, because clients send them with
`TZID`.

## Environment

- Vikunja 2.6.0 (Docker `vikunja/vikunja:2`), PostgreSQL 18
- `VIKUNJA_SERVICE_TIMEZONE: Europe/Berlin` (UTC+2 at the time of the test)
- Client: iOS 27.0 Reminders (`remindd/4046.11`) over CalDAV

## Steps to reproduce

1. Set `VIKUNJA_SERVICE_TIMEZONE` to a zone with a non-zero UTC offset.
2. `PUT` a VTODO whose alarm trigger is an absolute UTC timestamp:

```
BEGIN:VTODO
UID:B9FABCCA-9F29-46C2-A0FF-41219037F923
SUMMARY:Test
DUE;TZID=Europe/Berlin:20260914T070000
BEGIN:VALARM
TRIGGER;VALUE=DATE-TIME:20260914T050000Z
ACTION:DISPLAY
DESCRIPTION:Test
END:VALARM
END:VTODO
```

3. Read the stored reminder back (API or database).

## Expected

The reminder is stored as `2026-09-14 05:00:00Z`, i.e. 07:00 Europe/Berlin.
That is the instant the client sent, and it matches the task's due date.

## Actual

The reminder is stored as `2026-09-14 03:00:00Z`, i.e. 05:00 Europe/Berlin, two
hours early. `due_date` is correct at `2026-09-14 05:00:00Z`, so the task and its
alarm disagree by exactly the server's UTC offset. The size of the error follows
DST: one hour in winter, two in summer. A reminder created before a DST change
for a date after it is off by whatever offset was in effect when it was parsed.

## Root cause

`pkg/caldav/parsing.go`, `caldavTimeToTimestamp()`:

```go
format := DateFormat                  // 20060102T150405

if strings.HasSuffix(tstring, "Z") {
    format = `20060102T150405Z`       // the Z is a literal here, not a zone
}
...
} else {
    t, err = time.ParseInLocation(format, tstring, config.GetTimeZone())
}
```

Go's reference layout has no bare `Z` that means UTC. `Z` only denotes a zone
when written as `Z07:00` or `Z0700`. Here the trailing `Z` matches as an
ordinary character, and the timestamp is then read in `config.GetTimeZone()`.

The outbound direction is correct. `makeCalDavTimeFromTimeStamp()` in
`pkg/caldav/caldav.go` does `ts.In(time.UTC).Format(DateFormat) + "Z"`. So the
bug is one-sided and does not cancel out over a round trip: the client writes
07:00, the server stores 05:00 and serves 05:00 back.

## Suggested fix

Pick the location too, not just the layout, and let a `Z` suffix win over any
`TZID` (RFC 5545 forbids combining the two):

```go
loc := config.GetTimeZone()
if strings.HasSuffix(tstring, "Z") {
    format = `20060102T150405Z`
    loc = time.UTC
}
```

A regression test with a `Z`-suffixed `TRIGGER;VALUE=DATE-TIME` under a non-UTC
`VIKUNJA_SERVICE_TIMEZONE` would catch this. The existing round-trip tests don't,
because reader and writer only agree with each other when the configured offset
happens to be zero.

## Impact

Any client that sends UTC timestamps without a `TZID` is affected, not only iOS.
For task alarms the visible effect is that reminders fire at the wrong time, and
that wrong time goes back to the client on the next sync.

anandghegde:
Credit to @jnsvgt for the root-cause analysis and suggested fix in #3871. This PR implements that fix and adds regression tests.

`caldavTimeToTimestamp()` used the `Z` suffix only to pick the layout, where it matches as a literal character. The value was then parsed in the server timezone. With a non-UTC `service.timezone`, every UTC value without a `TZID` was shifted by the server's offset. Absolute VALARM triggers are always UTC, so reminders were stored hours early.

Fix: when the value ends in `Z`, parse it in UTC and ignore any `TZID` (RFC 5545 doesn't allow combining the two). Floating values and `TZID` values work as before. DTSTART, DUE, DTEND, DTSTAMP and COMPLETED all go through the same helper, so they get the fix too.

Tests, with the server timezone set to Europe/Berlin:
- `TestCaldavTimeToTimestamp_NonUTCServerTimezone`: covers UTC, floating, TZID, and UTC combined with TZID
- `TestParseTaskFromVTODO_UTCAlarmNonUTCServerTimezone`: the VTODO from the issue; the reminder should match the due date

Before the fix, the UTC cases and the alarm test fail (`05:00 CEST` instead of `05:00 UTC`). After the fix, they pass. `go test ./pkg/caldav`, `mage test:caldav` and `mage lint` also pass.

I wrote this with help from Claude Code, then reviewed and tested it myself.

Note on #1068: that change made floating times (no `Z`, no `TZID`) use the configured timezone, which this PR keeps. It also routed `Z`-suffixed UTC times through the same path; this PR only restores UTC parsing for those.

Fixes #3871


kolaente (approved):
Thanks!

anandghegde:
Thanks for the review! The red test-api jobs are a Go module proxy hiccup, not the change (`proxy.golang.org ... stream error: INTERNAL_ERROR` while downloading go-sqlite3 during setup). A re-run of the failed jobs should clear it — I don't have permission to trigger one.

PR #3964, opened 2026-09-19
fix(frontend): hide PWA banners until the user is logged in

### LINKED ISSUE #3880: PWA banners show on the login page
### Pre-submission checklist

### Description

On my self-hosted instance, the update banner and add-to-home-screen prompt can both appear on the login page. They don't make much sense before someone has signed in.

`App.vue` renders both components at the top level:

    <AddToHomeScreen v-if="!isQuickAddMode" />
    <UpdateNotification v-if="!isQuickAddMode" />

They're siblings of the `ContentAuth` / `ContentLinkShare` / `NoAuthWrapper` switch. The only condition on either component is `!isQuickAddMode`, so the login page gets them too.

The update banner is particularly confusing there. In production, `registerServiceWorker.ts` registers `sw.js` and dispatches `swUpdated` from its `updated()` callback. `UpdateNotification.vue` listens for that event and sets `baseStore.updateAvailable`. Clicking its button sends `skipWaiting` to `registration.waiting`, then reloads the page. A logged-out visitor just sees "There is an update available!", which sounds like an update to the hosted service.

`AddToHomeScreen.vue` has no auth check either. Its `shouldShowMessage` only checks the `hideAddToHomeScreenMessage` localStorage flag and whether `window.matchMedia('(display-mode: standalone)')` matches.

Could these two components only render after authentication? The update message could also say something like "A new version of this page is ready. Reload to use it", which is clearer about what the button does.

Issues #3251 and #3263 touched `UpdateNotification` for the first service worker install reloading and wiping the login form. This is a separate login-page case in the same component.

### Vikunja Version

v2.6.0 (`vikunja/vikunja` Docker image)

### Browser and version

Desktop browser

Yes

mnkprs:
Fixes #3880

The update banner and the "add to home screen" banner showed on the login page, because `App.vue` mounts them outside the logged-in layout switch.

Both components now check `authStore.authenticated` themselves. I kept the check inside the components instead of adding a `v-if` in `App.vue`: `UpdateNotification` listens for the one-time `swUpdated` event (`{once: true}`), and that event often fires while the user is still on the login page. Unmounting it there would lose the update prompt for the rest of the session.

`authenticated` is also true for link-share guests, so they still see both banners as before. Using `authUser` instead would hide them for link shares too, if you prefer that.

I left the `update.available` wording as is. Happy to change it in a follow-up.

## How to verify

1. Open `/login` in a phone-sized window: no banners.
2. In the console, run `document.dispatchEvent(new CustomEvent('swUpdated', {detail: {waiting: null}}))`: still no banners.
3. Log in: both banners show.

The new `tests/e2e/misc/pwa-banners.spec.ts` covers these steps. Its logged-out test fails without the fix.

I used Claude Code to help write this change, and reviewed and tested it myself.


frontend/src/components/home/UpdateNotification.vue
```diff
@@ -19,10 +19,13 @@
 <script lang="ts" setup>
 import {computed, ref} from 'vue'
 import {useBaseStore} from '@/stores/base'
+import {useAuthStore} from '@/stores/auth'
 
 const baseStore = useBaseStore()
+const authStore = useAuthStore()
 
-const updateAvailable = computed(() => baseStore.updateAvailable)
+// Gate here, not in App.vue: unmounting would miss the one-time swUpdated event fired on the login page
```
kolaente: please tell claude to remove comments like these

frontend/src/components/home/AddToHomeScreen.vue
```diff
@@ -26,13 +26,19 @@ import BaseButton from '@/components/base/BaseButton.vue'
...
+import {useAuthStore} from '@/stores/auth'
 
 const baseStore = useBaseStore()
+const authStore = useAuthStore()
 
 const hideMessage = useLocalStorage('hideAddToHomeScreenMessage', false)
 const hasUpdateAvailable = computed(() => baseStore.updateAvailable)
 
 const shouldShowMessage = computed(() => {
+	if (!authStore.authenticated) {
```
kolaente, line 38: Isnt this redundant?

kolaente (changes requested):

frontend/src/components/home/UpdateNotification.vue
mnkprs: Removed.

frontend/src/components/home/AddToHomeScreen.vue
mnkprs: Not redundant: without it the banner still renders on the login page, the logged-out e2e test fails on `.add-to-home-screen`. It sits in the component to match `UpdateNotification`, which has to stay mounted for the `swUpdated` listener. Happy to move this one to a `v-if` in `App.vue` instead if you prefer the gate there.

PR #3991, opened 2026-09-21
fix(user): fallback empty user timezone to valid UTC location (#3965)

### LINKED ISSUE #3965: TOTP login failed due to time zone being ''
### Pre-submission checklist

### Description

Firstly, thanks for the great app.

After a new install, I wanted to add 2FA. I scanned QR code, entered the 6 digit password and it automatically logged out. I wanted to login and it was quick enough that TOTP _was same as before_. This caused the error in the image (timezone one). After doing nothing, I tried to login again, but the TOTP had been updated by then, so I got the error in the second image (as expected). After entering correct TOTP, I successfully logged in. 

When I checked "General Settings > Time Zone", it was set to GMT. After logging in and out several times, I didn't face any issues.

I am using docker image: vikunja/vikunja:2.6

### Vikunja Version

2.6

### Browser and version

_No response_

No

### Screenshots

Image
Image

Tyagiquamar:
Fixes #3965: Fallback empty user timezone to valid UTC location on user creation and update to prevent TOTP login failures.


pkg/user/user_create.go
```diff
@@ -76,6 +76,12 @@ func CreateUser(s *xorm.Session, user *User, options ...CreateUserOptions) (newU
 	user.DefaultProjectID = config.DefaultSettingsDefaultProjectID.GetInt64()
 	user.WeekStart = config.DefaultSettingsWeekStart.GetInt()
 	user.Timezone = config.DefaultSettingsTimezone.GetString()
+	if user.Timezone == "" {
+		user.Timezone = config.GetTimeZone().String()
+		if user.Timezone == "" {
+			user.Timezone = "UTC"
```
kolaente: Don't fall back the user timezone again, the global time zone setting should be enough. The if on line 81 will never be true with your change in the config setting

pkg/config/config.go
```diff
@@ -297,6 +297,9 @@ var timezone *time.Location
 // it way easier, especially when testing.
 func GetTimeZone() *time.Location {
 	tz := ServiceTimeZone.GetString()
+	if tz == "" {
+		tz = "UTC"
```
kolaente: This should log a warning on start up

kolaente (changes requested):

pkg/user/user_create.go
Tyagiquamar: Addressed in b4aaf36b3: removed the duplicate user-level fallback branch so the global timezone fallback is used directly.

pkg/config/config.go
Tyagiquamar: Addressed in b4aaf36b3: added validation in InitConfig to log a warning when service.timezone or defaultsettings.timezone is unconfigured or invalid, and added unit tests in TestTimezoneFallback.

pkg/config/config.go
```diff
@@ -820,9 +823,18 @@ func InitConfig() {
 		MigrationMicrosoftTodoRedirectURL.Set(ServicePublicURL.GetString() + "migrate/microsoft-todo")
 	}
 
+	if tz := ServiceTimeZone.GetString(); tz == "" {
+		log.Warning("service.timezone is not configured, falling back to UTC")
+		ServiceTimeZone.Set("UTC")
```
kolaente, line 825: If we're just setting this here we don't need the fallback in GetTimeZone

kolaente (changes requested):

pkg/config/config.go
Tyagiquamar: Addressed in 3d7f2f707: removed the redundant fallback check from GetTimeZone so it relies directly on InitConfig initialization.

